PR #35, opened 2023-05-11
feat: missing stack features & refactoring

enitrat:
Refactors the Stack implementation and adds missing methods like `pop_n` and `swap_i`
This implementation is a bit closer to Cairo0's implementation, but with more abstractions thanks to Cairo1's features.

One thing I would like to discuss is: should we panic (like when trying to `pop_n` with `n>len`) or return Options that will be handled afterwards?

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


tests/test_stack.cairo
```diff
@@ -7,94 +7,291 @@ use dict::Felt252DictTrait;
...
-    stack.push(val_2);
-    let result_len = stack.len();
-    let result_is_empty = stack.is_empty();
-    let low = stack.items.get(0);
-    let high = stack.items.get(1);
+    // When
+    stack.push(1);
+
+    // Then
+    let high = stack.items.get(0);
```
ClementWalter, line 62: we shouldn't access this internal impl detail given the name of the test, only assert that `stack.peek()` returns the pushed value

tests/test_stack.cairo
```diff
@@ -7,94 +7,291 @@ use dict::Felt252DictTrait;
...
+
+#[test]
+#[available_gas(2000000)]
+#[should_panic(expected: ('Kakarot: StackOverflow', ))]
+fn test__push__should_fail_when_overflow() {
+    // Given
+    let mut stack = StackTrait::new();
+
+    // When
+    stack.len = constants::STACK_MAX_DEPTH;
```
ClementWalter, line 79: see comment above, I don't think this is a good test. Can we `loop` to fill the stack instead?

tests/test_stack.cairo
```diff
@@ -7,94 +7,291 @@ use dict::Felt252DictTrait;
...
+
+    // Then
+    stack.push(1);
+}
+
+#[test]
+#[available_gas(2000000)]
+fn test__pop__should_pop_an_element_to_the_stack() {
```
ClementWalter, line 87: pop *from* the stack

tests/test_stack.cairo
```diff
@@ -7,94 +7,291 @@ use dict::Felt252DictTrait;
...
+
+    // When
+    let elements = stack.pop_n(3);
+
+    // Then
+    assert(stack.len == 0, 'wrong length');
+    assert(elements.len() == 3, 'wrong returned array length');
+    assert(*elements[0] == 3, 'wrong result');
+    assert(*elements[1] == 2, 'wrong result');
+    assert(*elements[2] == 1, 'wrong result');
```
ClementWalter, line 120: is it required to add an error message in the `assert` ? this "wrong result" is not very informative

src/stack.cairo
```diff
@@ -36,15 +38,17 @@ impl DestructStack of Destruct<Stack> {
 trait StackTrait {
     fn new() -> Stack;
     fn push(ref self: Stack, item: u256) -> ();
-    fn pop(ref self: Stack) -> Option<u256>;
+    fn pop(ref self: Stack) -> u256;
+    fn pop_n(ref self: Stack, n: usize) -> Array<u256>;
     fn peek(ref self: Stack) -> Option<u256>;
```
ClementWalter, line 43: why does peek return an Option btw?

ClementWalter (changes requested):

tests/test_stack.cairo
```diff
@@ -7,94 +7,291 @@ use dict::Felt252DictTrait;
...
 #[available_gas(2000000)]
-fn stack_is_empty_test() {
+fn test__len__should_return_the_length_of_the_stack() {
+    // Given
     let mut stack = StackTrait::new();
-    let result = stack.is_empty();
-    assert(result == true, 'stack should be empty');
+
+    // Then
+    assert(stack.len() == 0, 'stack length should be 0');
```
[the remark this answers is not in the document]
enitrat: I'm in favor of keeping both actually. In the future, we might want to have len as a private field (which we can't do yet)

src/stack.cairo
```diff
@@ -82,11 +111,46 @@ impl StackImpl of StackTrait {
...
+    /// Peeks at the item at the given index on the stack.
+    /// index is 0-based, 0 being the top of the stack.
+    /// # Arguments
+    /// * `self` - the Stack instance
+    /// * `index` - the index of the item to peek at
+
+    /// Returns
+    /// * u256 The item at the given index, or None if the stack is empty.
+    fn peek_at(ref self: Stack, index: usize) -> u256 {
+        if index > self.len() {
```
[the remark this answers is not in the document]
enitrat: Good catch

src/stack.cairo
```diff
@@ -119,19 +183,19 @@ impl StackU256HelperImpl of StackU256HelperTrait {
...
     ///
     /// # Arguments
     /// * `item` - the 256-bit unsigned integer to insert into the stack
     /// * `index` - the index at which to insert the item in the stack
-    fn insert_u256(ref self: Stack, item: u256, index: felt252) {
-        let dict_len: felt252 = self.dict_len();
-        self.items.insert(dict_len, item.low);
-        self.items.insert(dict_len + 1, item.high);
+    fn insert_u256(ref self: Stack, item: u256, index: usize) {
+        let real_index: felt252 = index.into() * 2;
```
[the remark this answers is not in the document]
enitrat: It acted like a `push_u256` instead of a more generic `insert_u256` at any position. We have tests but only on the Stack itself, and since they don't cover insertion at random places because that doesn't happen in a stack.

tests/test_stack.cairo
```diff
@@ -7,94 +7,291 @@ use dict::Felt252DictTrait;
...
+    // When
+    stack.len = 1;
+    // Then
+    assert(stack.len() == 1, 'stack length should be 1');
 }
 
 #[test]
 #[available_gas(2000000)]
-fn stack_push_test() {
+fn test__push__should_add_an_element_to_the_stack() {
```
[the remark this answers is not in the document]
enitrat: What for? They're already inside the `test_stack` module

tests/test_stack.cairo
enitrat: Why not? We know that `stack.push` increments the length, and that when the length is maximal, we can't push anymore.

tests/test_stack.cairo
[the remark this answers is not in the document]
ClementWalter: ok so maybe just make sure that we don't use any `.len` outside of the `impl`

tests/test_stack.cairo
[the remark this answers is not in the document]
ClementWalter: namespace for each system under testing (push, peek, pop, etc.) like you could have in js

```
describe("Stack", () => {
    describe("push", () => {
        it("should add an item", () => {}
    }
}
```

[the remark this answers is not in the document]
enitrat: The best I can do is split them in different modules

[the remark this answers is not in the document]
ClementWalter: like this?

```
mod push {
    fn test_should_add_an_element()
}
```

sounds good to me, and shorten the function names then

tests/test_stack.cairo
ClementWalter: is it something that could change in the future?

tests/test_stack.cairo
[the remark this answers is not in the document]
enitrat: yep 👍

src/stack.cairo
enitrat: Makes sense. My reasoning was that in the specific case of KKT, we shouldn't encounter a case where we would pop an empty stack; and if such a case were to happen, we would want to panic. That's how it was done in the v0 implementation. 
On the other hand, given that peek is just a `view` operation, I thought that there might be use cases where returning `None` could be useful.
I will add your changes if you think it is the right way

ClementWalter: ok so looking at the cairo 0 codebase, the only place where we actually use this `peek` method is for DUP ops, which have an error case that "Not enough values on the stack." (see https://www.evm.codes/#81?fork=shanghai) that we'll want to handle (we just panic in cairo 0). So let's keep the `Option`

enitrat: I will also modify `pop` to return an Option as well for consistency purposes.

enitrat:
@ClementWalter everything should be good now

src/stack.cairo
ClementWalter: so pop & peek return `Option` but `pop_n` and `peek_at` no, we should probably have option everywhere then right?

src/stack.cairo
```diff
@@ -58,7 +62,11 @@ impl StackImpl of StackTrait {
     /// * self The stack instance.
     /// * item The item to push onto the stack.
     fn push(ref self: Stack, item: u256) -> () {
-        self.insert_u256(item, self.dict_len());
+        // we can store at most 1024 256-bits words
+        if self.len == constants::STACK_MAX_DEPTH {
```
ClementWalter, line 66: can we use `self.len()` everywhere?

ClementWalter: or maybe `self.len`, but mixing is confusing

src/stack.cairo
```diff
@@ -82,16 +111,52 @@ impl StackImpl of StackTrait {
...
+    /// 
+    /// # Arguments
+    /// * `self` - the Stack instance
+    /// * `index` - the top-based index of the item to swap with the top of the stack
+    fn swap_i(ref self: Stack, index: usize) {
+        if index >= self.len() {
+            panic_with_felt252('Kakarot: StackUnderflow');
+        }
+        let position_0 = self.len() - 1;
+        let index_from_top = position_0 - index;
```
ClementWalter, line 147: it's actually `index_from_top` in argument and `index = position_0 - index_from_top` here right?

src/stack.cairo
```diff
@@ -82,16 +111,52 @@ impl StackImpl of StackTrait {
...
+        self.insert_u256(top_item, index_from_top);
+        self.insert_u256(swapped_item, position_0);
+    }
+
     /// Returns the length of the stack.
     /// Parameters
     /// * self The stack instance.
     /// Returns
     /// * usize The length of the stack.
+    #[inline(always)]
```
ClementWalter, line 159: inlining saves steps right?

ClementWalter (commented):
the stack overflow is much better with the loop!

src/stack.cairo
enitrat: I used `self.len()` for all "views" of the length of the array and `self.len` when there's a mutation of the length.

src/stack.cairo
enitrat: For consistency with other methods, `index` always refers to the item from the top of the Stack. The variable naming here `index_from_top` is indeed not ideal, I think `position_item` makes more sense

src/stack.cairo
enitrat: yes, it avoids a function call

src/stack.cairo
enitrat: I think it depends on how we want to manage eventual errors if there are any. I don't know enough about the context in which these methods are called, but adding Option adds small overhead costs. If KKT is built in a way where we're sure that `pop_n` and `peek_at` should always panic in case of over/underflow, then I think we should keep it this way.

enitrat:
ping @ClementWalter, can we get this one merged and come back on it later if required?

ClementWalter (approved):

Issue #45, opened 2023-08-02
dev: refactor Stack to use `Nullable<u256>` instead of storing (low, high) manually


enitrat:
The current implementation of the stack stores u256 by storing both the low and high part in two consecutive dict locations. We should refactor it to store `Nullable<u256>` in the dict instead.

PR #47, opened 2023-08-03
refactor: use Nullable<u256> in Stack

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: #45 

## What is the new behavior?

- Stack uses `Nullable<u256>` instead of using helper functions storing `(low,high)` limbs
-
-

- [ ] Yes
- [x] No

## Other information

Stack tests available_gas values have been reduced to a value close to the minimum required - it will allow us to detect eventual efficiency decreases

### DIFF
--- a/src/errors.cairo
+++ b/src/errors.cairo
@@ -1,0 +1,2 @@
+const STACK_OVERFLOW: felt252 = 'Kakarot: StackOverflow';
+const STACK_UNDERFLOW: felt252 = 'Kakarot: StackUnderflow';
--- a/src/stack.cairo
+++ b/src/stack.cairo
@@ -17,22 +17,32 @@
 // Core lib imports
 use dict::Felt252DictTrait;
 use option::OptionTrait;
+use array::ArrayTrait;
 use traits::Into;
 use result::ResultTrait;
-use array::ArrayTrait;
 use kakarot::utils::constants;
 use debug::PrintTrait;
+use box::BoxTrait;
+use nullable::{nullable_from_box, NullableTrait};
+use kakarot::errors;
 
 
-struct Stack {
-    items: Felt252Dict<u128>,
-    len: usize,
+// This should be in the corelib
+trait NullableTraitExt<T> {
+    fn new(value: T) -> Nullable<T>;
 }
 
-impl DestructStack of Destruct<Stack> {
-    fn destruct(self: Stack) nopanic {
-        self.items.squash();
+impl NullableTraitExtImpl of NullableTraitExt<u256> {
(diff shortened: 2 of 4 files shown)


src/errors.cairo
```diff
@@ -0,0 +1,2 @@
+const STACK_OVERFLOW: felt252 = 'Kakarot: StackOverflow';
+const STACK_UNDERFLOW: felt252 = 'Kakarot: StackUnderflow';
```
ftupas, line 2: Nice to create errors like these!

src/stack.cairo
```diff
@@ -17,24 +17,34 @@
...
+    fn new(value: T) -> Nullable<T>;
 }
 
-impl DestructStack of Destruct<Stack> {
-    fn destruct(self: Stack) nopanic {
-        self.items.squash();
+impl NullableTraitExtImpl of NullableTraitExt<u256> {
+    fn new(value: u256) -> Nullable<u256> {
+        let nullable = nullable_from_box(BoxTrait::new(value));
+        nullable
```
ftupas, line 38: Maybe we can add a `TODO` here to remove it once available in corelib

src/stack.cairo
```diff
@@ -145,10 +162,10 @@ impl StackImpl of StackTrait {
         }
         let position_0 = self.len() - 1;
         let position_item = position_0 - index;
-        let top_item = self.get_u256(position_0);
-        let swapped_item = self.get_u256(position_item);
-        self.insert_u256(top_item, position_item);
-        self.insert_u256(swapped_item, position_0);
+        let top_item = self.items.get(position_0.into());
+        let swapped_item = self.items.get(position_item.into());
+        self.items.insert(position_0.into(), swapped_item.into());
```
ftupas, line 167: I see that we do `position_0.into()` and `position_item.into()` here a couple of times, maybe we can do once at the top and reuse it

ftupas (commented):
Looks good! Just some minor comments 🚀

ClementWalter (approved):
lgtm, do we have an idea of the overhead introduced by the `Box`? in any case, I'd rather go cleaner first, then optimize

enitrat:
>  we have an idea of the overhead introduced by the Box

Very little, I compared the previous implementation with this one by running the tests with minimal gas values for both and it was roughly equal.

src/stack.cairo
enitrat: Once thing I'm not sure about is whether this create any CASM code. If not then this is purely for the compiler

ftupas: Gotcha, great to know thanks!

PR #57, opened 2023-08-09
tests: execution context

enitrat:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

- Apply [recommendation proposed by SW](https://github.com/starkware-libs/cairo/issues/3863#issuecomment-1670873494) regarding a bug in the tests. 
- New tests
-

- [ ] Yes
- [ ] No


src/tests/test_execution_context.cairo
```diff
@@ -3,54 +3,91 @@ use traits::Into;
...
+    assert(execution_context.memory.bytes_len == 0, 'wrong memory');
+    assert(execution_context.gas_used == gas_used, 'wrong gas_used');
+    assert(execution_context.gas_limit == gas_limit, 'wrong gas_limit');
+    assert(execution_context.gas_price == gas_price, 'wrong gas_price');
+    assert(execution_context.starknet_address == starknet_address, 'wrong starknet_address');
+    assert(execution_context.evm_address == evm_address, 'wrong evm_address');
+    assert(execution_context.destroy_contracts == destroy_contracts, 'wrong destroy_contracts');
+    assert(execution_context.events.len() == events.len(), 'wrong events');
+    assert(execution_context.create_addresses == create_addresses, 'wrong create_addresses');
+    // can't compare dictionaries directly
```
Eikix, line 86: it's no immediate that this comment is about 
`revert_contract_state`

can we make it more explicit?
should we add a TODO to be able to compare `revert_contract_state`?

enitrat: hmmm, I don't really know what would be a good way of testing whether a dict actually is empty... It would require checking that all keys are empty.
But a good heuristic could be to juste verify that the First element is empty but Im not sure it really adds value. Given that it has been instznciated with default, I think we can leave it as is

Eikix: should we simply modify this comment to mention it is about `revert_contract_state`

enitrat: ok!

Eikix (dismissed):

Eikix (approved):
lgtm

Issue #62, opened 2023-08-10
feat: two's complement signed integer division


enitrat:
# Feature Request

The SDIV Opcode performs signed division on u256 values.
There is no such thing as signed division on unsigned integers implemented yet in the core library, and the signed integer types do not go up to 256 bits yet.
**Describe the Feature Request**
Write a math library for signed integer division, using 256-bits values using two's complement
**Describe Preferred Solution**

**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

PR #68, opened 2023-08-10
feat:two's complement division

enitrat:
Implements a math library to perform division on u256s signed values encoded as two's complement.
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: #62 

## What is the new behavior?

- Source & tests for two's complement division.
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/src/tests/test_utils/test_u256_signed_math.cairo
+++ b/src/tests/test_utils/test_u256_signed_math.cairo
@@ -1,6 +1,4 @@
-use kakarot::utils::u256_signed_math::{
-    u256_not, u256_neg, u256_signed_div_rem, SHIFT, ALL_ONES, MAX_U256, TWO_POW_127
-};
+use kakarot::utils::u256_signed_math::{u256_not, u256_neg, u256_signed_div_rem, ALL_ONES, MAX_U256};
 use integer::u256_safe_div_rem;
 use debug::PrintTrait;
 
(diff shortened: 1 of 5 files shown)


src/utils/u256_signed_math.cairo
```diff
@@ -0,0 +1,73 @@
...
+    };
+    let div_positive = div.high < TWO_POW_127;
+    let div = if div_positive {
+        div
+    } else {
+        u256_neg(div)
+    };
+
+    // Compute the quotient and remainder.
+    let (quot, rem) = u256_safe_div_rem(a, div.try_into().unwrap());
```
danilowhk, line 57: Is there any way this would panic, or its not possible because div is of type u256?

src/tests/test_utils/test_u256_signed_math.cairo
```diff
@@ -0,0 +1,145 @@
+use kakarot::utils::u256_signed_math::{
+    u256_not, u256_neg, u256_signed_div_rem, SHIFT, ALL_ONES, MAX_U256, TWO_POW_127
```
danilowhk, line 2: I think we can remove SHIFT and TWO_POW_127, as we are not using it

src/utils/u256_signed_math.cairo
enitrat: Will never happen as the first instruction is a check whether div is 0 or not and the function returns if yes

src/tests/test_utils/test_u256_signed_math.cairo
enitrat: ok!

danilowhk (approved):
Lgtm!

PR #90, opened 2023-08-16
refactor: u256_signed_div

### LINKED ISSUE #89: dev: refactor `u256_signed_div_rem` to take a NonZero divider
The u256_signed_div_rem should remain as generic as possible, and zero cases should be handled prior to calling this function

enitrat:
Refactors `u256_signed_div` to enforce `div !=0` through the type system.
This is open to discussion if you believe that the benefits are not high enough to implement it.
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: Closes #89 

## What is the new behavior?

- Zero case is explicitly handled in SDIV opcode
-
-

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/instructions/stop_and_arithmetic_operations.cairo
+++ b/src/instructions/stop_and_arithmetic_operations.cairo
@@ -90,8 +90,12 @@ impl StopAndArithmeticOperations of StopAndArithmeticOperationsTrait {
         let a = *popped[0];
         let b = *popped[1];
 
-        // Compute the division
-        let (result, _) = u256_signed_div_rem(a, b);
+        let mut result = 0;
+        if b != 0 {
+            // Won't panic since 0 case is handled manually
+            let (q, _) = u256_signed_div_rem(a, b.try_into().unwrap());
+            result = q;
+        }
 
         self.stack.push(result);
     }
--- a/src/instructions/stop_and_arithmetic_operations.cairo
+++ b/src/instructions/stop_and_arithmetic_operations.cairo
@@ -2,9 +2,9 @@
 
 use integer::{
     u256_overflowing_add, u256_overflow_sub, u256_overflow_mul, u256_safe_divmod,
-    u512_safe_div_rem_by_u256
+    u512_safe_div_rem_by_u256, u256_try_as_non_zero
 };
-use traits::TryInto;
+use traits::{TryInto};
 use option::OptionTrait;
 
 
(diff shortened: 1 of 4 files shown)


src/instructions/stop_and_arithmetic_operations.cairo
```diff
@@ -145,8 +145,12 @@ impl StopAndArithmeticOperations of StopAndArithmeticOperationsTrait {
         let a = *popped[0];
         let b = *popped[1];
 
-        // Compute the division
-        let (result, _) = u256_signed_div_rem(a, b);
+        let mut result = 0;
+        if b != 0 {
```
Eikix, line 149: can we try (pseudo code below)

```cairo
let maybe_non_zero_b: Result<NonZero<u256>, Err> = b.try_into();
match maybe_non_zero_b {
Ok(NonZero) => { let (q, _) = division without unwrap
Err()  => 0_u256
}
```

enitrat: I prefer the current way 🤔

enitrat: but perhaps we should avoid `unwrap` like in Rust?

enitrat: actually this is not possible at all because the way done in the pseudo code is not something that can be valid in Cairo

the closes we have is that `try_into` would return an `Option<u128>`

but look at its implementation

```rust
#[panic_with('u128 is 0', u128_as_non_zero)]
fn u128_try_as_non_zero(a: u128) -> Option<NonZero<u128>> nopanic {
    match u128_is_zero(a) {
        IsZeroResult::Zero => Option::None,
        IsZeroResult::NonZero(x) => Option::Some(x),
    }
}

impl U128TryIntoNonZero of TryInto<u128, NonZero<u128>> {
    fn try_into(self: u128) -> Option<NonZero<u128>> {
        Option::Some(u128_as_non_zero(self))
    }
}
```

This will panic - before you even try to unwrap it - if the conversion can't be done. Check this [section](https://book.cairo-lang.org/ch09-01-unrecoverable-errors-with-panic.html?highlight=panic_with#panic_with-attribute) in the Cairo Book

src/instructions/stop_and_arithmetic_operations.cairo
```diff
@@ -127,15 +140,19 @@ impl StopAndArithmeticOperations of StopAndArithmeticOperationsTrait {
...
-            result = r;
-        }
+        let a: u256 = *popped[0];
+        let b: u256 = *popped[1];
+        let n = *popped[2];
+
+        let result: u256 = match u256_try_as_non_zero(n) {
+            Option::Some(nonzero_n) => {
+                // This is more gas efficient than computing (a mod N) + (b mod N) mod N
+                let add_res = u256_wide_add(*popped[0], *popped[1]);
```
Eikix, line 150:
```suggestion
                let sum = u256_wide_add(*popped[0], *popped[1]);
```

src/instructions/stop_and_arithmetic_operations.cairo
```diff
@@ -127,15 +140,19 @@ impl StopAndArithmeticOperations of StopAndArithmeticOperationsTrait {
...
+                let (_, r) = u512_safe_div_rem_by_u256(add_res, nonzero_n);
```
Eikix, line 151:
```suggestion
                let (_, r) = u512_safe_div_rem_by_u256(sum, nonzero_n);
```

Eikix (dismissed):
lgtm, one small suggestion but can be merged already

PR #140, opened 2023-08-20
Refactor ExecutionContext by Separating into Static and Dynamic structs

danilowhk:
### Description
This pull request refactors the execution context within the codebase, focusing on optimizations that target the inherent read-only nature of Cairo's memory. The key challenge addressed is that in Cairo, each update to the execution context requires the entire struct to be recreated, leading to inefficiencies.

#### Changes
1. **Separation into Static and Dynamic Contexts**: To overcome the update inefficiency, the execution context has been separated into `StaticExecutionContext` and `DynamicExecutionContext`. The intention is to place both contexts within a Box pointer, allowing updates to be made to the pointer rather than the entire struct. This approach avoids the need to recopy the whole struct with each update.
    - **Note**: Currently, the `BoxTrait` implementation is facing some issues, so the optimization using Box pointers is pending. This PR lays the groundwork by separating the contexts but does not yet implement the Box pointers.

2. **Introduction of Specific Traits and Structs**: New traits and structs such as `StaticExecutionContextTrait`, `DynamicExecutionContextTrait`, `StaticExecutionContext`, and `DynamicExecutionContext` have been created to facilitate the separation of contexts.

3. **Update of Method Implementations**: Various method implementations have been refined to align with the new structure, enhancing both efficiency and code clarity.

4. **Test Adaptations**: Tests have been updated to match the new structure, ensuring that existing functionalities remain intact.

#### Impact
The refactor represents a vital step towards optimizing the execution context handling in Cairo. While the full optimization awaits the resolution of issues with the `BoxTrait`, this change sets the stage for that enhancement and provides immediate benefits in code maintainability and structure.

#### Review Request
I request a detailed review to ensure that the separation into static and dynamic contexts aligns with the intended design and that the changes are correctly implemented. Feedback on the planned use of Box pointers for future optimization is also welcome.

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: N/A

## What is the new behavior?

-
-
-

- [ ] Yes
- [X] No

- [X] Yes
- [ ] No

### DIFF
--- a/src/context.cairo
+++ b/src/context.cairo
@@ -12,7 +12,7 @@ use option::OptionTrait;
 use starknet::get_caller_address;
 
 /// The call context.
-#[derive(Destruct)]
+#[derive(Destruct, Copy)]
 struct CallContext {
     /// The bytecode to execute.
     bytecode: Span<u8>,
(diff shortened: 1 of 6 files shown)


src/context.cairo
```diff
@@ -60,31 +60,150 @@ impl DefaultCallContextImpl of Default<CallContext> {
...
+    fn evm_address(self: @StaticExecutionContext) -> EthAddress {
+        *self.evm_address
+    }
+
+    #[inline(always)]
+    fn is_read_only(self: @StaticExecutionContext) -> bool {
+        *self.read_only
+    }
+    #[inline(always)]
+    fn bytecode(self: @StaticExecutionContext) -> Span<u8> {
```
enitrat, line 113: this can be removed imo - As it's a part of the CallContext.

src/context.cairo
```diff
@@ -60,31 +60,150 @@ impl DefaultCallContextImpl of Default<CallContext> {
...
+            create_addresses: Default::default(),
+            revert_contract_state: Default::default(),
+            return_data,
+            reverted: false,
+            stopped: false
+        }
+    }
+
+    #[inline(always)]
+    fn destroy_contracts(self: @DynamicExecutionContext) -> Span<EthAddress> {
```
enitrat, line 144: Just a note here - but unlike Rust we can't have `&ref mut`, so we can't really have a setter for these fields. Although, this is fine for a getter

enitrat: On a second note I think these methods should be defined on `ExecutionContext` directly and not on `DynamicExecutionContext`, which is just an code-optimisation pattern.

enitrat: That way we keep the "API" abstracted - our `ExecutionContext` has a specification we conform to, but let's not have too much non-important code

src/evm.cairo
```diff
@@ -21,7 +21,7 @@ fn execute(
         call_context, starknet_address, evm_address, gas_limit, gas_price, returned_data, false
     );
     // Compute the intrinsic gas cost for the current transaction and increase the gas used.
-    ctx.process_intrinsic_gas_cost();
+    // ctx.process_intrinsic_gas_cost();
```
enitrat, line 24: can we remove it totally already? or add TODO

src/interpreter.cairo
```diff
@@ -53,7 +54,10 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
     fn decode_and_execute(ref self: EVMInterpreter, ref context: ExecutionContext) {
         // Retrieve the current program counter.
         let pc = context.program_counter;
-        let bytecode = (@context).call_context.bytecode(); //Note: here, bytecode returns a Span
+        let bytecode = (@context)
```
enitrat, line 57: here we should only be calling `context.call_context.bytecode()`, abstracting the static/dynamic implementations

src/interpreter.cairo
```diff
@@ -43,6 +43,7 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
         // Decode and execute the current opcode.
         self.decode_and_execute(ref context);
         // Check if the execution is complete.
+        // TODO: Check why context.dynamic_context.is_stopped() movess the context variable
```
enitrat, line 46: probably because you access "move" dynamic_context when accessing it directly here. perhaps (@context.dynamic_context) would not cause any problem, to verify

enitrat:
@danilowhk 
Just FYI, this error we had
```shell
thread 'main' panicked at 'Internal error, cycle detected:

DatabaseKeyIndex { group_index: 5, query_index: 16, key_index: 1 }
DatabaseKeyIndex { group_index: 5, query_index: 15, key_index: 2 }
DatabaseKeyIndex { group_index: 5, query_index: 14, key_index: 2 }
```

appears when we do 
```rust
    let mut ctx = ExecutionContextTrait::new(
        call_context, starknet_address, evm_address, gas_limit, gas_price, returned_data, false
    );
 ```
 
 in `evm.cairo`

**Comment by enitrat** [comment]
After further investigation I isolated the bug which comes from a `Derive(Destruct)` instruction. I will report it to the compiler team. 

```rust
#[derive(Destruct)]
struct MinimalStruct {}

#[generate_trait]
impl MinimalStructImpl of MinimalStructTrait {
    fn new() -> MinimalStruct {
        MinimalStruct {}
    }
}

fn execute() {
    let mut ctx = MinimalStructTrait::new();
}


```

src/tests/test_execution_context.cairo
```diff
@@ -69,36 +72,32 @@ fn test_execution_context_new() {
 
     // Then
     let call_context = setup_call_context();
-    assert(execution_context.call_context == call_context, 'wrong call_context');
+    assert(execution_context.static_context.call_context == call_context, 'wrong call_context');
```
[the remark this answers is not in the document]
Eikix: Discussed in Daily: For callcontext, it is the edge case `execution_context.call_context().calldata()`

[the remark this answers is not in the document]
Eikix: But for dynamic and static-> directly implement the getter: execution_context.pc()

src/context.cairo
```diff
@@ -12,7 +12,7 @@ use option::OptionTrait;
 use starknet::get_caller_address;
 
 /// The call context.
-#[derive(Destruct)]
+#[derive(Destruct, Copy)]
```
enitrat, line 15:
```suggestion
#[derive(Drop, Copy)]
```

src/context.cairo
```diff
@@ -60,31 +60,150 @@ impl DefaultCallContextImpl of Default<CallContext> {
 }
 
 
-/// The execution context.
-/// Stores all data relevant to the current execution context.
-#[derive(Destruct)]
-struct ExecutionContext {
+#[derive(Destruct, Copy)]
```
enitrat, line 63:
```suggestion
#[derive(Drop, Copy)]
```

src/context.cairo
```diff
@@ -60,31 +60,150 @@ impl DefaultCallContextImpl of Default<CallContext> {
...
-    //sub_context: Nullable<ExecutionContext>,
+    read_only: bool,
+}
+
+impl DefaultStaticExecutionContext of Default<StaticExecutionContext> {
+    fn default() -> StaticExecutionContext {
+        StaticExecutionContext {
+            call_context: Default::default(),
+            starknet_address: 0.try_into().unwrap(),
+            evm_address: 0.try_into().unwrap(),
```
Eikix, line 76: I think we should implement Default trait for these types rather

Eikix: Or even implement a AddressZero trait method

src/interpreter.cairo
danilowhk: It worked with:

        if !((@context).dynamic_context.is_stopped()) {
            // Execute the next opcode.
            self.run(ref context);
        }
        
But in any case have changed to context.is_stopped

src/context.cairo
danilowhk: Implemented default traits that will default to 0 addresses

src/context.cairo
```diff
@@ -164,74 +216,127 @@ impl ExecutionContextImpl of ExecutionContextTrait {
...
+        *self.dynamic_context.stopped
+    }
+
+    #[inline(always)]
+    fn call_context(self: @ExecutionContext) -> CallContext {
+        *self.static_context.call_context
+    }
+
+    #[inline(always)]
+    fn destroy_contracts(self: @ExecutionContext) -> Span<EthAddress> {
```
enitrat, line 255: `destroy` is an active verb that implies and action. If we're getting a view, maybe`destroyed_contracts?`

```suggestion
    fn destroyed_contracts(self: @ExecutionContext) -> Span<EthAddress> {
```

src/context.cairo
```diff
@@ -164,74 +216,127 @@ impl ExecutionContextImpl of ExecutionContextTrait {
...
+        self.dynamic_context.events.span()
+    }
+
+    #[inline(always)]
+    fn create_addresses(self: @ExecutionContext) -> Span<EthAddress> {
+        self.dynamic_context.create_addresses.span()
+    }
+
+    #[inline(always)]
+    fn revert_contract_state(self: @ExecutionContext) -> @Felt252Dict<felt252> {
```
enitrat, line 270: I think that's the only case where we will need to access it without getters - as a snapshot to a Dict is basically useless

enitrat: you can delete this one

src/interpreter.cairo
```diff
@@ -54,7 +54,7 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
     fn decode_and_execute(ref self: EVMInterpreter, ref context: ExecutionContext) {
         // Retrieve the current program counter.
         let pc = context.program_counter;
-        let bytecode = (@context).call_context.bytecode(); //Note: here, bytecode returns a Span
+        let bytecode = (@context).call_context().bytecode(); //Note: here, bytecode returns a Span
```
enitrat, line 57:
```suggestion
        let bytecode = context.call_context().bytecode(); //Note: here, bytecode returns a Span
```

no need to explicitly @ here, its done by the compiler

src/context.cairo
enitrat: 0 for default sounds good for me tbh

src/context.cairo
Eikix: or selfdestruct_contracts for even more explicit usecase

danilowhk: Went with destroyed_contracts as it felt cleaner

enitrat (approved):
LgTM

Issue #142, opened 2023-08-21
feat: Create gas snapshots to estimate design choice implications on gas usage


enitrat:
# Feature Request

It would make our design choices easier if we could have test execution gas snapshots that we would run after each architectural or implementation change in our codebase to quantify how gas-heavy such a change would be.

**Describe the Feature Request**

**Describe Preferred Solution**
The solution will imply using scarb nightlies, which rely on the latest main of the Cairo compiler, which natively provides gas estimations in the `cairo-test` output. We can capture these estimations when running `scarb cairo-test`, dump them in a JSON file, commit this file, and run a comparison script that outputs how the gas consumption changes for each test - and how it impacts the codebase overall
**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

PR #143, opened 2023-08-21
feat: scripts for gas estimates snapshots

enitrat:
This PR updates scarb to a nightly version, where cairo-test returns gas estimations on test execution. Note that these estimations are not 100% accurate yet, but provide a good guideline when estimating gas impacts of our changes.
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Issue Number: #142

## What is the new behavior?

- Adds a script to generate gas snapshots
- Adds a script to compare previous snapshot with current one, and output a detailed report
- Update scarb to nightly build

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No


gas_snapshots.json
```diff
@@ -0,0 +1,109 @@
+{
+    "kakarot::tests::test_execution_context::test_call_context_new": 62200,
```
Eikix, line 2: should we git ignore it?

.tool-versions
```diff
@@ -1 +1 @@
-scarb 0.6.2
+scarb nightly-2023-08-12
```
Eikix, line 1: coolio

scripts/compare_snapshot.py
```diff
@@ -0,0 +1,66 @@
+import subprocess
```
Eikix, line 1: what to do with those? 
should we add a python "toml" file, like poetry in kakarot cairo0? to enforce a python version and some dependencies?

or run them in the CI exclusively and not impose python to users?

enitrat: I think they're included by default in every python distribs?

imo it should be ran locally before pushing PRs. It's more of a devtool to help in development rather than smth we want to strictly enforce 🤔

gas_snapshots.json
enitrat: no the purpose is that its commited so that we can compare with the previous state each time we modify smth

scripts/compare_snapshot.py
Eikix: could be included as a pre-push hook, so that at least dev (as a warning) knows if they made gas consumption worse

Eikix: anyway! it's here now so we can always think about this later

gas_snapshots.json
enitrat: no, the whole purpose is that it's commited

scripts/compare_snapshot.py
enitrat: > and if worsened, had to be accepted by admin

this is too restrictive for now imo. pre-push would be nice yes

enitrat: https://github.com/kkrt-labs/kakarot-ssj/issues/144

Eikix:
risk of committing this file is that there'll be tons of conflicts for prs, exactly like changelog i think

may be worth to push the thing in gh artifacts rather than in our git history and dl from there for the scripts

enitrat:
> may be worth to push the thing in gh artifacts rather than in our git history and dl from there for the scripts

could you handle this in a subsequent PR ? Not super familiar with that. Also I don't want things to be too complex

Eikix (approved):
lgtm

PR #148, opened 2023-08-22
Feat/error handling execution

### LINKED ISSUE #138: feat: Only the current execution context should be reverted if not enough values are on the Stack
# Feature Request

**Describe the Feature Request**
Most opcodes modify the state of the current execution context. One important thing is that if there are not enough values on the Stack when executing an opcode (e.g. there is only one value on the Stack and you try to execute ADD), then only the current execution context should be reverted, but the whole tx should not be reverted (as per https://www.evm.codes/ indications).

Currently, using `stack.pop_n()` will revert the entire tx if there are not enough values on the stack.

**Describe Preferred Solution**
TBD

**Additional Context**

- [x] Yes
- [ ] No

Eikix:
Light PR to start thinking about result based error management in execution context.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves #138 

## What is the new behavior?

-
-
-

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


src/tests/test_kakarot.cairo
```diff
@@ -5,20 +5,3 @@ use kakarot::evm;
 use kakarot::stack::StackTrait;
 use kakarot::context::CallContext;
 use kakarot::errors;
-
-
```
Eikix, line 9: this test is no longer relevant as execute does not panic anymore, 
and evm::execute does not return an error yet

src/interpreter.cairo
```diff
@@ -42,132 +45,149 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
...
+                if context.reverted() { // TODO: Revert logic
+                }
+                if context.stopped() { // TODO: stopped logic
+                }
+            },
+            Result::Err(error) => {
+                // If an error occurred, revert execution context.
+                // Currently, revert reason is a Span<u8>. 
+                context.revert(u256_to_bytes_array(error.into()).span());
+                self.run(ref context);
```
enitrat, line 66: do we run the next instruction after a revert? Shouldn't we handle revert instead

Eikix: this is to not duplicate code with an revert that may be triggered by REVERT opcode

enitrat: But I don't get it, normally you should revert straight away and not continue the execution any further

Eikix: Fair point

Eikix: ah yeah, my bad though it doenst work here

enitrat (approved):

Issue #14, opened 2023-02-15
feat: implement 0x1A - BYTE opcode


jobez:
---
fork: Frontier
group: Comparison & Bitwise Logic Operations
---

*Index 1 is top of the stack. See [PUSH](https://www.evm.codes/#60).*

## Stack input

0. `i`: byte offset starting from the most significant byte.
1. `x`: 32-byte value.

## Stack output

0. `y`: the indicated byte at the least significant position. If the byte offset is out of range, the result is 0.

## Examples

| * | Input | Output | * | * | Input | Output |
|--:|------:|-------:|--:|--:|------:|-------:|
| `1` | `31` | `0xFF` | * | `1` | `30` | `0xFF` |
| `2` | `0xFF` | | * | `2` | `0xFF00` | |

[Reproduce in playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code='z1~0xFF~31vyyz2w2%200xFF00~30v'~w1%20z%2F%2F%20Example%20y%5CnwyPUSHvyBYTE%01vwyz~_).

## Error cases

The state changes done by the current context are [reverted](#FD) in those cases:
- Not enough values on the stack.

enitrat:
you did byte right @Eikix ?

PR #162, opened 2023-08-22
feat: impl byte

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

solves #14 

## What is the new behavior?

-
-
-

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/src/instructions/comparison_operations.cairo
+++ b/src/instructions/comparison_operations.cairo
@@ -92,17 +92,17 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
     /// Retrieve single byte located at the byte offset of value, starting from the most significant byte.
     fn exec_byte(ref self: ExecutionContext) -> Result<(), EVMError> {
         let popped = self.stack.pop_n(2)?;
-        let value = *popped[0];
-        let offset = *popped[1];
+        let i = *popped[0];
+        let x = *popped[1];
 
         /// If the byte offset is out of range, we early return with 0.
-        if offset > 31 {
+        if i > 31 {
             self.stack.push(0);
             return Result::Ok(());
         }
 
         // Right shift value by offset bits and then take the least significant byte by applying modulo 256.
-        let result = (value / 2.pow((31 - offset) * 8)) % 256;
+        let result = (x / 2.pow((31 - i) * 8)) % 256;
         self.stack.push(result);
         Result::Ok(())
     }
--- a/src/instructions/comparison_operations.cairo
+++ b/src/instructions/comparison_operations.cairo
@@ -6,6 +6,7 @@ use kakarot::errors::STACK_UNDERFLOW;
 use option::{OptionTrait};
 use kakarot::errors::EVMError;
 use result::ResultTrait;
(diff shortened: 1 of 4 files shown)


src/instructions/comparison_operations.cairo
```diff
@@ -87,7 +88,21 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
...
+        let offset = *popped[1];
+
+        /// If the byte offset is out of range, we early return with 0.
+        if offset > 31 {
+            self.stack.push(0);
+            return Result::Ok(());
+        }
+
+        // Right shift value by offset bits and then take the least significant byte by applying modulo 256.
+        let result = (value / 2.pow((31 - offset) * 8)) % 256;
```
enitrat, line 104: Let's introduce a bit manipulation math library.
We can use Herodotus' implementation which is good
https://github.com/HerodotusDev/cairo-lib/blob/main/src/utils/bitwise.cairo

src/instructions/comparison_operations.cairo
```diff
@@ -87,7 +88,21 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
 
     /// 0x1A - BYTE
     /// # Specification: https://www.evm.codes/#1a?fork=shanghai
+    /// Retrieve single byte located at the byte offset of value, starting from the most significant byte.
     fn exec_byte(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let popped = self.stack.pop_n(2)?;
+        let value = *popped[0];
```
enitrat, line 94: didn't you invert these?

enitrat: > Stack input
> 1. i: byte offset starting from the most significant byte.
> 2. x: 32-byte value.

so popped[0] should be offset, popped[1] should be value. maybe it's better to keep the naming from the spec

src/tests/test_instructions/test_comparison_operations.cairo
```diff
@@ -163,3 +163,35 @@ fn test_not_random_uint() {
...
     );
 }
+
+#[test]
+#[available_gas(20000000)]
+fn test_byte_random_u256() {
+    // Given
+    let mut ctx = setup_execution_context();
+    ctx.stack.push(0x08);
+    ctx.stack.push(0xf7ec8b2ea4a6b7fd5f4ed41b66197fcc14c4a37d68275ea151d899bb4d7c2ae7);
```
enitrat, line 173: You should write the tests with the values from EVM.codes, you would've caught the inversion here because the test would've failed

enitrat (changes requested):

src/instructions/comparison_operations.cairo
Eikix: my bad

src/instructions/comparison_operations.cairo
Eikix: should we create another issue for it? to keep things light

enitrat (approved):

PR #164, opened 2023-08-22
refactor: scarb workspaces

### LINKED ISSUE #160: feat: investigate a migration towards Scarb workspaces
# Feature Request

I'd like if possible to migrate towards a Scarb workspace structure, where we can separate:
- EOA codebase (RLP, EOA account abstraction logic, validation)
- Kakarot main smart contracts and Contract Account
- Contract Account codebase 
- Kakarot standalone utils such as https://github.com/Eikix/kakarot/blob/main/src/utils/modexp/modexp_utils.cairo or https://github.com/Eikix/kakarot/blob/main/src/utils/utils.cairo

That way, we try to decouple the three from one another

- [ ] Yes
- [ ] No

enitrat:
Refactor project structures in workspaces with different distinct crates
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Issue Number: Closes #160 

## What is the new behavior?

- Packages are now distinct workspace crates
-
-

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No


crates/eoa/src/eth_transaction.cairo
```diff
@@ -1,4 +1,5 @@
 use array::ArrayTrait;
+use starknet::EthAddress;
```
Eikix, line 2: how come we can use starknet::* even though we didnt write 
```
[dependencies]
alexandria_math.workspace = true
```

in the crate's scarb.toml?

enitrat: You mean dependencies Starknet no? Because it's written at the workspace level so they all inherit

Eikix (approved):
lgtm

crates/eoa/src/eth_transaction.cairo
enitrat: It's written that they're "inherited" so I dont think you need to re declare them

Eikix: then the code snippet they provide is wrong, no?

enitrat: Ahhhh yeah I just realized, it's weird indeed

enitrat: I think it's because we don't have any contracts yet, so it will compiles

crates/evm/src/execution.cairo
ClementWalter, line 1: this seems to be a contract and not a module, similar to the EVM.cairo contract for testing in the previous codebase

ClementWalter (commented):
if we create a `utils` workspace`, we need to use it! let's then move utils, rlp and eth_transaction therein

crates/evm/src/execution.cairo
enitrat: Indeed, for now the contract logic is not implemented (not a priority until final contract syntax)

crates/evm/Scarb.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "evm"
+version = "0.1.0"
+
+# See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest.html
+
+[dependencies]
+starknet.workspace = true
+utils = { path = "../utils" }
```
Eikix, line 9: so beautiful;D

crates/utils/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
Eikix, line 1: i thought target was hoisted at the root?

Eikix:
```All packages share a common output directory, which defaults to a directory named target in the workspace root.
```

enitrat: you can also locally build the package if you `cd` into it

Eikix: ah yes tqtq

Eikix: ah but should that mean we could have a root .gitignore for the whole repo?
since its just one git repo

and the gitignore has wildcard notation for ignoring targets

enitrat: this gitignore was just created automatically by `scarb new`. I guess we could also do it like this

ClementWalter (approved):

PR #141, opened 2023-08-21
Feat/stack nopanic

Eikix:
## Pull Request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

addresses #138 while not completely resolving it.
splitting this issue into 3 PRs for clarity

## What is the new behavior?

-
-
-

## Does this introduce a breaking change?

- [x] Yes
- [ ] No

## Added changes to [`CHANGELOG.md`](../CHANGELOG.md)

- [x] Yes
- [ ] No

PR #182, opened 2023-08-24
bug: add missing `?` operators in methods and `unwrap`s in tests

lambda-0x:
## Pull Request type

- [x] Bugfix

Related: #141 

## What is the new behavior?

- Linked PR missed to add this operator during refactoring, so fixed it.
- Without this it won't propagate stackoverflow error and ignore it silently.
- Inside tests call `unwrap` on methods that can fail
  - due to this `unwrap`s needed to increase gas for few tests

- [x] No

- [x] No

### DIFF
--- a/crates/evm/src/instructions/comparison_operations.cairo
+++ b/crates/evm/src/instructions/comparison_operations.cairo
@@ -56,7 +56,7 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
         let a = *popped[0];
         let b = *popped[1];
         let result = a & b;
-        self.stack.push(result);
+        self.stack.push(result)?;
         Result::Ok(())
     }
 
@@ -73,7 +73,7 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
         let a = *popped[0];
         let b = *popped[1];
         let result = a ^ b;
-        self.stack.push(result);
+        self.stack.push(result)?;
         Result::Ok(())
     }
 
@@ -83,7 +83,7 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
     fn exec_not(ref self: ExecutionContext) -> Result<(), EVMError> {
         let a = self.stack.pop()?;
         let result = ~a;
-        self.stack.push(result);
+        self.stack.push(result)?;
         Result::Ok(())
     }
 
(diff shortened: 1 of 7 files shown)


lambda-0x:
(i don't think changelog is really necessary here, because its doesn't change behaviour of the program. But I can add it if you want)

lambda-0x:
this bug is kinda subtle to spot, fortunately I worked on the PR that this refactors so i was able to spot it. 

Opened an issue in cairo: https://github.com/starkware-libs/cairo/issues/3960

crates/evm/src/instructions/duplication_operations.cairo
```diff
@@ -23,7 +23,7 @@ mod internal {
         let i: u8 = i.into();
 
         let item = context.stack.peek_at((i - 1).into())?;
-        context.stack.push(item);
+        context.stack.push(item)?;
```
Eikix, line 26: nice catch! is this the only place where we forget to add "?"

Eikix (approved):

crates/evm/src/instructions/duplication_operations.cairo
lambda-0x: i just happened to notice this while working on other PR, haven't checked fully PR.

lambda-0x: actually there are other places as well, will add them to this PR.

lambda-0x:
(added `?` to lot more places, and called `unwrap` on methods that return result in tests)

crates/evm/src/instructions/comparison_operations.cairo
```diff
@@ -56,7 +56,7 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
         let a = *popped[0];
         let b = *popped[1];
         let result = a & b;
-        self.stack.push(result);
+        self.stack.push(result)?;
```
enitrat, line 59: I'm too late for the review but this can just be simplified in 
```
self.stack.push(result)
```

No explicit Result::Ok() behind

crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
```diff
@@ -186,8 +186,8 @@ fn test_byte_random_u256() {
 fn test_byte_offset_out_of_range() {
     // Given
     let mut ctx = setup_execution_context();
-    ctx.stack.push(0x01be893aefcfa1592f60622b80d45c2db74281d2b9e10c14b0f6ce7c8f58e209);
-    ctx.stack.push(32_u256);
+    ctx.stack.push(0x01be893aefcfa1592f60622b80d45c2db74281d2b9e10c14b0f6ce7c8f58e209).unwrap();
```
enitrat, line 189: I think we can just do

ctx.stack.push(...)?;

and mark the test function with
https://book.cairo-lang.org/ch09-01-unrecoverable-errors-with-panic.html?highlight=panic_with#panic_with-attribute

Didn't try it yet though

crates/evm/src/instructions/comparison_operations.cairo
enitrat: we will create a dev refactoring issue for this but it's low prio

crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
lambda-0x: that would be helpful if there is only one place it can fail, but in test there are probably many places where function returning `Result` is used so its better to unwrap so we know where it actually failed

PR #191, opened 2023-08-26
feat(evm): add basefee opcode

LucasLvy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #157 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/evm/src/instructions/block_information.cairo
+++ b/crates/evm/src/instructions/block_information.cairo
@@ -70,6 +70,9 @@ impl BlockInformation of BlockInformationTrait {
     /// Get base fee.
     /// # Specification: https://www.evm.codes/#48?fork=shanghai
     fn exec_basefee(ref self: ExecutionContext) -> Result<(), EVMError> {
-        Result::Ok(())
+        // Get the current base fee. (Kakarot doesn't use EIP 1559 
+        // so basefee doesn't really exists there)
+        let base_fee = 0;
+        self.stack.push(base_fee)
     }
 }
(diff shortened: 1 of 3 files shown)


crates/evm/src/instructions/block_information.cairo
```diff
@@ -76,6 +76,9 @@ impl BlockInformation of BlockInformationTrait {
     /// Get base fee.
     /// # Specification: https://www.evm.codes/#48?fork=shanghai
     fn exec_basefee(ref self: ExecutionContext) -> Result<(), EVMError> {
-        Result::Ok(())
+        // Get the current base fee. (Kakarot doesn't use EIP 1559 
+        // so basefee doesn't really exists there)
+        let base_fee = 0;
```
Eikix, line 81: since eip1559 does not exist, base fee should return the gas price (since there's no priority fee)

it should return the current gas price of starknet

Eikix (commented):
can we actually find a way to return the current gas price in sn?

crates/evm/src/instructions/block_information.cairo
LucasLvy: we would need a syscall for that

Eikix: okay i guess then it should return the gasprice from the current execution context then since:

signed gas price of a transaction == basefee + priorityFee 
but priorityfee == 0

so basefee == gasprice imho

LucasLvy: Ok

enitrat:
> can we actually find a way to return the current gas price in sn?

we need a syscall

PR #196, opened 2023-08-26
refacto(evm): improve split word by 2x

LucasLvy:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


crates/utils/src/helpers.cairo
```diff
@@ -75,23 +75,25 @@ fn split_word(mut value: u256, mut len: usize, ref dst: Array<u8>) {
...
             assert(value == 0, 'split_words:value not 0');
-            break ();
+            break;
         }
-
-        let low_part = (value % constants::FELT252_PRIME) % base;
-        dst.append(low_part.try_into().unwrap());
-
-        value = (value - low_part) / 256;
+        dest.append((value % 256).try_into().unwrap());
```
Eikix, line 84: should it be "& 0xFF" to remain consistent with the codebase ?
in order to select the least significant byte

crates/utils/src/helpers.cairo
```diff
@@ -75,23 +75,25 @@ fn split_word(mut value: u256, mut len: usize, ref dst: Array<u8>) {
...
+        value /= 256;
```
Eikix, line 85: should it be .shr ?

Eikix: ah the pr for shr library is ongoing

crates/utils/src/helpers.cairo
```diff
@@ -75,23 +75,25 @@ fn split_word(mut value: u256, mut len: usize, ref dst: Array<u8>) {
...
         len -= 1;
-    };
+    }
+}
+
+/// Splits a u256 into `len` bytes, little-endian, and returns the bytes array.
+fn split_word_little(mut value: u256, mut len: usize) -> Array<u8> {
```
Eikix, line 91: should value not be mut? and at the beginning of the fn we copy it into a mut var (to have consistent fn signature)

crates/utils/src/helpers.cairo
LucasLvy: https://github.com/kkrt-labs/kakarot-ssj/blob/f49c9efdfed7753421aa2186d9c18d315c3bb230/crates/utils/src/helpers.cairo#L160-L161

crates/utils/src/helpers.cairo
LucasLvy: The value is mutable in almost all the functions I think

LucasLvy: and idk if this is your worry but this compiles:
```rs
    2  fn main() {
    1      let c = 2;
    3      foo(c);
    1  }
    2
    3  fn foo(mut a: u32) {
    4      a += 1;
    5  }
    ```

crates/utils/src/helpers.cairo
Eikix: i mean with the newest way to do things

LucasLvy: so we should change everywhere ?

Eikix: @enitrat wdyt

Eikix: let's go with semantics first since optim is negligeable

Eikix (approved):
lgtm, but waiting on @enitrat to merge

PR #197, opened 2023-08-27
feat: implement 0x37 - CALLDATACOPY Opcode

khaeljy:
## Pull Request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Resolves: #123

## What is the new behavior?

-
-
-

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

Issue #185, opened 2023-08-25
dev: move all constant powers (e.g. 256.pow(16) ) variables in utils, evm and memory into their own constants


Eikix:
In the codebase, we have some power function that are applied to constants. let's move them to actual constants!

```[tasklist]
### Tasks
- [ ] Go through the codebase and find where we do apply pow, wrapping_pow, shift_left, shift_right, shift_left_wide, and shift_right_wide to constants
- [ ] Replace all these occurences by their actual value in the right type. You may use Python CLI to compute these values fast, even if they're u256!
```

lambda-0x:
i can take this up

lambda-0x:
i can take this up

lambda-0x:
i can take this up

Eikix:
we should wait a little bit until the codebase is stabilized

Eikix:
Important update: we are gathering some bugs in the [Kakarot v0 codebase](https://github.com/kkrt-labs/kakarot), we need to make sure each issue and each PR in Kakarot-ssj is aware of the lists of known bugs. Look at this link everytime you take an issue and check your issue isn't targeted by a known bug.

Eikix:
Important update: we are gathering some bugs in the [Kakarot v0 codebase](https://github.com/kkrt-labs/kakarot), we need to make sure each issue and each PR in Kakarot-ssj is aware of the lists of known bugs. Look at [this tracking issue](https://github.com/kkrt-labs/kakarot-ssj/issues/324) everytime you take an issue and check your issue isn't targeted by a known bug. Will add this reminder in many places to make sure we keep track of known bugs.

enitrat:
@lambda-0x should be ready now

lambda-0x:
on it

PR #198, opened 2023-08-27
feat: implement 0x35 - CALLDATALOAD Opcode

Quentash:
## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #121

## What is the new behavior?

-
-
-

- [ ] Yes
- [X] No


crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,6 +50,42 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+            if i > 31 {
+                break;
+            }
+
+            // Shift left
+            data_to_load *= 256;
+
+            // Append if not out of bounds
+            if i + offset < calldata_len {
+                data_to_load += (*calldata[i + offset]).into();
```
Eikix, line 81: at some point it'd be nice to have u256::from(slice of bytes)

Eikix (commented):
hi! welcome and thanks for this first pr 🙏🙏
will review a second time tomorrow

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,6 +50,42 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
     /// Push a word from the calldata onto the stack.
     /// # Specification: https://www.evm.codes/#35?fork=shanghai
     fn exec_calldataload(ref self: ExecutionContext) -> Result<(), EVMError> {
+        // Stack input:
+        // 0 - offset: calldata offset of the word we read (32 byte steps).
```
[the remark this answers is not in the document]
Quentash: That was copy/pasted from the 0.X version, but I will remove the () as it brings confusion.
"calldata offset of the word we read" is explanatory engough but I guessed the () wanted to say we read 32 bytes every calldataload call

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,6 +50,42 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+        let offset = self.stack.pop()?.try_into().unwrap();
```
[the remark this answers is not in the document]
Quentash: I could implement a U256ToU32 function in helpers and having it handling a new custom error "Kakarot: into() overflow" or whatever you'd prefer ?

crates/evm/src/instructions/environmental_information.cairo
Quentash: You mean like a new helper function ?
Like u256::from(offset) => returns the u256 bytes from the offset-th most value byte ?
Or like u256::from(offset,len) => returns len u256 bytes from the offset-th most value byte ?

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,7 +51,41 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
     /// Push a word from the calldata onto the stack.
     /// # Specification: https://www.evm.codes/#35?fork=shanghai
     fn exec_calldataload(ref self: ExecutionContext) -> Result<(), EVMError> {
-        Result::Ok(())
+        // Stack input:
+        // 0 - offset: calldata offset of the word we read
+        let offset: u32 = u256_to_u32(self.stack.pop()?)?;
```
enitrat, line 56: for consistency this should be done in the same way as in https://github.com/kkrt-labs/kakarot-ssj/pull/197

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,7 +51,41 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+
+        let calldata = self.call_context().call_data();
+        let calldata_len = calldata.len();
+
+        // Early return if the offset is beyond calldata
+        if offset >= calldata_len.into() {
+            self.stack.push(0);
+            return Result::Ok(());
```
enitrat, line 64:
```rust
            return self.stack.push(0)
```

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,7 +51,41 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+        }
+
+        // Init data to load
+        let mut data_to_load: u256 = 0;
+        data_to_load += (*calldata[0 + offset]).into();
+
+        let mut i: usize = 1;
+        loop {
+            if i > 31 {
```
enitrat, line 73: use equality checks instead of range checks, it's cheaper in Cairo (because the instruction set is limited and doesn't contain native range checks )

```suggestion
            if i == 32 {
```

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,7 +51,41 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+                break;
+            }
+
+            // Shift left
+            data_to_load *= 256;
```
enitrat, line 78: note - related to https://github.com/kkrt-labs/kakarot-ssj/issues/185 for future reference

enitrat: @quentash any risk of this operation overflowing? In that case, do we want to wrap around? If yes you can import `u256_overflow_mul` from `integer` and use it instead

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,7 +51,41 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+
+            // Append if not out of bounds
+            if i + offset < calldata_len {
```
enitrat, line 81: imo `i` should be initialized at offset directly to make it simpler

crates/evm/src/instructions/environmental_information.cairo
enitrat: I think the behavior implemented here is just `load_word` from `utils::helpers`. If so we can just replace that part.

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -41,3 +43,117 @@ fn test__exec_callvalue() {
...
+    // When
+    ctx.exec_calldataload();
+
+    // Then
+    let result: u256 = ctx.stack.pop().unwrap();
+    let mut results: Array<u8> = u256_to_bytes_array(result);
+
+    let mut i: u32 = 0;
+    loop {
+        if i > 31 {
```
enitrat, line 68:
```suggestion
        if i == 32 {
```

not very important given that it's a test but it's a good habit. We know that `i` is always initialized under the limit and increased only by 1

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -41,3 +43,117 @@ fn test__exec_callvalue() {
...
+            break;
+        }
+
+        if i + offset < call_data_len {
```
enitrat, line 72: actually, instead of doing this, you could just slice call_data starting from`offset` for 32 elements, and use `SpanPartialEq` to compare both arrays.

crates/evm/src/instructions/environmental_information.cairo
Quentash: No, since we do the operation is always and only made 31 times, it would never overflow.

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -50,7 +51,41 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
     fn exec_calldataload(ref self: ExecutionContext) -> Result<(), EVMError> {
-        Result::Ok(())
+        // Stack input:
+        // 0 - offset: calldata offset of the word we read
+        let offset: u32 = u256_to_u32(self.stack.pop()?)?;
+
+        let calldata = self.call_context().call_data();
+        let calldata_len = calldata.len();
+
+        // Early return if the offset is beyond calldata
```
enitrat, line 61:
```suggestion
        // All bytes after the end of the calldata are set to 0.
```

crates/evm/src/tests/test_instructions/test_environment_information.cairo
enitrat: > verify that len isn't >= 32

I would rather not if that condition is never triggered because we don't call that function with len >=32 in the codebase - given that this function interacts with memory we want it as optimized as possible


> modify load_word for it to accepts an offset

you wouldn't need to, you could just do something like

```rust
 let bytes_len = cmp::min(32, calldata_len - offset);
 let slice = calldata.slice(offset, bytes_len);
```
and then call the load_word function which would result in the same thing as starting with an offset.

the only problem remaining here is appending it with zeroes. Which we can be easily overcomed.

Here is my proposal

```rust

        // Copy len_bytes from calldata
        let len_bytes = cmp::min(32, calldata_len - offset);
        let slice = calldata.slice(offset, len_bytes);
        let mut loaded_data = load_word(len_bytes, slice);

        // fill the rest of the word with zeros
       //TODO: optimize once we have dw-based exponentiation
        let mut i = 32 - len_bytes;
        loop {
            if i == 0 {
                break;
            }
            loaded_data *= 256;
            i -= 1;
        };

        self.stack.push(loaded_data)
  ```


fwiw this is equivalent to doing `loaded_data *= 256.pow(i)` - but it's slightly more expensive

**Comment by ClementWalter** [comment]
> fwiw this is equivalent to doing `loaded_data *= 256.pow(i)` - but it's slightly more expensive 

are you saying that native `core::integers::u256::pow` is more expensive than this loop? is it because it makes some extra checks?
In terms of effectivness, should we instead 
- pad first the `sliced` array with `[0, ..]`
- do `load_word`, and no `loop` nor `pow`

File: crates/evm/src/tests/test_instructions/test_environment_information.cairo [hunk 12]
```diff
@@ -196,3 +196,111 @@ fn test_returndatasize() {
     assert(ctx.stack.len() == 1, 'stack should have one element');
     assert(ctx.stack.pop().unwrap() == size.into(), 'wrong returndatasize');
 }
+
+#[test]
+#[available_gas(20000000)]
+fn test_calldata_load() {
+    // Given
+    let mut ctx = setup_execution_context();
+    let call_data: Span<u8> = ctx.call_context().call_data();
```
ClementWalter, line 205: it's `calldata` and not `call_data`

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -196,3 +196,111 @@ fn test_returndatasize() {
...
+    let call_data_len = call_data.len();
```
ClementWalter, line 206: here as well

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -196,3 +196,111 @@ fn test_returndatasize() {
...
+
+    // When
+    ctx.exec_calldataload();
+
+    // Then
+    let result: u256 = ctx.stack.pop().unwrap();
+
+    let mut expected = load_word(call_data_len, call_data);
+
+    let mut i = 32 - call_data_len;
```
ClementWalter, line 220: you use the same logic in the function and in the test!

I guess that here it'd be better to use hard coded expexted results, there are fyi 2 examples in the playgroud

crates/utils/Scarb.toml
```diff
@@ -5,3 +5,4 @@ version = "0.1.0"
 # See more keys and their definitions at https://docs.swmansion.com/scarb/docs/reference/manifest.html
 
 [dependencies]
+evm = { path = "../evm" }
```
ClementWalter, line 8: why did you add this?

Quentash: Probably forgot to remove it after tests. Now removed!

enitrat:
> are you saying that native core::integers::u256::pow is more expensive than this loop? is it because it makes some extra checks?

I'm not sure where you got this one, it doesn't exist and this is our own implementation of `pow` in `utils::math`. But yes, it's more expensive.

> In terms of effectivness, should we instead

pad first the sliced array with [0, ..]
do load_word, and no loop nor pow

will check

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -196,3 +196,111 @@ fn test_returndatasize() {
     assert(ctx.stack.len() == 1, 'stack should have one element');
     assert(ctx.stack.pop().unwrap() == size.into(), 'wrong returndatasize');
 }
+
+#[test]
+#[available_gas(20000000)]
+fn test_calldata_load() {
```
enitrat, line 202:
```suggestion
fn test_calldataload() {
```

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -196,3 +196,111 @@ fn test_returndatasize() {
...
+        }
+        expected *= 256;
+        i -= 1;
+    };
+    assert(expected == result, 'wrong results');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_calldata_load_with_offset() {
```
enitrat, line 233:
```suggestion
fn test_calldataload_with_offset() {
```

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -196,3 +196,111 @@ fn test_returndatasize() {
...
+        }
+        expected *= 256;
+        i -= 1;
+    };
+    assert(expected == result, 'wrong results');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_calldata_load_with_offset_beyond_calldata() {
```
enitrat, line 266:
```suggestion
fn test_calldataload_with_offset_beyond_calldata() {
```

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -196,3 +196,111 @@ fn test_returndatasize() {
...
+
+    // Then
+    let result: u256 = ctx.stack.pop().unwrap();
+    assert(result == 0, 'result should be 0');
+}
+
+
+#[test]
+#[available_gas(20000000)]
+fn test_calldata_load_with_offset_conversion_error() {
```
enitrat, line 287:
```suggestion
fn test_calldataload_with_offset_conversion_error() {
```

Quentash:
Corrected all the mentioned mistakes.
Added hardcoded expected results.
Added create_context_with_calldata() helper function to be able to feed desired data into the context calldata.
Removed scarb.toml unnecessary entry.

ClementWalter:
minor changes plz

Quentash:
Minor changes done.

enitrat (approved):
lgtm

Issue #190, opened 2023-08-25
refactor: refactor opcodes that end with `stack.push(some_value)?;` and `Result::Ok(())` into `stack.push(some_value)`


Eikix:
Currently, when inside a function that returns a result, but manipulates result-based functions itself, we often bubble up errors and unwrap results using "?" operator.

This is useful if we want to do something with the result, if we just want to propagate it as is, e.g. `stack.push(some_value)?;` is on the pre-last line, then we could just do:
`stack.push(some_value)` ,thus returning the result itself!

lambda-0x:
i can work on this

PR #201, opened 2023-08-27
refactor: simplify returns

lambda-0x:
## Pull Request type

Resolves: #190 

## What is the new behavior?

- simplifies the returns

- [x] No

### DIFF
--- a/crates/evm/src/instructions/comparison_operations.cairo
+++ b/crates/evm/src/instructions/comparison_operations.cairo
@@ -56,8 +56,7 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
         let a = *popped[0];
         let b = *popped[1];
         let result = a & b;
-        self.stack.push(result)?;
-        Result::Ok(())
+        self.stack.push(result)
     }
 
     /// 0x17 - OR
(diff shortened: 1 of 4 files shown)


enitrat:
Can you also do the same for the `decode_and_execute` function?

lambda-0x:
image

need to use `return` instead of `?;` , is that fine?

enitrat:
> need to use return instead of ?; , is that fine?

yep, this is intended, this is because you want to propagate the error to the calling function

enitrat:
> also there is a bug in that function, we should be never reaching this (Unknown opcode) part of code because u8 max value is 255 and we are covering all possible values in if branches.

there are less than 255 opcodes actually, so you could end up trying to hit an unexisting value, which would result in calling an unknown opcode. Even if max_u8 is 255

lambda-0x:
(made suggested changes and rebased branch to latest main)

lambda-0x:
(CI failure unrelated to PR, might be related to: https://github.com/kkrt-labs/kakarot-ssj/pull/217)

enitrat:
> (CI failure unrelated to PR, might be related to: https://github.com/kkrt-labs/kakarot-ssj/pull/217)

yes, I think the action doesn't have the required permissions, but if tests pass it's ok

enitrat (approved):
lgtm

PR #234, opened 2023-08-30
fix: integer underflow

lambda-0x:
## Pull Request type

- [x] Bugfix

Resolves: #233 

## What is the new behavior?

-
-
-

- [x] No

### DIFF
--- a/crates/evm/src/instructions/environmental_information.cairo
+++ b/crates/evm/src/instructions/environmental_information.cairo
@@ -88,14 +88,21 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
 
         let bytecode: Span<u8> = self.call_context().bytecode();
 
-        let slice_size = if (offset + size > bytecode.len()) {
-            bytecode.len() - offset
+        // No. of bytes that are read from bytecode
+        let slice_size = if (offset > bytecode.len()) {
+            // When offset is larger then bytecode len, no bytes to be written from bytecode
+            0
         } else {
+            // How many bytes to be read from memory
+            let size = if (offset + size > bytecode.len()) {
+                bytecode.len() - offset
+            } else {
+                size
+            };
+            let mut data_to_copy: Span<u8> = bytecode.slice(offset, size);
+            self.memory.store_n(data_to_copy, dest_offset);
             size
         };
-
-        let data_to_copy: Span<u8> = bytecode.slice(offset, slice_size);
-        self.memory.store_n(data_to_copy, dest_offset);
 
         // For out of bound bytes, 0s will be copied.
         if (slice_size < size) {
--- a/crates/evm/src/instructions/environmental_information.cairo
+++ b/crates/evm/src/instructions/environmental_information.cairo
@@ -94,14 +94,14 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
             0
         } else {
             // How many bytes to be read from memory
-            let size = if (offset + size > bytecode.len()) {
+            let mem_size = if (offset + size > bytecode.len()) {
                 bytecode.len() - offset
             } else {
                 size
             };
-            let mut data_to_copy: Span<u8> = bytecode.slice(offset, size);
+            let mut data_to_copy: Span<u8> = bytecode.slice(offset, mem_size);
             self.memory.store_n(data_to_copy, dest_offset);
-            size
+            mem_size
         };
 
         // For out of bound bytes, 0s will be copied.
--- a/crates/evm/src/instructions/environmental_information.cairo
+++ b/crates/evm/src/instructions/environmental_information.cairo
@@ -88,23 +88,18 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
 
         let bytecode: Span<u8> = self.call_context().bytecode();
 
-        // No. of bytes that are read from bytecode
-        let slice_size = if (offset > bytecode.len()) {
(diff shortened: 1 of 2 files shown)


crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -132,6 +132,12 @@ fn test_codecopy_with_out_of_bound_bytes() {
     test_codecopy(32, 0, 8);
 }
 
+#[test]
+#[available_gas(20000000)]
+fn test_codecopy_with_out_of_bound_offset() {
+    test_codecopy(0, 0xFFFFFFFE, 2);
```
lambda-0x, line 138: currently i don't see a way to test case in which `offset + size` > usize because `Array<T>` and `Span<T>` implements index view only on `usize`

khaeljy: See https://github.com/kkrt-labs/kakarot-ssj/issues/233#issuecomment-1699482972

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -88,15 +88,22 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
 
         let bytecode: Span<u8> = self.call_context().bytecode();
 
-        let slice_size = if (offset + size > bytecode.len()) {
-            bytecode.len() - offset
+        // No. of bytes that are read from bytecode
+        let slice_size = if (offset > bytecode.len()) {
```
enitrat, line 92:
```rust

        // Slice of the size: 
        // - size if (offset + size) in bounds, otherwise
        // - (bytecode_len - offset) if bytecode_len > offset, othewise
        // - 0
        let slice_size = if (offset + size > bytecode.len()) {
            match integer::u32_overflowing_sub(bytecode.len(), offset) {
                Result::Ok(result) => result,
                Result::Err(_) => 0
            }
        } else {
            size
        };
        
```

here's a more optimized way of achieving the same result

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -88,15 +88,22 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+            // When offset is larger then bytecode len, no bytes to be written from bytecode
+            0
         } else {
+            // How many bytes to be read from memory
+            let size = if (offset + size > bytecode.len()) {
```
enitrat, line 97: this line can be simplified directly to

`else if (offset + size > bytecode.len()) {`

crates/evm/src/instructions/environmental_information.cairo
enitrat: let's keep it simple with something easier to review

```rust
        let slice_size = if offset > bytecode.len() {
            0
        } else if (offset + size > bytecode.len()) {
            bytecode.len() - offset
        } else {
            size
        };
```

enitrat:
curious why CI returns no gas changes here, when there clearly are 🤔

crates/evm/src/instructions/environmental_information.cairo
lambda-0x: i tried doing that first but when `offset > bytecode.len()` we cannot call `bytecode.slice` to write to memory because it will panic due to offset being out of index (we don't even need to).

So i chose to handle that by moving memory copy related logic in the else part.

crates/evm/src/instructions/environmental_information.cairo
lambda-0x: cannot because there are two cases 1 where `offset > bytecode.len()` other when `offset <= bytecode.len()` (this internally has two branch). see: https://github.com/kkrt-labs/kakarot-ssj/pull/234#discussion_r1310753678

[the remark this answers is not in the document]
lambda-0x: good point, will update it to `mem_size`

crates/evm/src/instructions/environmental_information.cairo
enitrat: I see - then let's do it as 

```rust

    fn exec_codecopy(ref self: ExecutionContext) -> Result<(), EVMError> {
        let popped = self.stack.pop_n(3)?;
        let dest_offset: u32 = Into::<u256, Result<u32, EVMError>>::into((*popped[0]))?;
        let offset: u32 = Into::<u256, Result<u32, EVMError>>::into((*popped[1]))?;
        let size: u32 = Into::<u256, Result<u32, EVMError>>::into((*popped[2]))?;

        let bytecode: Span<u8> = self.call_context().bytecode();

        let copied: Span<u8> = if offset > bytecode.len() {
            // When offset is larger then bytecode len, no bytes to be written from bytecode
            array![].span()
        } else if (offset + size > bytecode.len()) {
            // prevent size OOB errors
            bytecode.slice(offset, bytecode.len() - offset)
        } else {
            // size is in [0, bytecode.len() - offset]
            bytecode.slice(offset, size)
        };

        self.memory.store_n(copied, dest_offset);

        if (copied.len() < size) {
            let mut out_of_bounds_bytes: Array<u8> = ArrayTrait::new();
            loop {
                if (out_of_bounds_bytes.len() + copied.len() == size) {
                    break;
                }

                out_of_bounds_bytes.append(0);
            };

            self.memory.store_n(out_of_bounds_bytes.span(), dest_offset + copied.len());
        }

        Result::Ok(())
    }
    ```
    
    Comments are just indicative, you can get rid of them

Issue #239, opened 2023-09-01
dev: implement Into<bool,u256> trait


enitrat:
which allows us to naturally do
```rust
let result:u256 = a.slt(b).into();
```
instead of 
```rust
let result = if (a.slt(b)) {
            1
        } else {
            0
        };
```

khaeljy:
@enitrat Can I do it?

enitrat:
yes, please implement it under utils::traits

khaeljy:
Perhaps we should implement `Into<bool, u8>` instead, which would allow conversion to any integer size?

`BoolIntoU8::into(boolean_value).into()`

enitrat:
I said Into<bool,u32> so that we could directly push the values to the stack, but I meant `Into<bool,u256>`. of course we could do all integer types but we don't need it right now

PR #240, opened 2023-09-01
dev: implement Into<bool,u256> trait

khaeljy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #239 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


crates/utils/src/traits.cairo
```diff
@@ -21,3 +21,44 @@ impl ContractAddressDefault of Default<ContractAddress> {
         0.try_into().unwrap()
     }
 }
+
+impl BoolIntoU8 of Into<bool, u8> {
+    fn into(self: bool) -> u8 {
```
enitrat, line 26: there is a special libfunc defined in the compiler called `bool_to_felt252` which converts a bool to a felt252. I don't think it has any impact on performance or anything, but I would rather use the bool -> felt252 -> uX  conversion rather than this specific if / else case

enitrat: After researching a bit it seems that intos/unwraps have a tiny associated overhead cost. One elegant solution would be the following, avoiding code duplication while maximizing perfs

```rust
use math::Oneable;
/// ...
impl BoolIntoNumeric<T, impl TZeroable: Zeroable<T>, impl TOneable: Oneable<T>> of Into<bool, T> {
    #[inline(always)]
    fn into(self: bool) -> T {
        if self {
            Oneable::<T>::one()
        } else {
            Zeroable::<T>::zero()
        }
    }
}
```

enitrat (changes requested):

crates/utils/src/traits.cairo
khaeljy: Oh great, that's a better solution!

enitrat (approved):
lgtm 💯

Issue #15, opened 2023-02-15
feat: implement 0x1B - SHL opcode


jobez:
---
fork: Constantinople
group: Comparison & Bitwise Logic Operations
---

*Index 1 is top of the stack. See [PUSH](https://www.evm.codes/#60).*

## Notes

Shift the bits towards the most significant one. The bits moved after the 256th one are discarded, the new bits are set to 0.

## Stack input

0. `shift`: number of bits to shift to the left.
1. `value`: 32 bytes to shift.

## Stack output

0. `value << shift`: the shifted value. If `shift` is bigger than 255, returns 0.

## Examples

| * | Input | Output |
|--:|------:|-------:|
| `1` | `1` | `2` |
| `2` | `1` | |

| * | Input | Output |
|--:|------:|-------:|
| `1` | `4` | `0xF000000000000000000000000000000000000000000000000000000000000000` |
| `2` | `0xFF00000000000000000000000000000000000000000000000000000000000000` | |

[Reproduce in playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code='y1z1z1swwy2v32%200xFFuuuuutz4s'~tttzv1%20y%2F%2F%20Example%20w%5CnvwPUSHu~~t00swSHL%01stuvwyz~_).

## Error cases

The state changes done by the current context are [reverted](#FD) in those cases:
- Not enough values on the stack.

trbutler4:
dibs

danilowhk:
Assigned @trbutler4

PR #241, opened 2023-09-02
 feat: 0x1B - SHL opcode

trbutler4:
implemented the shift left operator
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #15 

## What is the new behavior?
shift left operator implemented using utils::math wrapping shl 

-
-
-

- [ ] Yes
- [x] No


crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
```diff
@@ -213,6 +213,43 @@ fn test_exec_gt_true() {
...
+    // Given 
+    let mut ctx = setup_execution_context();
+    ctx.stack.push(0xff00000000000000000000000000000000000000000000000000000000000000).unwrap();
+    ctx.stack.push(4_u256).unwrap();
+
+    // When 
+    ctx.exec_shl();
+
+    // Then
+    assert(ctx.stack.len() == 1, 'stack should have one element');
```
enitrat, line 228: you can group both assertions in the same test case. 
For a single test case with common test data, we must verify two different assertions, so it doesn't really make sense to separate the assertions

enitrat (commented):
Can you introduce a test case for `shift>255` as well?

crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
trbutler4: i misunderstood what you meant on the previous pr when you asked me to separate into two tests with explicit naming. Put both assertions back

trbutler4: also added test for shift > 255

trbutler4: also added test for shift > 255

crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
```diff
@@ -213,6 +213,50 @@ fn test_exec_gt_true() {
     assert(ctx.stack.peek().unwrap() == 1, 'stack top should be 1');
 }
 
+#[test]
+#[available_gas(20000000)]
+fn test_exec_shl_wrapping() {
```
enitrat, line 218:
```suggestion
fn test_exec_shl() {
```

crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
```diff
@@ -213,6 +213,50 @@ fn test_exec_gt_true() {
...
+            .stack
+            .peek()
+            .unwrap() == 0xf000000000000000000000000000000000000000000000000000000000000000,
+        'stack top should be 0xf00000...'
+    );
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_exec_shl_shift_too_large() {
```
enitrat, line 240:
```suggestion
fn test_exec_shl_wrapping() {
```

enitrat (dismissed):

enitrat (approved):

crates/evm/src/instructions/comparison_operations.cairo
```diff
@@ -116,7 +116,17 @@ impl ComparisonAndBitwiseOperations of ComparisonAndBitwiseOperationsTrait {
     /// 0x1B - SHL
     /// # Specification: https://www.evm.codes/#1b?fork=shanghai
     fn exec_shl(ref self: ExecutionContext) -> Result<(), EVMError> {
-        Result::Ok(())
+        let popped = self.stack.pop_n(2)?;
+        let shift = *popped[0];
+        let val = *popped[1];
+
+        // if shift is bigger than 255 return 0
```
ClementWalter, line 123: is this only to save gas? if so, could you add a comment

ClementWalter:
just a questino regarding the early return

trbutler4:
> just a questino regarding the early return

the reason for the early return was just to match the evm opcode specification: https://www.evm.codes/?fork=shanghai

where it says: value << shift: the shifted value. If shift is bigger than 255, returns 0.

should note that the shr opcode includes the same specification, but does not have an early return. Maybe needs to be updated.

Issue #254, opened 2023-09-02
feat: implement 0x58 - PC Opcode


khaeljy:
# Feature Request

| Since | Group |
| --- | --- |
| Frontier| Stack Memory Storage and Flow Operations |

*Index 1 is top of the stack. See [PUSH](https://www.evm.codes/#60).*

## Notes

The program counter (PC) is a byte offset in the deployed [code](https://www.evm.codes/about). It indicates which instruction will be executed next. When an [ADD](https://www.evm.codes/#01) is executed, for example, the PC is incremented by 1, since the instruction is 1 byte. The [PUSH](https://www.evm.codes/#60) instructions are bigger than one byte, and so will increment the counter accordingly.

## Stack output

1. `counter`: PC of this instruction in the current program.

## Example

[See in playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code='~0x~1xJUMPDESTzesq2x~3xPUSH1%201%20l4x~6%20%7Bpreviouminstructionmtakem2%20bytes%7D'~PCwwwlz%20%2F%2F%20Offx%5Cnw%20%20qt%20ms%20lzseq%01lmqwxz~_).

## Error cases

The state changes done by the current context are [reverted](https://www.evm.codes/#FD) in those cases:
- Not enough gas.
- Stack overflow.

lambda-0x:
can i work on this?

Eikix:
Sure, we're trying something out right now, which is to start by working on the test suite first to make sure the opcode is well understood first

Eikix:
You'd have to submit a PR for https://github.com/kkrt-labs/kakarot-ssj/issues/255 first

lambda-0x:
@Eikix actually for this implementation is just a single line hence included it in same PR. will keep in mind next time.

Eikix:
i see

Issue #255, opened 2023-09-02
feat: implement the test suite for 0x58 - PC opcode


khaeljy:
```[tasklist]
### Tasks
- [ ] Create the trait for the PC opcode category
- [ ] Create the PC opcode function, write it only as a panic_with_felt252('PC not implement yet')
- [ ] Create the test suite for PC, taking Geth's examples
- [ ] Create the test suite for PC taking Kakarot cairo0 examples
- [ ] Create the test suite for PC taking the evm.codes examples
```

PR #264, opened 2023-09-02
feat: implement PC opcode

lambda-0x:
## Pull Request type

- [x] Feature

Resolves: #254 #255 

## What is the new behavior?

- implements `PC` opcode
- kakarot on cairo0 just increments the PC manually, which I don't think will be very useful
- i didn't find tests related PC specifically in geth.
- evm.codes also dont specify any examples.

- [x] No

### DIFF
--- a/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
+++ b/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
@@ -26,6 +26,49 @@ fn test_pc_basic() {
     assert(ctx.stack.len() == 1, 'stack should have one element');
     assert(ctx.stack.pop().unwrap() == 0, 'PC should be 0');
 }
-// TODO: Test that PC updates properly when instructions are executed
 
 
+#[test]
+#[available_gas(20000000)]
+fn test_pc_gets_updated_properly_1() {
+    // Given
+    let mut ctx = setup_execution_context();
+
+    // When
+    ctx.set_pc(9000);
+    ctx.exec_pc();
+
+    // Then
+    assert(ctx.stack.len() == 1, 'stack should have one element');
+    assert(ctx.stack.pop().unwrap() == 9000, 'updating PC failed');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_pc_gets_updated_properly_2() {
+    // Given
+    let mut ctx = setup_execution_context();
+
+    // When
+    ctx.set_pc(42);
+    ctx.exec_pc();
+
+    // Then
+    assert(ctx.stack.len() == 1, 'stack should have one element');
+    assert(ctx.stack.pop().unwrap() == 42, 'updating PC failed');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_pc_gets_updated_properly_3() {
+    // Given
+    let mut ctx = setup_execution_context();
+
+    // When
+    ctx.set_pc(7);
+    ctx.exec_pc();
+
+    // Then
+    assert(ctx.stack.len() == 1, 'stack should have one element');
+    assert(ctx.stack.pop().unwrap() == 7, 'updating PC failed');
+}
--- a/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
+++ b/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
@@ -1,0 +1,31 @@
+use evm::instructions::{MemoryOperationTrait, EnvironmentInformationTrait};
+use evm::tests::test_utils::{
+    setup_execution_context, setup_execution_context_with_bytecode, evm_address, callvalue
+};
+use evm::stack::StackTrait;
+use evm::memory::{InternalMemoryTrait, MemoryTrait};
+use option::OptionTrait;
+use starknet::EthAddressIntoFelt252;
+use utils::helpers::{EthAddressIntoU256, u256_to_bytes_array};
+use evm::errors::{EVMError, TYPE_CONVERSION_ERROR};
+use evm::context::{
+    ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionContextDestruct, CallContextTrait,
+};
+
+
+#[test]
+#[available_gas(20000000)]
+fn test_pc_basic() {
+    // Given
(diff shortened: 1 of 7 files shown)


crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,31 @@
...
+
+    // When
+    ctx.exec_pc();
+
+    // Then
+    assert(ctx.stack.len() == 1, 'stack should have one element');
+    assert(ctx.stack.pop().unwrap() == 0, 'PC should be 0');
+}
+// TODO: Test that PC updates properly when instructions are executed
+
```
lambda-0x, line 30: i am not sure if it would be a good idea to add a test for each instruction manually, instead we can make sure that each instruction is updating the PC as expected. 

both kakarot in cairo 0 and geth doesn't seem it test PC opcode manually.

lambda-0x: we need to call `decode_and_execute` with proper context to actually update the PC, any ideas on how to test all the cases cleanly?

khaeljy: Maybe wait until issue #250 is implemented, but that wouldn't really be a unit test...

lambda-0x: i am not sure how that would be useful here?

khaeljy: You can execute JUMP to a given index, and then you can check whether the new PC has the right value.

khaeljy: Or simply use `ctx.program_counter = GIVEN_INDEX`

lambda-0x: what would that be testing though? 

the point of testing is that you have an `initial setup` -> `do some processing` -> `assert on the resulting` state.

If we set PC manually it won't test that running an instruction produces correct PC (maybe checking for correctness maybe out of scope of this PR but i just wanted to point that out)

lambda-0x: does this PR needs to check that all opcodes results in a proper PC value?

cc: @enitrat @Eikix

khaeljy: ah ok sorry, I didn't quite get your point.

ClementWalter: there is actually a test in kakarot 0, see `test__exec_pc__should_update_after_incrementing`

regarding what to test from a unit testing perspective:
- the `exec_pc` opcode is supposed to put whatever the PC value is in the Stack, so setting the PC to something and asserting that the stack got the same value is fine
- the `exec_pc` is not responsible for actually incrementing the PC, so we shall not test that PC is incremented properly in a PC unit test
- as you pointed out, `decode_and_execute` is responsible for incrementing the PC, and asserting this increment is done properly will be done in a `decode_and_execute` test

ClementWalter (changes requested):
please just do the same test as in kakarot (see https://github.com/ClementWalter/kakarot/blob/b6619adb6092984f4c6b4fa512ebf07ec8d47299/tests/src/kakarot/instructions/test_memory_operations.cairo#L20).

ie put a value and test that PC returns it, for several values

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,31 @@
...
+    ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionContextDestruct, CallContextTrait,
+};
+
+
+#[test]
+#[available_gas(20000000)]
+fn test_pc_basic() {
+    // Given
+    let mut ctx = setup_execution_context();
+
```
enitrat, line 21: can you add an inlined `set_pc` method to the 
 `ExecutionContext` struct implementation, and manually set this value in the test to assert that the value popped is not 0 (default value)

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
enitrat: agreed with clement

lambda-0x: i thought because we where setting the value manually the test doesn't make much sense, but yeah when you think about that way it make sense. will add.

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
lambda-0x: done

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,74 @@
...
+    ctx.exec_pc();
+
+    // Then
+    assert(ctx.stack.len() == 1, 'stack should have one element');
+    assert(ctx.stack.pop().unwrap() == 9000, 'updating PC failed');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_pc_gets_updated_properly_2() {
```
enitrat, line 48: only 1 test of this kind is enough

ClementWalter:
> only 1 test of this kind is enough 

we dramatically miss a way to parametrize the tests, in cairo0 we had tests for PC in `range(1, 15)`

enitrat:
> we dramatically miss a way to parametrize the tests, in cairo0 we had tests for PC in range(1, 15)

Agreed but here the logic is fairly simple and doesn't present any other edge cases than pc > range bound. I wouldn't duplicate test for no reason

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
lambda-0x: thought so but because clement told to test few cases, i added it anyway.

enitrat (approved):
lgtm

ClementWalter (approved):

PR #265, opened 2023-09-02
feat: implement the test suite for MLOAD opcode

khaeljy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #245

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,63 @@
...
+#[should_panic(expected: ('MLOAD not implement yet',))]
+fn test_exec_mload() {
+    let data: Array<(u256, u256, u256, u32)> = array![
+        // (memory_value, input, expected_value, expected_memory_size)
+
+        // Geth: no test ?!
+
+        // Kakarot cairo0
+        (0x0000000000000000000000000000000000000000000000000000000000000001, 0, 0x1, 32),
+        // This is the test case in kakarot cairo0, but playground say, the memory size should be 64 ???
```
khaeljy, line 20: I'm not sure about the expected memory size, the test in kakarot cairo0 and evm.codes playground don't say the same thing.

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,63 @@
...
+        (
+            0x0000000000000000000000000000000000000000000000000000000000000001,
+            16,
+            0x100000000000000000000000000000000,
+            16 + 32
```
khaeljy, line 25: I'm not sure about the expected memory size, the test in kakarot cairo0 and evm.codes playground don't say the same thing.

khaeljy: The memory load() function seems to add 32 to the offset.

```
fn load(ref self: Memory, offset: usize) -> (u256, usize) {
        let gas_cost = self.ensure_length(32 + offset);
        let loaded_element = self.load_internal(offset);
        (loaded_element, gas_cost)
    }
```

khaeljy: This playground seems confirm that the `MSIZE` isn't `offset + 32` but `(offset % 32) * (32 + 1)`

https://www.evm.codes/playground?fork=shanghai&unit=Wei&codeType=Mnemonic&code='%2F%2F%20Put%20the%20state%20in%20memoryw32%200xppppplFFq0zMSTOREvzq1zMLOADv'~lllz%5CnwzPUSHvzzMSIZEqw1%20p~~l00%01lpqvwz~_

crates/evm/src/instructions/memory_operations.cairo
```diff
@@ -3,64 +3,92 @@
...
 
-/// MLOAD operation.
-/// Load word from memory and push to stack.
-fn exec_mload(ref context: ExecutionContext) {}
+#[generate_trait]
+impl MemoryOperationsImpl of MemoryOperationsTrait {
+    /// 0x51 - MLOAD operation.
+    /// Load word from memory and push to stack.
+    fn exec_mload(ref self: ExecutionContext) -> Result<(), EVMError> {
+        panic_with_felt252('MLOAD not implement yet')
```
ClementWalter, line 13: why not panicking for all the non implemented opcode?

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,63 @@
...
+        ),
+        // This is the test case in kakarot cairo0, but playground say, the memory size should be 736 ???
+        (0x0000000000000000000000000000000000000000000000000000000000000001, 684, 0x0, 684 + 32),
+        // evm.codes
+        (0x00000000000000000000000000000000000000000000000000000000000000FF, 0, 0xFF, 32),
+        (0x00000000000000000000000000000000000000000000000000000000000000FF, 1, 0xFF00, 64)
+    ];
+
+    let mut i = 0;
+    loop {
```
ClementWalter, line 35: don't do a loop in a test to test several different cases, write one test per case, or use fixtures if there are the same body

ClementWalter (changes requested):
Please write one test per case, and put them the same name as they add in cairo 0

crates/evm/src/instructions/memory_operations.cairo
enitrat: I think it's just because he added the test and verifies that it panis until it's implemented

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,63 @@
...
+#[available_gas(20000000000)]
+#[should_panic(expected: ('MLOAD not implement yet',))]
+fn test_exec_mload() {
+    let data: Array<(u256, u256, u256, u32)> = array![
+        // (memory_value, input, expected_value, expected_memory_size)
+
+        // Geth: no test ?!
+
+        // Kakarot cairo0
+        (0x0000000000000000000000000000000000000000000000000000000000000001, 0, 0x1, 32),
```
enitrat, line 19: > 0x0000000000000000000000000000000000000000000000000000000000000001

why so many zeroes, it's just the value 1, you can just use `0x1` or simply `1`

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
enitrat, line 25: normally expected memory size is always a multiple of 32

enitrat: > The memory load() function seems to add 32 to the offset.

Memory expansion is performed in `expand` which only increments modulo 32

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
enitrat: Let's do this if n_test is <5, otherwise, let's keep it data & loop-based (e.g. other geth tests with n>10 test cases can stay as is

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
khaeljy: Yes, i'm totally agree with you, but this is not what is tested in Kakarot cairo0.
In kakarot cairo0, the test is ` test_offset + 32` ([cf](https://github.com/kkrt-labs/kakarot/blob/b6619adb6092984f4c6b4fa512ebf07ec8d47299/tests/src/kakarot/instructions/test_memory_operations.cairo#L131C38-L131C50))

crates/evm/src/instructions/memory_operations.cairo
khaeljy: Yes, I've implemented it only on the function I want to test.

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
khaeljy: > I see actually 3 tests in `kakarot`
> 
> ```
> test__exec_mload_should_load_a_value_from_memory
> test__exec_mload_should_load_a_value_from_memory_with_memory_expansion
> test__exec_mload_should_load_a_value_from_memory_with_offset_larger_than_msize
> ```
> 
> which one are you talking about there?

This one : [test__exec_mload_should_load_a_value_from_memory_with_memory_expansion](https://github.com/kkrt-labs/kakarot/blob/b6619adb6092984f4c6b4fa512ebf07ec8d47299/tests/src/kakarot/instructions/test_memory_operations.cairo#L131C38-L131C50)

enitrat: This seems wrong then 🤔 Ideally I would like contributors to not look at the KKT0 codebase, at least not before finishing writing the cairo 1 behavior, because there might be some errors in the KKT0 codebase and we should not assume that it is correct and blindly reproduce the behavior. The Cairo1 implementation and tests should only be derived from the EVM specification

khaeljy: The issue specifies that geth, evm.codes and KKT0 tests must be retrieved. Since we have 3 sets of tests from 3 different sources, I think, on the contrary, that this helps to highlight bugs that might be present in KKT0.

enitrat: > that this helps to highlight bugs that might be present in KKT0.

Definitely, but the goal here is to spot bugs in KKT0, not reproduce them in SSJ! And if you look at the implementation in KKT0 you might have bias when working on the SSJ one

enitrat: I see what you mean, but unfortunately we shouldn't blindly trust it

khaeljy: I see, and I agree with you.
We should therefore modify the issues to remove the task: `Create the test suite for MLOAD, taking Kakarot cairo0 examples`

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -0,0 +1,44 @@
...
+}
+
+#[test]
+#[available_gas(20000000000)]
+#[should_panic(expected: ('MLOAD not implement yet',))]
+fn test_exec_mload_should_load_a_value_from_memory_with_offset_larger_than_msize() {
+    assert_mload(0x1, 684, 0x0, 736);
+}
+
+fn assert_mload(memory_value: u256, input: u256, expected_value: u256, expected_memory_size: u32) {
```
ClementWalter, line 30: can we stick to the MLOAD semantic? meaning renaming `memory_value` to `value` and `input` to `offset`

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
ClementWalter: just for the record if anyone pass by, just try on the playground

```
PUSH i
MLOAD
```

and memory size is
- i = 0 => 32
- i = 16 => 64
- i = 684 => 736

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -43,6 +43,44 @@ fn test_pc_gets_updated_properly_1() {
...
+fn test_exec_mload_should_load_a_value_from_memory_with_offset_larger_than_msize() {
+    assert_mload(0x1, 684, 0x0, 736);
+}
+
+fn assert_mload(value: u256, output: u256, expected_value: u256, expected_memory_size: u32) {
+    // Given
+    let mut ctx = setup_execution_context();
+    ctx.memory.store(value, 0);
+
+    ctx.stack.push(output);
```
enitrat, line 74: what is output? do you mean `offset`?

khaeljy: It's a bad copy / paste, I change it!

enitrat:
Moving discussion and review to #266

PR #271, opened 2023-09-04
feat: implement 0x13 - SGT Opcode

TAdev0:
## Pull Request type

- Feature

Resolves: #8 

## What is the new behavior?

Implementation of SGT EVM opcode

- No


crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
```diff
@@ -708,6 +708,431 @@ fn assert_slt(a: u256, b: u256, expected: u256) {
     assert(ctx.stack.peek().unwrap() == expected, 'slt failed');
 }
 
+#[test]
+#[available_gas(22000000)]
+fn test_exec_sgt() {
+    // https://github.com/ethereum/go-ethereum/blob/master/core/vm/testdata/testcases_sgt.json
+    assert_sgt(
+        0x0000000000000000000000000000000000000000000000000000000000000000,
```
enitrat, line 716: it would be nice to replace all `0x000...1` by `0x1 ` and all `0x000..00` by `0x0`

crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
```diff
@@ -708,6 +708,431 @@ fn assert_slt(a: u256, b: u256, expected: u256) {
...
+        0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff,
+        0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff,
+        0
+    );
+}
+
+fn assert_sgt(a: u256, b: u256, expected: u256) {
+    // Given
+    let mut ctx = setup_execution_context();
+    ctx.stack.push(a).unwrap();
```
enitrat, line 1125: extremely misleading to push `a` before `b` here, when in the implementation `a` is stack[0] (top) and `b` is stack[1]. 

It would be much better if you renamed these parameters

ClementWalter, line 1125: why `unwrap()` here?

crates/utils/src/u256_signed_math.cairo
```diff
@@ -88,7 +88,15 @@ impl U256SignedPartialOrd of SignedPartialOrd<u256> {
     }
 
     fn sgt(self: u256, other: u256) -> bool {
-        // TODO: This method will been implemented by the issue #8 (https://github.com/kkrt-labs/kakarot-ssj/issues/8)
-        panic_with_felt252("not implemented")
+        let self_positive = self.high < TWO_POW_127;
```
ClementWalter, line 91: can't we just use `SignedPartialOrd::slt(other, self)`?

ClementWalter: also, `#[inline]`, and `slt` as well

ClementWalter (changes requested):

crates/evm/src/tests/test_instructions/test_comparison_operations.cairo
TAdev0: copied SLT style already validated in main , but didn't really understand why... should have asked before.

enitrat: > actually, positive integers ares all values of the uint256 starting from TWO_POW_127, right

Not exactly, positive integers are values starting from `2**255 -1`, which is`0x7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`. But since the low part is always 2**128-1 we can just check that the high part is under `2**127`

enitrat: >  it also pushes a and b in the wrong order

It's not that they're pushed in the wrong order, it's just that they have the wrong name if we want to respect naming from the implementation. 

when you do `assert_slt(0,1,1)`, you're actually checking whether `1<0`, which is false, so the expected result is correct. I think the misleading part is that you expect parameters to be in the same order as the `slt` function so `1.slt(0)` is false, but the input to `assert_slt` is the order in which values are pushed to the stack.

TAdev0: understood ! thank you so much

crates/utils/src/u256_signed_math.cairo
TAdev0: thanks for the quick answer. 
what do you mean by  ```#[inline]```? 

Yes ican use ```fn slt```for SGT implementation !

enitrat: when you annotate a function with `#[inline(always)]`, the code of the function will be inlined in the functions that call this function, meaning that the code will be executed directly without a `call` to invoke a specific function. In some cases, it's a better option since it can increase performances, but we have to be careful because it can also drastically increase code size.

enitrat: https://en.wikipedia.org/wiki/Inline_expansion

crates/utils/src/u256_signed_math.cairo
```diff
@@ -86,9 +85,4 @@ impl U256SignedPartialOrd of SignedPartialOrd<u256> {
             self < other
         }
     }
-
-    fn sgt(self: u256, other: u256) -> bool {
```
enitrat, line 90: no, instead of deleting this function, what you should do is

```rust
fn sgt(self: u256, other: u256) -> bool {
    SignedPartialOrd::slt(other,self)
}
```

crates/utils/src/u256_signed_math.cairo
```diff
@@ -69,7 +69,6 @@ fn u256_signed_div_rem(a: u256, div: NonZero<u256>) -> (u256, u256) {
 // Signed integer comparaison
 trait SignedPartialOrd<T> {
     fn slt(self: T, other: T) -> bool;
```
enitrat, line 71: and above this line you can add the `#[inline(always)]` attribute. Also above `fn sgt`

crates/utils/src/u256_signed_math.cairo
TAdev0: thank you. i'll add it to SLT and SGT. 

do you want me to rename SLT like this?

```
fn assert_slt(b: u256, a: u256, expected: u256) {
    // Given
    ctx.stack.push(b);
    ctx.stack.push(a);
```

TAdev0:
@enitrat  everything should be ok

crates/utils/src/u256_signed_math.cairo
```diff
@@ -86,9 +87,8 @@ impl U256SignedPartialOrd of SignedPartialOrd<u256> {
             self < other
         }
     }
```
enitrat, line 89: add empty line here between functions

TAdev0: done

ClementWalter (approved):

Issue #270, opened 2023-09-04
bug: `store_aligned_words` is bugged for elements >=16


enitrat:
# Bug Report

**Kakarot version:**
main

**Current behavior:**
```rust
    fn store_aligned_words(ref self: Memory, mut chunk_index: usize, mut elements: Span<u8>) {
        loop {
            if elements.len() == 0 {
                break;
            }

            let current: felt252 = ((*elements[0]).into() * 256.wrapping_pow(15)
                + (*elements[1]).into() * 256.wrapping_pow(14)
                + (*elements[2]).into() * 256.wrapping_pow(13)
                + (*elements[3]).into() * 256.wrapping_pow(12)
                + (*elements[4]).into() * 256.wrapping_pow(11)
                + (*elements[5]).into() * 256.wrapping_pow(10)
                + (*elements[6]).into() * 256.wrapping_pow(9)
                + (*elements[7]).into() * 256.wrapping_pow(8)
                + (*elements[8]).into() * 256.wrapping_pow(7)
                + (*elements[9]).into() * 256.wrapping_pow(6)
                + (*elements[10]).into() * 256.wrapping_pow(5)
                + (*elements[11]).into() * 256.wrapping_pow(4)
                + (*elements[12]).into() * 256.wrapping_pow(3)
                + (*elements[13]).into() * 256.wrapping_pow(2)
                + (*elements[14]).into() * 256.wrapping_pow(1)
                + (*elements[15]).into() * 256.wrapping_pow(0));

            self.items.insert(chunk_index.into(), current.try_into().unwrap());
            chunk_index += 1;
            elements = elements.slice(0, 16);
        }
    }
```

The slice will slice from [0,16] at each iteration, effectively creating an infinite loop as elements.len() will never reach zero
**Expected behavior:**

The slice increments the elements pointer by 16 and reduces the total length by 16
```rust
elements = elements.slice(16, elements.len() - 16);
```

**Steps to reproduce:**

**Related code:**

```
insert short code snippets here
```

**Other information:**

PR #273, opened 2023-09-04
fix: store_aligned_words slice

enitrat:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #270 

## What is the new behavior?

- Array is now moved 16 elements further instead of staying the same
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -113,7 +113,7 @@ impl MemoryImpl of MemoryTrait {
             let aligned_bytes = elements
                 .slice(
                     16 - offset_in_chunk_i,
-                    elements.len() - 16 - offset_in_chunk_i - offset_in_chunk_f,
+                    elements.len() - (16 - offset_in_chunk_i) - offset_in_chunk_f,
                 );
             self.store_aligned_words(initial_chunk + 1, aligned_bytes);
         }
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -262,7 +262,7 @@ impl InternalMemoryMethods of InternalMemoryTrait {
 
             self.items.insert(chunk_index.into(), current.try_into().unwrap());
             chunk_index += 1;
-            elements = elements.slice(0, 16);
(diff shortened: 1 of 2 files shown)


crates/evm/src/memory.cairo
```diff
@@ -262,7 +262,7 @@ impl InternalMemoryMethods of InternalMemoryTrait {
 
             self.items.insert(chunk_index.into(), current.try_into().unwrap());
             chunk_index += 1;
-            elements = elements.slice(0, 16);
+            elements = elements.slice(16, elements.len() - 16);
```
ClementWalter, line 265: isn't this more expensive that just incrementing a start index in the loop?

ClementWalter (changes requested):
isn't there any test to update? if no, a test to add I guess!

crates/evm/src/memory.cairo
enitrat: it's slightly better this way

crates/evm/src/tests/test_memory.cairo
```diff
@@ -112,6 +112,58 @@ fn test_store_n_no_aligned_words() {
     assert(memory.bytes_len == 32, 'memory should be 32 bytes');
 }
 
+#[test]
+#[available_gas(200000000)]
+fn test_store_n_2_aligned_words() {
+    let mut memory = MemoryTrait::new();
+    let bytes_arr = array![
```
Eikix, line 119: should we have a util that inserts sequentially elements inside an array? 
pseudo-code:
```
fill_array_from_to(start: T, end: T) -> array<T>
```

might need to take oneable and add trait logic from herodotus

can be in a separate PR if you think its worth

enitrat: I don't see a usage in multiple places for now, let's open a PR later if it becomes required

enitrat:
@ClementWalter can you re-approve pls

Eikix (approved):

ClementWalter (approved):

Issue #115, opened 2023-08-17
feat: implement 0x20 - SHA3 Opcode


Eikix:
[...]

PR #281, opened 2023-09-05
feat: implement 0x20 - SHA3 Opcode 

Quentash:
Implementation of SHA3 opcode (0x20).
The code has been run against official ethereum tests found at https://github.com/ethereum/tests/blob/develop/src/GeneralStateTestsFiller/VMTests/vmTests/sha3Filler.yml#L52

Only a few tests couldn't be run because the actual `memory.bytes_len` is a u32 and some tested value are u64, It has been discussed with @Eikix to leave it as it is for now and eventually refactor it later to handle u64 memory size.

I have added comments in the code ( to eventually remove later ) to explain how the code bypasses continuous memory allocation when trying to access a memory slot beyond `memory.bytes_len` and will instead ensure that the memory is allocated at the end of the process, this seems way more efficient since some tests will fail otherwise.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #115

## What is the new behavior?

-
-
-

- [ ] Yes
- [X] No


crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
+impl Sha3Impl of Sha3Trait {
+    /// SHA3 operation.
+    /// Hashes n bytes in memory at a given offset in memory.
+    /// # Specification: https://www.evm.codes/#20?fork=shanghai
+    fn exec_sha3(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let offset: u32 = Into::<u256, Result<u32, EVMError>>::into((self.stack.pop()?))?;
+        let mut size: u32 = Into::<u256, Result<u32, EVMError>>::into((self.stack.pop()?))?;
+        let init_size = size;
+
+        let mut toHash: Array<u64> = ArrayTrait::<u64>::new();
```
enitrat, line 24:
```suggestion
        let mut to_hash: Array<u64> = Default::default();
```

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
+        let mut last_input: u64 = 0;
+        let mut counter = 0;
+        loop {
+            if size < 32 {
+                break;
+            }
+            // If we try to read unallocted memory slot, we'll feed the data to hash with 0s
+            // and allocated the memory space the end of the process, which is cheaper than allocating every load().
```
enitrat, line 32:
```suggestion
            // and allocate the memory space at the end of the process, which is cheaper than allocating every load().
```

is this what you meant?

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
+        let init_size = size;
+
+        let mut toHash: Array<u64> = ArrayTrait::<u64>::new();
+        let mut last_input: u64 = 0;
+        let mut counter = 0;
+        loop {
+            if size < 32 {
+                break;
+            }
+            // If we try to read unallocted memory slot, we'll feed the data to hash with 0s
```
enitrat, line 31:
```suggestion
            // If we try to read unallocted memory slot, we'll pad left the data to hash with 0s
```

is this what you meant?

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
+                size -= 32;
+                continue;
+            }
+            // Load the 32 words and reverse the bytes order,
+            let mut loaded = self.memory.load_internal((offset + (32 * counter)));
+            loaded.low = integer::u128_byte_reverse(loaded.low);
+            loaded.high = integer::u128_byte_reverse(loaded.high);
+
+            // Split the loaded word into u64 to feed cairo_keccak
+            let (highL, lowL) = u128_split(loaded.high);
```
enitrat, line 48: snake_case everywhere

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
+            toHash.append(lowL);
+            toHash.append(highL);
+            toHash.append(lowH);
+            toHash.append(highH);
+
+            counter += 1;
+            size -= 32;
+        };
+
+        if size > 0 {
```
enitrat, line 59: can you add a small comment here :)

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
+        let mut toHash: Array<u64> = ArrayTrait::<u64>::new();
+        let mut last_input: u64 = 0;
+        let mut counter = 0;
+        loop {
+            if size < 32 {
+                break;
+            }
+            // If we try to read unallocted memory slot, we'll feed the data to hash with 0s
+            // and allocated the memory space the end of the process, which is cheaper than allocating every load().
+            if (offset + (32 * counter)) > self.memory.bytes_len {
```
enitrat, line 33: can you make `offset + (32 * counter)` a variable with an explicit name that indicates what it represents?

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
 
-/// SHA3 operation.
-/// Hashes n bytes in memory at a given offset in memory.
-/// # Specification: https://www.evm.codes/#20?fork=shanghai
-fn exec_sha3(ref context: ExecutionContext) {}
+use array::ArrayTrait;
+
+#[generate_trait]
+impl Sha3Impl of Sha3Trait {
+    /// SHA3 operation.
```
enitrat, line 16: A rustdoc-like description of what the function does would be helpful for other developers reading this. 
You need to describe the algorithm applied in natural language (e.g. context on why we apply, why we split u128s into u64s, etc

crates/evm/src/instructions/sha3.cairo
enitrat: A comment that reflects the actions performed in the code is not useful - the comment must inform on the why, not the how. Here if I understand well, it would be something along the lines of

```rust
// Pad left the hash input with zeroes for out-of-bound bytes
```

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,103 @@
...
+            toHash.append(lowH);
+            toHash.append(highH);
+
+            counter += 1;
+            size -= 32;
+        };
+
+        if size > 0 {
+            let mut loaded = 0;
+            if (offset + (32 * counter)) > self.memory.bytes_len {
```
enitrat, line 61:
```rust 
            let mut loaded = if (offset + (32 * counter)) > self.memory.bytes_len { ...
```

crates/utils/src/helpers.cairo
```diff
@@ -102,6 +102,20 @@ fn split_word_128(value: u256, ref dst: Array<u8>) {
     split_word(value, 16, ref dst)
 }
 
+/// Try to convert u128 to u64,
+fn u128_to_u64(input: u128) -> u64 {
```
enitrat, line 106: this is useless, why not simply `input.try_into().unwrap()` where it's needed?

crates/utils/src/helpers.cairo
```diff
@@ -102,6 +102,20 @@ fn split_word_128(value: u256, ref dst: Array<u8>) {
...
+    input.try_into().unwrap()
+}
+
+/// Splits a u128 into two u64.
+fn u128_split(input: u128) -> (u64, u64) {
```
enitrat, line 111: I see that it's something from `corelib/src.keccak`. You can import it into your sha3 file with

`use keccak::u128_split;`

crates/evm/src/instructions/sha3.cairo
Quentash: It is, indeed, mb

crates/evm/src/instructions/sha3.cairo
Quentash: That is very instructive, I never realized this was doable, thanks!

Quentash:
Used Default::default() as array initiator.
Corrected various comment and added some.
Added SHA3 rustdoc-like comment. ( Please tell me if this one is explicit enough ).
Added `slot_to_read` variable which made me realize I could dispose the `counter` one.
Replaced sneaky camels by snakes.

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,10 +1,110 @@
...
+    /// Format 32 bytes chunk of data read from memory into 64 bits chunk in little endian
+    /// to be able to call cairro_keccak.
+    /// 
+    /// # Specification: https://www.evm.codes/#20?fork=shanghai
+    fn exec_sha3(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let offset: u32 = Into::<u256, Result<u32, EVMError>>::into((self.stack.pop()?))?;
+        let mut size: u32 = Into::<u256, Result<u32, EVMError>>::into((self.stack.pop()?))?;
+        let init_size = size;
+
+        let mut to_hash: Array<u64> = Default::default();
```
enitrat, line 31: this part should be split into multiple sub functions. Define internal functions inside an `internal` mod for everything related to
- Padding of input
- Transformation to u64 words
- Bytes reversing

This entire function should not be more than 10 - 20 lines long. Separate it by concern as much as you can to make the code easier to read, understand and debug. Use helper functions where required. These functions will need to be unit tested as well

Quentash: I will do the same thing with the LOG opcode because it has been written the same way than this.

Quentash:
I have pushed a new version of the sha3 algorithm hoping it is clearer and better structured. ( There is a significant improvement in gas cost concerning big sized test 6993735300 -> 1703355056 for the bigger one. )
I still have tests to add but I pushed this commit to allow you to check on `Memory.load()`.

Indeed, it appears that the memory expansion gas cost calculation makes one of my test fail with what firstly appears to be an infinite loop that ends up with a `error: process did not exit successfully: signal: 9 (SIGKILL)` message.
If I comment those lines, my tests appear to run normally. "Funny" thing to note, this is not even the "biggest" test that fails, see screenshots.

With comments :
image

Without comments :
image

enitrat (commented):
There is still some refactoring to be done on how the function is constructed internally, otherwise it's a good work!

Here's an example of a keccak256 function taking an array of bytes as an input. As you can see it's around 12 lines long, and the intent by each line can be directly understood, even without having functions implementations available. This is ideally what you should aim for when writing code

```rust
fn cairo_keccak_le_bytes(mut self: Array<u8>) -> u256 {
    let mut words64: Array<u64> = array![];
    let mut self = self.span();
    let (last_word, last_word_bytes) = loop {
        // Specifically handle last word
        if self.len() < 8 {
            let (value, n_bytes) = U64Trait::from_le_bytes(self);
            break (value, n_bytes);
        };
        let mut current_word = self.slice(0, 8);
        let (value, n_bytes) = U64Trait::from_le_bytes(current_word);
        words64.append(value);
        self = self.slice(8, self.len() - 8);
    };
    let mut hash = cairo_keccak(ref words64, last_word, last_word_bytes);
    reverse_endianness(hash)
}


fn reverse_endianness(value: u256) -> u256 {
    let new_low = integer::u128_byte_reverse(value.high);
    let new_high = integer::u128_byte_reverse(value.low);
    u256 { low: new_low, high: new_high }
}
```

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,110 @@
...
+    }
+}
+
+
+mod internal {
+    use utils::helpers::split_u256_into_u64_little;
+    /// Return the last u64 chunk to hash, given a size, from an u256.
+    /// This function is used to prepare inputs for keccak::cairo_keccak.
+    ///
+    /// This function will split a given u256 into little endian u64 and
```
enitrat, line 82: Even the (GPT-generated?) description is not accurate 😛

Quentash: Hahaha !! No, I sadly don't need AI to sound confusing ! My bad !
I do understand why it is inaccurate seeing the other comments tho, will change that.

enitrat: haha I said that because I noticed a phrasing pattern I often see when I ask it to generate doc 😆

Quentash:
> > I have pushed a new version of the sha3 algorithm hoping it is clearer and better structured. ( There is a significant improvement in gas cost concerning big sized test 6993735300 -> 1703355056 for the bigger one. ) I still have tests to add but I pushed this commit to allow you to check on `Memory.load()`.
> > Indeed, it appears that the memory expansion gas cost calculation makes one of my test fail with what firstly appears to be an infinite loop that ends up with a `error: process did not exit successfully: signal: 9 (SIGKILL)` message. If I comment those lines, my tests appear to run normally. "Funny" thing to note, this is not even the "biggest" test that fails, see screenshots.
> > With comments : image
> > Without comments : image
> 
> I tried locally and I didn't have this problem!

Ok that is even more weird. I tried it again and it gives me the same results.

Here is my scarb --version output : 
scarb 0.7.0 (58cc88efb 2023-08-23)
cairo: 2.2.0 (https://crates.io/crates/cairo-lang-compiler/2.2.0)
sierra: 1.3.0
In case yours differ ?

Otherwise, could it be because I'm on WSL ?
I finish pushing the required corrections and I will try to run the tests on a native Linux.

Would you have any other ideas ?

enitrat:
> Otherwise, could it be because I'm on WSL ?

There's a chance it's because of this. I'm running on macOS

Quentash:
Added suggested functions.
Revamped the already existing functions and comments.
Corrected typos.
Noticed a small gas improvement on big sized test, ~1%ish, but still!

crates/evm/src/instructions/sha3.cairo
```diff
@@ -43,53 +99,45 @@ impl Sha3Impl of Sha3Trait {
...
-        // Bytes from unallocated memory are set to 0
+    /// This function will fill an u64 array with a given amount of 
+    /// 32 Bytes chunks of zeroes.
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - The context in which the memory is read
+    /// * `to_hash` - A reference to the array to fill
+    /// * `amount` - The amount of 32 Bytes chunks of zeroes to append to to hash
+    fn fill_array_with_zeroes(
```
enitrat, line 113: you can use `append_n` from `utils::helpers`

crates/evm/src/instructions/sha3.cairo
```diff
@@ -7,32 +7,88 @@ use evm::memory::MemoryTrait;
...
 
         let mut to_hash: Array<u64> = Default::default();
-        let mut last_input: u64 = 0;
 
-        // Fill to_hash with bytes from memory
+        let (chunks_from_mem, chunks_of_zeroes) = internal::compute_data_origin(
+            size, offset, self.memory.bytes_len
+        );
+        internal::fill_array_with_memory_chunks(ref self, ref to_hash, ref offset, chunks_from_mem);
+        internal::fill_array_with_zeroes(ref self, ref to_hash, chunks_of_zeroes);
```
enitrat, line 34:
```suggestion
    to_hash.append_n(0, chunks_of_zeroes);
```

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,157 @@
...
+    /// * `size` - The amount of bytes to read
+    /// 
+    /// # Specification: https://www.evm.codes/#20?fork=shanghai
+    fn exec_sha3(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let mut offset: u32 = Into::<u256, Result<u32, EVMError>>::into((self.stack.pop()?))?;
+        let mut size: u32 = Into::<u256, Result<u32, EVMError>>::into((self.stack.pop()?))?;
+
+        let mut to_hash: Array<u64> = Default::default();
+
+        let (chunks_from_mem, chunks_of_zeroes) = internal::compute_data_origin(
```
enitrat, line 30:
```suggestion
        let (nb_chunks, nb_zeroes) = internal::compute_data_origin(
```

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,157 @@
...
+    /// from the memory and how many chunks of 32 Bytes data must be filled
+    /// with zeroes.
+    ///
+    /// # Arguments
+    ///
+    /// * `size` - The amount of bytes to hash
+    /// * `offset` - Offset in memory
+    /// * `mem_len` - Size of the memory
+    /// Returns : (chunk_from_mem, chunks_of_zeroes)
+    fn compute_data_origin(size: u32, offset: u32, mem_len: u32) -> (u32, u32) {
```
enitrat, line 66: @ClementWalter I don't really like this naming but I'm out of inspiration, what would you name it?

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,157 @@
...
+    /// This function will fill an array with little endian u64
+    /// by splitting 32 Bytes chunk read from the memory.
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - The context in which the memory is read
+    /// * `to_hash` - A reference to the array to fill
+    /// * `offset` - Offset in memory
+    /// * `amount` - The amount of 32 Bytes chunks to read from memory
+    fn fill_array_with_memory_chunks(
```
enitrat, line 86: instead of passing `offset` by ref, return it explicitly at the end of the function

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,157 @@
...
+    /// depending on size from a 32 Bytes chunk of data and return
+    /// the u64 chunk containing the last bytes that aren't 8 Bytes long.
+    ///
+    /// # Arguments
+    ///
+    /// * `to_hash` - A reference to the array to fill
+    /// * `value` - The 32 Bytes chunk to split and get the bytes from
+    /// * `size` - The amount of bytes still required to hash
+    /// Returns the last u64 chunk that isn't 8 Bytes long.
+    fn fill_array_with_last_inputs(ref to_hash: Array<u64>, value: u256, size: u32) -> u64 {
```
enitrat, line 139:
```suggestion
    fn prepare_last_input(ref to_hash: Array<u64>, value: u256, size: u32) -> u64 {
```

if you call it `fill_array_with_last_inputs` then an unaware reader would assume that to_hash is filled with last_inputs, where in reality, not only, we prepare the last_input arguments for the call to `cairo_keccak`

enitrat (commented):
Almost there!

crates/evm/src/instructions/sha3.cairo
Quentash: Feeling so dumb, sorry

crates/evm/src/instructions/sha3.cairo
Quentash: Shall I suggest "compute_memory_chunks_amount" ?

Quentash:
I pushed a correction which include the modification of `compute_data_origin` into `compute_memory_chunks_amount` but I am totally open to any suggestions!

crates/evm/src/tests/test_instructions/test_sha3.cairo
```diff
@@ -0,0 +1,376 @@
...
+use evm::memory::{InternalMemoryTrait, MemoryTrait};
+use evm::stack::StackTrait;
+use option::OptionTrait;
+use evm::errors::{EVMError, TYPE_CONVERSION_ERROR};
+
+use debug::PrintTrait;
+
+#[test]
+#[available_gas(20000000)]
+fn test_sha3_size_0_offset_0() {
```
enitrat, line 14: all tests testing the `exec_sha3` function should be called `test_exec_sha3...`

This makes it easier if we want to execute only specific tests using filtering.

crates/evm/src/tests/test_instructions/test_sha3.cairo
```diff
@@ -0,0 +1,376 @@
...
+    // Then
+    let result = ctx.stack.peek().unwrap();
+    assert(
+        result == 0x2022ae07f3a362b08ac0a4bcb785c830cb5c368dc0ce6972249c6abbc68a5291, 'wrong result'
+    );
+}
+
+#[test]
+#[available_gas(20000000000)]
+fn test_sha3_internal_fill_array_with_memory_chunks() {
```
enitrat, line 277: unit tests of internal functions should not be called `test_sha3` as they're not testing sha3

crates/evm/src/tests/test_instructions/test_sha3.cairo
```diff
@@ -0,0 +1,376 @@
...
+#[test]
+#[available_gas(20000000000)]
+fn test_sha3_size_0x0C80_offset_0() {
+    // Given
+    let mut ctx = setup_execution_context();
+
+    ctx.stack.push(0x0C80);
+    ctx.stack.push(0x00);
+
+    let mut memDst: u32 = 0;
```
enitrat, line 254: snake_case everywhere

enitrat:
@Eikix good for a final review

crates/utils/src/helpers.cairo
```diff
@@ -103,6 +104,21 @@ fn split_word_128(value: u256, ref dst: Array<u8>) {
     split_word(value, 16, ref dst)
 }
 
+/// Splits an u256 into 4 little endian u64.
+/// Returns ((high_high, high_low),(low_high, low_low))
+fn split_u256_into_u64_little(value: u256) -> ((u64, u64), (u64, u64)) {
```
Eikix, line 109: imo these two helpers should be methods implemented on the u256 type, e.g. with an auto generated trait

Eikix: i also propose to have ubiquitous naming for little endian and big endian:
a `le` suffix and `be` suffix on function names, to match rust and python implementation of endianness utils

enitrat: it also matches Cairo's

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,138 @@
...
+    /// from the memory and how many chunks of 32 bytes data must be filled
+    /// with zeroes given a target size, an memory offset and the length of the memory.
+    ///
+    /// # Arguments
+    ///
+    /// * `size` - The amount of bytes to hash
+    /// * `offset` - Offset in memory
+    /// * `mem_len` - Size of the memory
+    /// Returns : (nb_chunks, nb_zeroes)
+    fn compute_memory_chunks_amount(size: u32, offset: u32, mem_len: u32) -> (u32, u32) {
```
Eikix, line 68: this is a bit misleading because afaik, chunks in memory usual language designates a 16-bit interval

Eikix: i propose `word` to designate a 32-byte piece of data
memory `chunk` to designate a 16-byte piece of data

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,138 @@
...
+        if offset > mem_len {
+            return (0, size / 32);
+        }
+        if (mem_len - offset < 32) && (size > 32) {
```
Eikix, line 72: can add comments to explain what is happening?

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,138 @@
...
+            return (1, (size / 32) - 1);
+        }
+        let nb_chunks = cmp::min(mem_len - offset, size) / 32;
```
Eikix, line 75: can add comments?

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,138 @@
...
+    /// returns the next offset to read from.
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - The context in which the memory is read
+    /// * `to_hash` - A reference to the array to fill
+    /// * `offset` - Offset in memory to start reading from
+    /// * `amount` - The amount of 32 bytes chunks to read from memory
+    /// Return the new offset
+    fn fill_array_with_memory_chunks(
```
Eikix, line 90: same remark as above

crates/evm/src/instructions/sha3.cairo
Eikix: what happens here if mem_len == 32, offset == 0 and size == 31?

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,138 @@
...
+    /// Returns : (nb_chunks, nb_zeroes)
+    fn compute_memory_chunks_amount(size: u32, offset: u32, mem_len: u32) -> (u32, u32) {
+        if offset > mem_len {
+            return (0, size / 32);
+        }
+        if (mem_len - offset < 32) && (size > 32) {
+            return (1, (size / 32) - 1);
+        }
+        let nb_chunks = cmp::min(mem_len - offset, size) / 32;
+        (nb_chunks, (size / 32) - nb_chunks)
```
Eikix, line 76: what happens here if mem_len == 32, offset == 0 and size == 31? 

wouldn't (size / 32) - nb_chunks underflow?

crates/utils/src/helpers.cairo
Quentash: Agree with the proposed solutions.
I used this naming because other helpers functions already used this naming, maybe those should be changed too (ideally in an a separated PR) for the sake of consistency ? I can take care of that if you wish.

enitrat: OK to do it in another PR

crates/evm/src/instructions/sha3.cairo
enitrat: @Quentash did you forget to push? you marked as resolve but no changes

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,146 @@
...
+
+        let mut to_hash: Array<u64> = Default::default();
+
+        let (nb_words, nb_zeroes) = internal::compute_memory_words_amount(
+            size, offset, self.memory.bytes_len
+        );
+        let mut last_input_offset = internal::fill_array_with_memory_words(
+            ref self, ref to_hash, offset, nb_words
+        );
+        to_hash.append_n(0, 4 * nb_zeroes);
```
Eikix, line 36: can you add a comment why we need to do this? 
with links to the EVM resource that explains this zeroes padding problem

crates/evm/src/instructions/sha3.cairo
```diff
@@ -1,9 +1,146 @@
...
+
+        // Fill last_input with last bytes to hash
+        let last_input: u64 = if (size % 32 != 0) {
```
Eikix, line 39: can you add a comment to explain the logic behind last_input manipulation?

Eikix (dismissed):
as far as i'm concerned this PR is ready
did we make sure:
- we have all the tests that geth uses?
- all the tests that EELS uses?

truly nice display of patience and good cooperation, thank you for your hard work @Quentash @enitrat

crates/evm/src/instructions/sha3.cairo
Quentash: I don't think this really is a problem, the idea is to append 0s this way because it is faster/cheaper than reading them from memory. 
I've added a comment which explains this tho !

Have I misunderstood/miss something ?

Quentash:
> as far as i'm concerned this PR is ready did we make sure:
> 
> * we have all the tests that geth uses?
> * all the tests that EELS uses?
> 
> truly nice display of patience and good cooperation, thank you for your hard work @Quentash @enitrat

I have taken tests found here : https://github.com/ethereum/go-ethereum/blob/master/tests/solidity/test/opCodes.js
which are quite similiar to those found here : https://github.com/ethereum/legacytests/tree/97de5bce3a257a4b219666134df15b6d4d47f1f5/Constantinople/VMTests/vmSha3Test

Beside tests with size or offset > u64.maxvalue, I've taken all I could, some are even e bit repetitve and I'm not sure they are really relevant in our case but I still added them.
I'm open to suggestions !

crates/evm/src/instructions/sha3.cairo
Eikix: no problem! I was simply asking for comments explaining the choice for readability:)! no problem

crates/evm/src/instructions/sha3.cairo
```diff
@@ -33,9 +33,13 @@ impl Sha3Impl of Sha3Trait {
         let mut last_input_offset = internal::fill_array_with_memory_words(
             ref self, ref to_hash, offset, nb_words
         );
+        // Fill array to hash with zeroes for bytes out of memory bound
```
Eikix, line 36: do we still perform memory expansion? for bytes out of memory bound? 
do we need to?

Quentash: I've just realised that we don't do the "last memory expansion" for multiple of 32 size.
But otherwise, yes, my idea was to do the memory expansion only once at the end.
Should I modify it tho ? I'm afraid I don't really understand what would be the actual impact.
Because since we only read from memory, all the "expanded" bytes would be zeroes, is it therefore relevant to update de memory length since all it contains are 0 ?

Eikix: good question! I'm not sure here, as the tradeoff here is gas/performance vs. sticking to official spec
i imagine we need to perform all the expansion that the spec requires
@enitrat

enitrat: Ah, yes, it used to be here at the beginning but is no longer there after several refactors. 

The thing is, ideally, memory expansions is only handled from the `memory` module, not from outside, as it's an internal mechanism to the memory. However here as we're looking for optimisations and we avoid reading zero words from memory, we might better just call `memory.ensure_length(offset+size)` directly in sha3

Quentash: Roger that !

Quentash:
Ensure memory length to be set even if we skipped reading for zeroes.

enitrat:
@Quentash can you add a test that checks the memory length when we read out of bound bytes? To make sure that the length is correctly updated.

Quentash:
> @Quentash can you add a test that checks the memory length when we read out of bound bytes? To make sure that the length is correctly updated.

Of course, my bad !!

Quentash:
Added memory length verification to tests.
Made me realise it was better to ensure_length outside of the "if", thank you !

Eikix (approved):
the long awaited moment... 🔥🔥🔥🔥🔥

enitrat:
🥳

Issue #283, opened 2023-09-05
bug: Memory expansion should only increments modulo 32


khaeljy:
# Bug Report

**Kakarot version:**
main

**Current behavior:**
Memory expansion is not by increments modulo 32.

**Expected behavior:**
Memory expansion should only increments modulo 32. ([playground](https://www.evm.codes/playground?fork=shanghai&unit=Wei&codeType=Mnemonic&code='jInitiallyhkufirst~1mkx21upart%20ofgecond~2ms'~mqPOPqjNowgize%20is%20v%20%2F%2F%20uqMLOADvRead%20q%5Cnm%20wordkqPUSH1hjMSIZEvh%200g%20s%01ghjkmquv~_))

**Steps to reproduce:**
By initializing the memory, then expanding the memory by 1, the new size should be 64 bytes, not 33 bytes.

**Related code:**

```
// Given
let mut memory = MemoryTrait::new();
let value: u256 = 1;
let bytes_array = helpers::u256_to_bytes_array(value);

memory.store_n(bytes_array.span(), 0);

// When
let cost = memory.expand(1);

// Then
assert(cost >= 0, 'cost should be positive');
assert(memory.bytes_len == 64, 'memory should be 64bytes');
let value = memory.load_internal(0);
assert(value == 1, 'value should be 1');
```

khaeljy:
I can take this

Eikix:
Hi, thanks for the bug report!
I think there might be a misunderstanding. 

Because MLOAD loads a word that is 32 bytes long, it'll expand at the same time of loading the word, thus expanding 32 bytes per 32 bytes.

You can try with other opcodes that can trigger memory expansion, such as SHA3, and try in a way that is not modulo 32 for reference.

khaeljy:
MLOAD does indeed load 32 bytes, but not necessarily whole words. We can load 32 bytes from offset 1, which must extend memory by 32 bytes.

I've done the exercise with SHA3, loading 33 bytes into memory should expand it by 32 bytes.

See [playground](https://www.evm.codes/playground?fork=shanghai&unit=Wei&codeType=Mnemonic&code='kPutjrequired%20valubin%20memoryf32%200xFFFFFFFF____w0~MSTORE~g~~kCalljopcodew4s~w33s'~%5Cnz0000000wf1%20sw0~SHA3gk%2F%2F%20j%20thbg~MSIZEf~PUSHbe%20_zz%01_bfgjkswz~_).

[...]

PR #284, opened 2023-09-05
bug: Memory expansion should only increments modulo 32

khaeljy:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #283 

## What is the new behavior?

- Memory expansion is by increments of modulo 32.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -413,11 +413,15 @@ impl InternalMemoryMethods of InternalMemoryTrait {
     ///
     /// The cost of the expansion.
     fn expand(ref self: Memory, length: usize) -> usize {
+        if (length == 0) {
+            return 0;
+        }
+
         let last_memory_size_word = (self.bytes_len + 31) / 32;
         let mut last_memory_cost = (last_memory_size_word * last_memory_size_word) / 512;
         last_memory_cost += (3 * last_memory_size_word);
 
-        let new_bytes_len = self.bytes_len + length;
+        let new_bytes_len = self.bytes_len + (((length + 31) / 32) * 32);
         let new_memory_size_word = (new_bytes_len + 31) / 32;
         let new_memory_cost = (new_memory_size_word * new_memory_size_word) / 512;
         let new_memory_cost = new_memory_cost + (3 * new_memory_size_word);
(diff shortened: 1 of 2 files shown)


crates/evm/src/tests/test_memory.cairo
```diff
@@ -223,7 +223,47 @@ fn test__expand__should_return_expanded_memory_and_cost() {
...
+
+#[test]
+#[available_gas(2000000000)]
+fn test__expand__should_return_expanded_memory_by_exactly_one_word_and_cost() {
+    // Given
+    let mut memory = MemoryTrait::new();
+    let value: u256 = 1;
+    let bytes_array = helpers::u256_to_bytes_array(value);
+
+    memory.store_n(bytes_array.span(), 0);
```
enitrat, line 239: if we're doing a unit test on the `expand`, then we should start from an empty memory without any previous context. Same for the test below

crates/evm/src/memory.cairo
```diff
@@ -413,11 +413,15 @@ impl InternalMemoryMethods of InternalMemoryTrait {
...
+        if (length == 0) {
+            return 0;
+        }
+
         let last_memory_size_word = (self.bytes_len + 31) / 32;
         let mut last_memory_cost = (last_memory_size_word * last_memory_size_word) / 512;
         last_memory_cost += (3 * last_memory_size_word);
 
-        let new_bytes_len = self.bytes_len + length;
+        let new_bytes_len = self.bytes_len + (((length + 31) / 32) * 32);
```
enitrat, line 424: can you add an intermediate variable for the added step, e.g. `adjusted_length`

crates/evm/src/tests/test_memory.cairo
khaeljy: I based myself on existing tests (cf [test__expand__should_return_expanded_memory_and_cost](https://github.com/kkrt-labs/kakarot-ssj/blob/2d89cbfe9a8d287ce4f38bb80a82e8853eaf35f7/crates/evm/src/tests/test_memory.cairo#L213C4-L213C56)), but okay, I can start with an empty memory.

enitrat (approved):

Issue #262, opened 2023-09-02
feat: implement 0x50 - POP Opcode


khaeljy:
# Feature Request

| Since | Group |
| --- | --- |
| Frontier| Stack Memory Storage and Flow Operations |

*Index 1 is top of the stack. See [PUSH](https://www.evm.codes/#60).*

## Stack input

1. `y`: a stack item.

## Example

|  | Input | Output |
|--|------|-------|
| `1` | `125985` | |

[Reproduce in playground](https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code='PUSH3%20125985%5CnPOP'_).

## Error cases

The state changes done by the current context are [reverted](https://www.evm.codes/#FD) in those cases:
- Not enough gas.
- Not enough values on the stack.
```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/kakarot-ssj/issues/263
- [ ] After #263 is implemented, go and implement the business logic of POP
```

Quentash:
Can I have this one please ?

PR #285, opened 2023-09-05
feat: implement 0x50 - POP Opcode

Quentash:
POP opcode implementation,

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #262

## What is the new behavior?

-
-
-

- [ ] Yes
- [X] No

### DIFF
--- a/crates/evm/src/instructions/memory_operations.cairo
+++ b/crates/evm/src/instructions/memory_operations.cairo
@@ -64,7 +64,7 @@ impl MemoryOperation of MemoryOperationTrait {
     /// Pops the first item on the stack (top of the stack).
     /// # Specification: https://www.evm.codes/#50?fork=shanghai
     fn exec_pop(ref self: ExecutionContext) -> Result<(), EVMError> {
-        let result = self.stack.pop()?;
+        self.stack.pop()?;
         // self.stack.pop() returns a Result<u256, EVMError> so we cannot simply return its result
         Result::Ok(())
     }
--- a/crates/evm/src/instructions/memory_operations.cairo
+++ b/crates/evm/src/instructions/memory_operations.cairo
@@ -64,6 +64,7 @@ impl MemoryOperation of MemoryOperationTrait {
     /// Pops the first item on the stack (top of the stack).
     /// # Specification: https://www.evm.codes/#50?fork=shanghai
     fn exec_pop(ref self: ExecutionContext) -> Result<(), EVMError> {
+        panic_with_felt252('POP not implement yet');
         Result::Ok(())
     }
 
--- a/crates/evm/src/instructions/memory_operations.cairo
+++ b/crates/evm/src/instructions/memory_operations.cairo
@@ -64,7 +64,10 @@ impl MemoryOperation of MemoryOperationTrait {
     /// Pops the first item on the stack (top of the stack).
     /// # Specification: https://www.evm.codes/#50?fork=shanghai
     fn exec_pop(ref self: ExecutionContext) -> Result<(), EVMError> {
-        self.stack.pop();
+        let result = self.stack.pop();
+        if result.is_err() {
+            return Result::Err(result.unwrap_err());
+        }
         Result::Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


crates/evm/src/instructions/memory_operations.cairo
```diff
@@ -64,6 +64,10 @@ impl MemoryOperation of MemoryOperationTrait {
     /// Pops the first item on the stack (top of the stack).
     /// # Specification: https://www.evm.codes/#50?fork=shanghai
     fn exec_pop(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let result = self.stack.pop();
```
Eikix, line 67: you can just return `self.stack.pop` here, it'll bubble up the error if it is one

```
fn exec_pop(ref self: ExecutionContext) -> Result<(), EVMError> {
    self.stack.pop()
}

Comment (Quentash) [reply crates/evm/src/instructions/memory_operations.cairo:67]: I had a problem doing that because `self.stack.pop()` returns a `Result<u256, EVMError>` instead of a `Result<(), EVMError>`.
And changing the return type of exec_pop brought problems in interpreter.cairo on these following lines ;
```
if opcode == 80 {
    // POP
    return context.exec_pop();
}
```
because it is exepecting to forward a `Result<(), EVMError>` too.
So there's why I do this "trick".

Comment (Eikix) [reply crates/evm/src/instructions/memory_operations.cairo:67]: I see! makes sense

Comment (Eikix) [reply crates/evm/src/instructions/memory_operations.cairo:67]: Can you try:
```
let _ = self.stack.pop()?;
Result::Ok(())
```

Comment (Quentash) [reply crates/evm/src/instructions/memory_operations.cairo:67]: Sure !
It apparently works like that. I thought the "?" simply was an unwrapper, thank you !
Pushing the updates.

File: crates/evm/src/instructions/memory_operations.cairo [hunk 2]
```diff
@@ -64,6 +64,8 @@ impl MemoryOperation of MemoryOperationTrait {
     /// Pops the first item on the stack (top of the stack).
     /// # Specification: https://www.evm.codes/#50?fork=shanghai
     fn exec_pop(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let result = self.stack.pop()?;
+        // self.stack.pop() returns a Result<u256, EVMError> so we cannot simply return its result
```
enitrat, line 68: Would be more elegant if we had a method like this in Cairo https://doc.rust-lang.org/std/result/enum.Result.html#method.map

but we don't 😢 

Given that `result` is not used you can remove that unused variable and just call `self.stack.pop()?;`

Quentash: I agree that would have been good to have such functions !
My bad, removed the unused variable I forgot to remove.

enitrat (approved):
lgtm!

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -42,3 +42,38 @@ fn test_pc_gets_updated_properly_1() {
     assert(ctx.stack.len() == 1, 'stack should have one element');
     assert(ctx.stack.pop().unwrap() == 9000, 'updating PC failed');
 }
+
+
+#[test]
+#[available_gas(20000000)]
+fn test_pop_should_pop_an_item_from_execution_context() {
```
enitrat, line 49:
```suggestion
fn test_exec_pop_should_pop_an_item_from_stack() {
```

enitrat (dismissed):

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -42,3 +42,38 @@ fn test_pc_gets_updated_properly_1() {
...
+
+    // Then
+    assert(result.is_ok(), 'should have succeed');
+    assert(ctx.stack.len() == 1, 'stack should have one element');
+    assert(ctx.stack.peek().unwrap() == 0x01, 'stack peek should return 0x01');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_pop_should_stack_underflow() {
```
enitrat, line 67:
```suggestion
fn test_exec_pop_should_stack_underflow() {
```

enitrat (approved):

PR #321, opened 2023-09-07
feat: implement 0x53 - MSTORE8 Opcode

khaeljy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #244 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


crates/evm/src/instructions/memory_operations.cairo
```diff
@@ -73,6 +76,12 @@ impl MemoryOperation of MemoryOperationTrait {
     /// Save single byte to memory
     /// # Specification: https://www.evm.codes/#53?fork=shanghai
     fn exec_mstore8(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let popped = self.stack.pop_n(2)?;
+        let offset: u32 = Into::<u256, Result<u32, EVMError>>::into((*popped[0]))?;
+        let value: u256 = *popped[1];
+        let values: Span<u8> = u256_to_bytes_array(value).span();
```
danilowhk, line 82: I believe here we can just use div rem or bitwise to directly extract the last value, instead of transforming the whole u256 into bytes_array.

For example, in the case of 0x1122334455, this would cost 5 bitwise / div_rem operations(which are expensive in Cairo).

crates/evm/src/instructions/memory_operations.cairo
```diff
@@ -73,6 +76,12 @@ impl MemoryOperation of MemoryOperationTrait {
...
+
+        self.memory.store_n(values.slice(values.len() - 1, 1), offset);
```
enitrat, line 84: This just made me realised that `store_n` is not the best way to interact with the memory for the MSTORE8 opcode.  Memory interactions only happen through `MSTORE` and `MSTORE8` and `XCOPY` opcodes.

Either you store a 32-bytes values (MSTORE, calling `memory.store()`) or you store a simple 1-byte value (MSTORE8, in which case calling `store_n` doesn't make sense because you 're storing one single byte), or an arbitrary number of bytes (XCOPY), in which case `store_n` makes sense.

I'm ok to approve this PR, if you do it like this

```rust
let value:u8 = (*popped[1].low & 0xFF).try_into().unwrap();
let values = array![value].span();
self.memory.store_n(values, offset);
```

(make sure that it works correctly)

I will refactor memory later 
https://github.com/kkrt-labs/kakarot-ssj/issues/325

enitrat (changes requested):

crates/evm/src/instructions/memory_operations.cairo
khaeljy: Good point! I didn't like having to convert my value into bytes array!

I have create a new helper `U256IntoResultU8` for try to convert u256 to u8 and I use it like this: 
```rust
let value: u8 = Into::<u256, Result<u8, EVMError>>::into((*popped[1] & 0xFF))?;
```

What mean `.low`? I have remove it because it causes an error: `error: Type annotations needed. Failed to infer ?10`

enitrat: > I have create a new helper U256IntoResultU8 for try to convert u256 to u8 and I use it like this:

`*popped[1]` is of type `u256` which is a struct `u256 {low:u128, high:u128}` so given that you only want to keep the last byte, you can do your bitwise operation on the lower u128 directly.

> I have create a new helper U256IntoResultU8 for try to convert u256 to u8 and I use it like this:
Actually, it's not useful in this case because `x & 0xFF` will _always_ fit in a u8

enitrat: try this then

`let value: u8 = (*popped.at(1).low & 0xFF).try_into().unwrap();`

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -77,3 +77,82 @@ fn test_exec_pop_should_stack_underflow() {
         result.unwrap_err() == EVMError::StackError(STACK_UNDERFLOW), 'should return StackUnderflow'
     );
 }
+
```
enitrat, line 80: can you add one extra test with an offset > 31?

enitrat (approved):
lgtm

Issue #299, opened 2023-09-07
feat: A3 - LOG3 Opcode


Eikix:
# Append log record with three topics

Official Specification: https://github.com/ethereum/execution-specs/blob/master/src/ethereum/shanghai/vm/instructions/log.py#L33

Kakarot v0 implementation:
https://github.com/kkrt-labs/kakarot/blob/44dabfcab05a87f85bcfe0a457809059d11d83fb/src/kakarot/instructions/logging_operations.cairo#L83

And test cases from Geth:
https://github.com/ethereum/go-ethereum/tree/master/core/vm/testdata

Test cases from Kakarot v0:
https://github.com/kkrt-labs/kakarot/blob/44dabfcab05a87f85bcfe0a457809059d11d83fb/src/kakarot/instructions/logging_operations.cairo

Context on Logs in Cairo and Starknet: 
https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/starknet-events/
https://book.cairo-lang.org/ch99-01-03-03-contract-events.html?highlight=event#events

[...]

PR #322, opened 2023-09-07
feat: A0-A4 - LOG0-4 Opcode 

Quentash:
Implementation of the LOG0, LOG1, LOG2, LOG3, LOG4 opcode.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #296 #297 #298 #299 #300

## What is the new behavior?

-
-
-

- [ ] Yes
- [X] No


Quentash:
I have just realised I branched from the wrong source in git. But since I will anyway pull from my fork and merge with it before my final push, it shouldn't be problematic, right ?

Anyway, I made an early push to talk about a gas issue I have been facing. TL;DR : load_n() use a lot of gaz even for small byte size. I have confronted 2 solutions against my tests, here is the one using load_n() to retrieve the datas:

```
        let mut datas: Array<felt252> = ArrayTrait::new();

        let mut loaded: Array<u8> = ArrayTrait::new();
        self.memory.load_n(size, ref loaded, offset);
        let mut element: felt252 = 0;
        let mut i = 0;
        loop {
            if i == size {
                break;
            };
            element *= 256;
            element += (*loaded[i]).into();
            
            i+=1;
            if i%31 == 0 {
                datas.append(element);
                element = 0;
            };
        };
        if size%31 != 0 {
            datas.append(element);
        }
```
And here is the test output : 
```
test_exec_log0 ... ok (gas usage est.: 2252000)
test_exec_log1 ... ok (gas usage est.: 2336270)
test_exec_log2 ... fail (gas usage est.: 1999965750) ( augmented the test limit but still OOG )
test_exec_log3_and_log4 ... fail (gas usage est.: 1999929590) ( augmented the test limit but still OOG )
```

The other solution is the one in the pushed file.
It gives this output :
```
test_exec_log0 ... ok (gas usage est.: 887710)
test_exec_log1 ... ok (gas usage est.: 1988326)
test_exec_log2 ... ok (gas usage est.: 1850076)
test_exec_log3_and_log4 ... ok (gas usage est.: 4058812)
```

It seems that loading the memory by 32 bytes chunk multiple times is more efficient than load N bytes in once.
Sorry for such a long comment.

Quentash:
I just noticed that I branched from the wrong source, sorry. I will merge my fork with my final solution anyway to be sure to only bring the modifications related to this PR.

Anyway, I have had a gas issue using load_n() and was somehow "forced" to load the datas by 32 bytes chunk multiple times if needed as it seems cheaper than load N bytes in once.

Here are the comparaison :
Using load_n ( logic in comment below ) :
test_exec_log0 ... ok (gas usage est.: 2252000)
test_exec_log1 ... ok (gas usage est.: 2336270)
test_exec_log2 ... fail (gas usage est.: 1999965750)
test_exec_log3_and_log4 ... fail (gas usage est.: 1999929590)

Using load ( logic actually used in source )
test_exec_log0 ... ok (gas usage est.: 887710)
test_exec_log1 ... ok (gas usage est.: 1988326)
test_exec_log2 ... ok (gas usage est.: 1850076)
test_exec_log3_and_log4 ... ok (gas usage est.: 4058812)

crates/evm/src/instructions/logging_operations.cairo
```diff
@@ -2,49 +2,115 @@
...
-    internal::exec_log_i(ref context, 0);
-}
+        let popped = self.stack.pop_n(2 + topics_len.into())?;
+        let offset: u32 = Into::<u256, Result<u32, EVMError>>::into(*popped[0])?;
+        let size: u32 = Into::<u256, Result<u32, EVMError>>::into(*popped[1])?;
 
+        // Feed topics array for the new event
+        let mut topics: Array<u256> = Default::default();
+        let mut i = 0;
+        loop {
```
enitrat, line 31: Can you create a function `fn concat(ref self: Array<T>, other: Span<T>)` under `utils/src/helpers/ArrayExtension` that appends all elements of `other` in `self`?

crates/evm/src/instructions/logging_operations.cairo
```diff
@@ -2,49 +2,115 @@
...
-    internal::exec_log_i(ref context, 2);
-}
+        // Feed data array for the new event
+        let mut datas: Array<felt252> = Default::default();
+        let mut i = 0;
+        loop {
+            if 31 + i > size {
+                if i != size {
+                    let (mut loaded, _) = self.memory.load(offset + i);
+                    let mut chunk: Array<u8> = u256_to_bytes_array(loaded);
```
enitrat, line 47: is it better to `load_n` bytes directly instead of loading a word and splitting it?

crates/evm/src/instructions/logging_operations.cairo
```diff
@@ -2,49 +2,115 @@
...
+        };
 
-/// 0xA2 - LOG2 
-/// Append log record with two topics.
-/// # Specification: https://www.evm.codes/#a2?fork=shanghai
-fn exec_log2(ref context: ExecutionContext) {
-    internal::exec_log_i(ref context, 2);
-}
+        // Feed data array for the new event
+        let mut datas: Array<felt252> = Default::default();
```
enitrat, line 41: data doesn't take an `s` in the plural form 😄

crates/evm/src/instructions/logging_operations.cairo
```diff
@@ -2,49 +2,115 @@
...
+            loaded /= 256;
+            datas.append(loaded.try_into().unwrap());
+            i += 31;
+        };
+
+        // Create and store the new event in the dynamic context
+        let event: Event = Event { keys: topics, data: datas };
+
+        let mut dyn_ctx = self.dynamic_context.unbox();
+        dyn_ctx.events.append(event);
```
enitrat, line 72: You can create a `set_events` method in `context.cairo`

See PR #328 to see how `set_return_data` was used

crates/evm/src/instructions/logging_operations.cairo
enitrat: Actually can't we just `popped.slice(2, popped.len() -2)` ?

crates/evm/src/instructions/logging_operations.cairo
```diff
@@ -2,49 +2,115 @@
...
 
-/// 0xA2 - LOG2 
-/// Append log record with two topics.
-/// # Specification: https://www.evm.codes/#a2?fork=shanghai
-fn exec_log2(ref context: ExecutionContext) {
-    internal::exec_log_i(ref context, 2);
-}
+        // Feed data array for the new event
+        let mut datas: Array<felt252> = Default::default();
+        let mut i = 0;
```
enitrat, line 42: I have to admit I'm not really sure what's going on here. Is this just loading `size` bytes from memory starting at offset? If yes, why not use `load_n` ?

crates/evm/src/tests/test_utils.cairo
```diff
@@ -84,6 +84,20 @@ fn setup_execution_context_with_calldata(calldata: Span<u8>) -> ExecutionContext
     )
 }
 
+fn setup_read_only_execution_context() -> ExecutionContext {
```
enitrat, line 87: this is only used once. We can get rid of this, and where required, do 

```rust
ctx.read_only = true;
```

crates/evm/src/errors.cairo
```diff
@@ -14,12 +14,16 @@ const TYPE_CONVERSION_ERROR: felt252 = 'Kakarot: type conversion error';
 // RETURNDATA
 const RETURNDATA_OUT_OF_BOUNDS_ERROR: felt252 = 'Kakarot: ReturnDataOutOfBounds';
 
+// EVM STATE
+const STATE_MODIFICATION_ERROR: felt252 = 'Kakarot: StateModificationError';
```
enitrat, line 18: New convention: KKT instead of Kakarot to keep messages shorter
```suggestion
const STATE_MODIFICATION_ERROR: felt252 = 'KKT: StateModificationError';
```

enitrat:
Ok my bad I didn't see the comment you added on the VSCode UI 😛

If `load_n` uses more gas that doing `load` and then splitting bytes, then we should refactor load_n

crates/evm/src/instructions/logging_operations.cairo
Quentash: The fact `popped.slice(2, popped.len() -2)` returns a Span, I can't just assign it to `topics` which is an array.
But it therefore fits perfectly to do this :
`let mut topics: Array<u256> = Default::default();
ArrayExtensionTrait::concat(ref topics, popped.span().slice(2,popped.len()-2));` ( This already existed, mb )

enitrat: ok, then instead of using `slice`, you will do the following

```rust

        let offset = self.stack.pop_usize();
        let size = self.stack.pop_usize();
        let mut topics = self.stack.pop_n(topics_len.into())?;

        // Feed data array for the new event
```

crates/evm/src/tests/test_utils.cairo
Quentash: Isn't `read_only` a static context attribute and therefore cannot be assigned that way ?
It gives me "invalid left hand side assignement" when trying to do so, that is why I created that helper.
An other solution I thought about was creating the execution context without the helpers by calling `ExecutionContextTrait::new()` ?

Is there a way to change the read_only that I may ignore ? Are we in the only scenario that (will) requires test on read_only context ?

enitrat:
> I made an early push to talk about a gas issue I have been facing. TL;DR : load_n() use a lot of gaz even for small byte size. I have confronted 2 solutions against my tests, here is the one using load_n() to retrieve the data:

What you're trying to do is to load felts from memory, right?

Quentash:
> > I made an early push to talk about a gas issue I have been facing. TL;DR : load_n() use a lot of gaz even for small byte size. I have confronted 2 solutions against my tests, here is the one using load_n() to retrieve the data:
> 
> What you're trying to do is to load felts from memory, right?

Yes, I need to fill a data of felt252.
I will push the 2 solutions I have ( load and loadn ) so you can see if I'm doing something wrong that would alterate the gas cost, but for my usage, loadn() seems to be havier than load().

Quentash:
The last commit ( when writing this ) contains 2 functions in logging_operations.cairo.
One using an algorithm that will load chunks of 32 bytes, format it to keep the 31 required bytes in a felt252 that is then appended to the data array.
image

One using an algorithm that will load N Bytes and then concat them into a felt that will be appended in the data array.
image

Something I don't understand tho, is why the "read_only" test fails with 200k more gas although the revert is at the beginning of the function.

I'm well aware my code still contains stuff to optimize (those 2 functions for example) and I would like to add test for the internal function we will keep and for the `set_event` one added in context.
My internet just comes and goes so I felt like pushing this now so you may look at it too.

Quentash:
> The last commit ( when writing this ) contains 2 functions in logging_operations.cairo. One using an algorithm that will load chunks of 32 bytes, format it to keep the 31 required bytes in a felt252 that is then appended to the data array. image
> 
> One using an algorithm that will load N Bytes and then concat them into a felt that will be appended in the data array. image
> 
> Something I don't understand tho, is why the "read_only" test fails with 200k more gas although the revert is at the beginning of the function (and therefore shouldn't call any of the 2 functions )
> 
> I'm well aware my code still contains stuff to optimize (those 2 functions for example) and I would like to add test for the internal function we will keep and for the `set_event` one added in context. My internet just comes and goes so I felt like pushing this now so you may look at it too.

crates/evm/src/instructions/logging_operations.cairo
enitrat: Actually I think the Event model is not optimal at all.
In kakarot0 we were loading bytes from memory and logging events data as an array of bytes (Array<u8>). This would avoid us the cost of converting an Array<u256> to an Array<felt252> for nothing

crates/evm/src/tests/test_utils.cairo
enitrat: We can add a `set_read_only(value: bool)` method on `ExecutionContext` to change its value

Quentash:
Changed Event structure in model.cairo so Data is now an Array of u8, instead of felt. ( Similar to what kkt0 was  ).
Adapted code to this new structure.
Here is a screenshot of my tests results tho : 
image
You can notice it still looks more expensive this way than with the u256->felt way although, now, I only have to call load_n() for data.

crates/evm/src/context.cairo
```diff
@@ -343,6 +343,29 @@ impl ExecutionContextImpl of ExecutionContextTrait {
...
+    /// Store a new event in the dynamic context using given
+    /// topics and data.
+    /// The dynamic context has to be recreated to be modified.
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - The context to which the event will be added
+    /// * `topics` - Topics of the event
+    /// * `data` - Data of the event
+    fn set_events(ref self: ExecutionContext, topics: Array<u256>, data: Array<u8>) {
```
Eikix, line 363:
```suggestion
    fn append_event(ref self: ExecutionContext, event: Event) {
```

crates/evm/src/context.cairo
```diff
@@ -343,6 +343,29 @@ impl ExecutionContextImpl of ExecutionContextTrait {
...
+        let event: Event = Event { keys: topics, data };
```
Eikix, line 364: then this logic will have to change

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -0,0 +1,187 @@
...
+    ctx.stack.push(0x1F);
+    ctx.stack.push(0x00);
+
+    // When
+    let result = ctx.exec_log0();
+
+    // Then
+    assert(result.is_ok(), 'should have succeeded');
+    assert(ctx.stack.len() == 4, 'stack size should be 4');
+    let mut events = ctx.events();
```
enitrat, line 32: add an empty line above this (in all tests)

crates/evm/src/errors.cairo
```diff
@@ -14,13 +14,17 @@ const RETURNDATA_OUT_OF_BOUNDS_ERROR: felt252 = 'KKT: ReturnDataOutOfBounds';
 // JUMP
 const INVALID_DESTINATION: felt252 = 'KKT: invalid JUMP destination';
 
+// EVM STATE
+const STATE_MODIFICATION_ERROR: felt252 = 'KKT: StateModificationError';
```
Eikix, line 18: open question that just came to mind, why to we prefix our errors by KKT?
doesn't seem to add any information

crates/evm/src/instructions/logging_operations.cairo
```diff
@@ -1,48 +1,80 @@
...
+    ///
+    /// * `self` - The context to which the event will be added
+    /// * `topics_len` - The amount of topics to pop from the stack
+    fn exec_log_i(ref self: ExecutionContext, topics_len: u8) -> Result<(), EVMError> {
+        // Revert if the transaction is in a read only context
+        if self.read_only() {
+            return Result::Err(EVMError::StateModificationError(STATE_MODIFICATION_ERROR));
+        }
+
+        let offset = (self.stack.pop_usize())?;
```
Eikix, line 69: why do we need these parentheses?

crates/evm/src/errors.cairo
enitrat: so that we can track the origin of the error better e.g. if the error is from our EVM itself or from something that happened during tx exec or something related to corelib? I think it makes sense

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -0,0 +1,187 @@
...
+
+#[test]
+#[available_gas(20000000)]
+fn test_exec_log0() {
+    // Given
+    let mut ctx = setup_execution_context();
+
+    ctx.memory.store(BoundedInt::<u256>::max(), 0);
+
+    ctx.stack.push(BoundedInt::<u256>::max());
```
Eikix, line 19: can  you add comments to explain these first 4 values in the stack won't be used? it's not so clear

Eikix: to be fair, we can even delete those,
it's not the responsibility of this test to check that the stack isn't popped more than expected
-> if the test executes, it means we've not ecountered a `StackUnderflow` afaik

So we can just push the right amount of elements

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -0,0 +1,187 @@
...
+
+#[test]
+#[available_gas(20000000)]
+fn test_exec_log1() {
+    // Given
+    let mut ctx = setup_execution_context();
+
+    ctx.memory.store(BoundedInt::<u256>::max(), 0);
+
+    ctx.stack.push(BoundedInt::<u256>::max());
```
Eikix, line 50: same remark

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
enitrat: agreed

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -0,0 +1,187 @@
...
+    assert(*event.keys[0] == 0x0123456789ABCDEF, 'event key is not correct');
+    assert(*event.keys[1] == BoundedInt::<u256>::max(), 'event key is not correct');
+    assert(event.data.len() == 5, 'event should have 5 bytes');
+    assert(*event.data[0] == 0xff, 'event byte should be max_u8');
+    assert(*event.data[4] == 0xff, 'event byte should be max_u8');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_exec_log3_and_log4() {
```
enitrat, line 135: why are 3 and 4 tested together?

Eikix:
> Changed Event structure in model.cairo so Data is now an Array of u8, instead of felt. ( Similar to what kkt0 was ). Adapted code to this new structure. Here is a screenshot of my tests results tho : image You can notice it still looks more expensive this way than with the u256->felt way although, now, I only have to call load_n() for data.

I really like this! Great thinking and it's a win for the simplicity of the code. Make it work then make it fast.
Plus, I notice log2, 3 and 4 are more performant with your change. Good one

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
Quentash: I felt like it was a "decent idea" to check if 2 log operations could run one after an other.
I agree there isn't much plus-value tho, will split them

crates/evm/src/instructions/logging_operations.cairo
Quentash: We don't. Leftovers from copy/pasta :p

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
Quentash: Hooo, ok !
I kinda overkilled the test then, I will simplify them, mb

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
Quentash: Ho, I remember, I've put them in one single test to be sure the event is "appended" to the list of events ( and do not erase the previous one, etc... ).
I still think this is a decent idea, so I will leave them like that but add an individual test for log3 and log4.

I will ultimately do as you require tho !

crates/evm/src/context.cairo
```diff
@@ -343,6 +343,26 @@ impl ExecutionContextImpl of ExecutionContextTrait {
...
+    }
+
+    /// Store a new event in the dynamic context.
+    /// The dynamic context has to be recreated to be modified.
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - The context to which the event will be added
+    /// * `event` - Event to append
+    fn append_event(ref self: ExecutionContext, event: Event) {
```
Eikix, line 361: @enitrat do we have a rule of thumb for which methods should be inline or those which shouldn't?

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -0,0 +1,257 @@
...
+    assert(result.is_ok(), 'should have succeeded');
+    assert(ctx.stack.len() == 0, 'stack should be empty');
+
+    let mut events = ctx.events();
+    assert(events.len() == 1, 'context should have one event');
+
+    let event = events.pop_front().unwrap();
+    assert(event.keys.len() == 0, 'event should not have keys');
+    assert(event.data.len() == 31, 'event should have 31 bytes');
+    assert(*event.data[0] == 0xff, 'event byte should be max_u8');
```
Eikix, line 35: should we have a util to assess the equality between an event and an arbitrary amount of data? how performant is testing parts of the event.data byte array?

crates/evm/src/context.cairo
enitrat: depends on the size of the code inside the function, and the size of data passed to the function. In this case I think it's probably better inlined as the size is quite short? Unless boxing/unboxing generates too much CASM 🤔

Quentash:
Now comparing all bytes in tests. ( Array vs Array )
Made append_event inline.

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -132,9 +134,14 @@ fn test_exec_log3() {
     assert(*event.keys[2] == 0x00, 'event key is not correct');
 
     assert(event.data.len() == 40, 'event should have 40 bytes');
-    assert(*event.data[0] == 0xff, 'event byte should be max_u8');
-    assert(*event.data[4] == 0xff, 'event byte should be max_u8');
-    assert(*event.data[39] == 0xef, 'event byte should be 0xEF');
+    let data_expected = u256_to_bytes_array(BoundedInt::<u256>::max()).span().slice(0, 32);
```
Eikix, line 137: why do we need to slice here?

Quentash: We dont, jeez, sorry

Quentash:
Not slicing 32 bytes into 32 bytes anymore !

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -0,0 +1,350 @@
...
+    let event = events.pop_front().unwrap();
+    assert(event.keys.len() == 3, 'event should have 3 keys');
+    assert(*event.keys[0] == 0x0123456789ABCDEF, 'event key is not correct');
+    assert(*event.keys[1] == BoundedInt::<u256>::max(), 'event key is not correct');
+    assert(*event.keys[2] == 0x00, 'event key is not correct');
+
+    assert(event.data.len() == 40, 'event should have 40 bytes');
+    let data_expected = u256_to_bytes_array(BoundedInt::<u256>::max()).span();
+    assert(event.data.span().slice(0, 32) == data_expected, 'event data are incorrect');
+    let data_expected = u256_to_bytes_array(
```
Eikix, line 139: why not 
array![0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF].span()

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
```diff
@@ -0,0 +1,350 @@
...
+    let event = events.pop_front().unwrap();
+    assert(event.keys.len() == 4, 'event should have 4 keys');
+    assert(*event.keys[0] == 0x0123456789ABCDEF, 'event key is not correct');
+    assert(*event.keys[1] == BoundedInt::<u256>::max(), 'event key is not correct');
+    assert(*event.keys[2] == 0x00, 'event key is not correct');
+    assert(*event.keys[3] == BoundedInt::<u256>::max(), 'event key is not correct');
+
+    assert(event.data.len() == 10, 'event should have 10 bytes');
+    let data_expected = u256_to_bytes_array(
+        0x0123456789ABCDEF000000000000000000000000000000000000000000000000
```
Eikix, line 182: same question

Eikix (commented):
Lgtm, almost ready to merge, I forgot if I asked you already, did you look at test from execution-specs and geth?

crates/evm/src/tests/test_instructions/test_logging_operations.cairo
Quentash: I didn't know this syntax, thanks for it !

Quentash:
> Lgtm, almost ready to merge, I forgot if I asked you already, did you look at test from execution-specs and geth?

Yup I did !
Added a few tests after that ; One with size and offset at 0 + 2 failings tests because of one of those value being too big.
I've watched there : https://github.com/ethereum/execution-specs/blob/master/tests/byzantium/vm/test_logging_operations.py
and there : https://github.com/ethereum/legacytests/blob/97de5bce3a257a4b219666134df15b6d4d47f1f5/src/LegacyTests/Constantinople/VMTestsFiller/vmLogTest/log0_logMemStartTooHighFiller.json

It looks like they have more tests but they are actually testing every test scenarios on EACH LOG op. Since all our log use the log_n functions, I felt like it was ok to reunite thoses test. ( I don't see why a test on out of bound memory would succeed on LOG1 but not LOG2 ? ) Still open to suggestion tho.

Eikix:
> > Lgtm, almost ready to merge, I forgot if I asked you already, did you look at test from execution-specs and geth?
> 
> Yup I did ! Added a few tests after that ; One with size and offset at 0 + 2 failings tests because of one of those value being too big. I've watched there : https://github.com/ethereum/execution-specs/blob/master/tests/byzantium/vm/test_logging_operations.py and there : https://github.com/ethereum/legacytests/blob/97de5bce3a257a4b219666134df15b6d4d47f1f5/src/LegacyTests/Constantinople/VMTestsFiller/vmLogTest/log0_logMemStartTooHighFiller.json
> 
> It looks like they have more tests but they are actually testing every test scenarios on EACH LOG op. Since all our log use the log_n functions, I felt like it was ok to reunite thoses test. ( I don't see why a test on out of bound memory would succeed on LOG1 but not LOG2 ? ) Still open to suggestion tho.

I think that works for me!
Btw your rebase seems to have failed since it's writing Clement's commits instead of a `merge main branch` commit

enitrat:
tbh I think rebase is better than merge when bringing remote changes into a PR

Eikix:
> tbh I think rebase is better than merge when bringing remote changes into a PR

some merge are merge.rebase afaik, idk how gh handles those

Quentash:
Updated the code to fit the new rebase.
I'm just not very convinced by my "set_read_only" so if you'd mind double-checking it. Am I doing what I am supposed to do or am I overkilling something ?

crates/evm/src/machine.cairo
```diff
@@ -189,6 +189,22 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
         current_call_ctx.read_only()
     }
 
+    #[inline(always)]
+    fn set_read_only(ref self: Machine, value: bool) {
+        let mut current_call_ctx = self.call_context();
```
Eikix, line 194: maybe @enitrat can pitch in, but I think here self.call_context() copies the current callcontext,
might be cheaper to do:

```
let mut current_execution_ctx = self.current_context.unbox();
let mut current_call_ctx = current_execution_ctx.call_context.unbox();
current_call_ctx.read_only = value;
current_execution_ctx.call_context = BoxTrait::new(current_call_ctx);
self.current_context = BoxTrait::new(current_execution_ctx);
```

Eikix: I think you can simply add a test-util that creates a ExecutionContext with read_only true and then have a setup_machine for it!
like setup_execution_context_with_read_only() 

even later on we can introduce a builder pattern which @greged93 is a master of for our test utils ;D

crates/evm/src/machine.cairo
```diff
@@ -189,6 +189,22 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
...
+        let mut current_execution_ctx = self.current_context.unbox();
+        current_call_ctx.read_only = value;
+        current_execution_ctx.call_context = BoxTrait::new(current_call_ctx);
+        self.current_context = BoxTrait::new(current_execution_ctx);
+    }
+
+    #[inline(always)]
+    fn append_event(ref self: Machine, event: Event) {
```
Eikix, line 202: can't we duplicate this function in Machine and in Execution context, in order to do:

```
let mut current_execution_ctx = self.current_context.unbox();
current_execution_ctx.append_event(event);
self.current_context = BoxTrait::new(current_execution_ctx);
```

crates/evm/src/errors.cairo
```diff
@@ -36,7 +40,8 @@ impl EVMErrorIntoU256 of Into<EVMError, u256> {
             EVMError::JumpError(error_message) => error_message.into(),
             EVMError::NotImplemented => 'NotImplemented'.into(),
             // TODO: refactor with dynamic strings once supported
-            EVMError::UnknownOpcode => 'UnknownOpcode'.into()
+            EVMError::UnknownOpcode => 'UnknownOpcode'.into(),
+            EVMError::StateModificationError(error_message) => error_message.into(),
```
Eikix, line 44: Is this the same error as EELS?
https://github.com/ethereum/execution-specs/tree/master/src/ethereum/shanghai/vm

crates/evm/src/machine.cairo
enitrat: > I think we simply don't want to be able to change this after creating an execution context

If it's just used in tests then yes it's best not to have it in the source. Otherwise, I agree with your first recommended solution

Quentash: Hahaha, we created a set_read_only() because I first created a test utils like setup_execution_context_with_read_only() but it felt inappropriate since it was only used once (https://github.com/kkrt-labs/kakarot-ssj/pull/322#discussion_r1324918506)
I agree with you that this variable should not be changeable that easily tho.

crates/evm/src/machine.cairo
Quentash: Sorry but I'm afraid I don't understand, the function is already in machine.cairo, or are you saying something else ?

crates/evm/src/errors.cairo
Quentash: No, this is the message that was used in KKT0, the one from your link looks like it raises a " WriteInStaticContext " in thie situation. I put the one you prefer.

crates/evm/src/machine.cairo
Quentash: It is fine, I agree with everything you say anyway, I will go back to the setup_read_only_context so the read_only state isn't updatable from the source code

crates/evm/src/machine.cairo
Quentash: Ho, you mean ; I implement append_event in ExecutionContext, and re-write it in machine so that it call the one from the execution context, is that it ?

Quentash:
- Removed the `set_read_only()` function and added a `setup_machine_with_read_only()`. Which actually kind of duplicate a lot of code, I didn't find a way to give functions parameters a default value, is it possible in Cairo ? If yes, maybe refactoring the `setup_machine()` with a `read_only `paramater set to false by default would be a cleaner solution ?

- Duplicated the `append_event()` function in `Machine` from the `ExecutionContext` and made it use the execution context one.

- Left the error as `StateModificationError` as it was used in Kakarot0 as I find it pretty explicit. Tell me if you want me to change it to `WriteInStaticContext` for the sake of consistency with EELS.

crates/evm/src/machine.cairo
enitrat: @Eikix

crates/evm/src/errors.cairo
Eikix: Let's try and put exactly the same errors as EELS

crates/evm/src/tests/test_utils.cairo
```diff
@@ -148,3 +148,43 @@ fn setup_machine_with_calldata(calldata: Span<u8>) -> Machine {
         storage_journal: Default::default(),
     }
 }
+
+fn setup_machine_with_read_only() -> Machine {
+    Machine {
```
Eikix, line 153: to be fair maybe a mischievous trick here would be to write:
```
fn setup_machine_with_read_only() -> Machine {
     let mut machine = setup_machine();
    machine.set_read_only(true);
    }

and same for all the others?

Comment (enitrat) [reply crates/evm/src/tests/test_utils.cairo:153]: but we said no public `set_read_only` method? Only for testing we want to be able to change the read_only state?

**Comment by enitrat** [comment]
>  I didn't find a way to give functions parameters a default value, is it possible in Cairo ?

I don't think so. but what you can do is something like 
```rust
fn add(a: Option<i32>, b: Option<i32>) -> i32 {
    a.unwrap_or(1) + b.unwrap_or(2)
}
```

which is equivalent to
```rust
fn add(a: int = 1, b: int = 2) { a + b }
```

on a side note unwrap_or is not implemented in `OptionTrait`. so you would need to write your own custom impl of it before it's merged in Cairo

Eikix:
> > I didn't find a way to give functions parameters a default value, is it possible in Cairo ?
> 
> I don't think so. but what you can do is something like
> 
> ```rust
> fn add(a: Option<i32>, b: Option<i32>) -> i32 {
>     a.unwrap_or(1) + b.unwrap_or(2)
> }
> ```
> 
> which is equivalent to
> 
> ```rust
> fn add(a: int = 1, b: int = 2) { a + b }
> ```
> 
> on a side note unwrap_or is not implemented in `OptionTrait`. so you would need to write your own custom impl of it before it's merged in Cairo

Opened a feature request for Cairo corelib

https://github.com/starkware-libs/cairo/issues/4196

crates/evm/src/tests/test_utils.cairo
Eikix: oops i'm so dumb^^

then can do:
```
fn setup_machine_with_read_only() -> Machine {
     let mut machine = setup_machine();
     let mut current_ctx = machine.current_context.unbox();
     let mut current_call_context = ....
     // set read_only
    }
```

Quentash:
- Modified the error message to be consistent with EELS one.
- Updated the setup_machine_with_read_only() so that it changes the read_only variable after setting up a normal machine instead of duplicating a lot of code just to create a call_context with read_only set to true.

Eikix (approved):
lgtm 🔥

Eikix (approved):
lgtm

PR #328, opened 2023-09-08
feat(evm): add exec_return opcode

LucasLvy:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #301 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/evm/src/instructions/system_operations.cairo
+++ b/crates/evm/src/instructions/system_operations.cairo
@@ -9,6 +9,7 @@ use evm::context::{ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionC
 use evm::memory::MemoryTrait;
 use evm::stack::StackTrait;
 use evm::errors::EVMError;
+use evm::helpers::U256IntoResultU32;
 
 #[generate_trait]
 impl SystemOperations of SystemOperationsTrait {
@@ -34,16 +35,11 @@ impl SystemOperations of SystemOperationsTrait {
     /// RETURN
     /// # Specification: https://www.evm.codes/#f3?fork=shanghai
     fn exec_return(ref self: ExecutionContext) -> Result<(), EVMError> {
-        let offset = self.stack.pop()?;
-        let size = self.stack.pop()?;
+        let popped = self.stack.pop_n(2)?;
+        let offset: u32 = Into::<u256, Result<u32, EVMError>>::into(*popped[0])?;
+        let size: u32 = Into::<u256, Result<u32, EVMError>>::into(*popped[1])?;
         let mut return_data = array![];
-        self
-            .memory
-            .load_n(
-                size.low.try_into().expect('Too much return data'),
-                ref return_data,
-                offset.low.try_into().expect('Return data offset > u32')
-            );
+        self.memory.load_n(size, ref return_data, offset);
         self.set_return_data(return_data);
         self.stop();
         Result::Ok(())
--- a/crates/evm/src/instructions/system_operations.cairo
+++ b/crates/evm/src/instructions/system_operations.cairo
@@ -36,15 +36,15 @@ impl SystemOperations of SystemOperationsTrait {
     fn exec_return(ref self: ExecutionContext) -> Result<(), EVMError> {
         let offset = self.stack.pop()?;
         let size = self.stack.pop()?;
-        let mut dynamic_context = self.dynamic_context.unbox();
+        let mut return_data = array![];
         self
             .memory
             .load_n(
                 size.low.try_into().expect('Too much return data'),
-                ref dynamic_context.return_data,
+                ref return_data,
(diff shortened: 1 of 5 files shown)


crates/evm/src/instructions/system_operations.cairo
```diff
@@ -1,46 +1,87 @@
...
 
-/// RETURN
-/// # Specification: https://www.evm.codes/#f3?fork=shanghai
-fn exec_return(ref context: ExecutionContext) {}
+    /// RETURN
+    /// # Specification: https://www.evm.codes/#f3?fork=shanghai
+    fn exec_return(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let offset = self.stack.pop()?;
+        let size = self.stack.pop()?;
+        let mut dynamic_context = self.dynamic_context.unbox();
```
enitrat, line 39: use the `return_data` method on `self` instead

crates/evm/src/tests/test_instructions/test_system_operations.cairo
```diff
@@ -0,0 +1,24 @@
+use evm::instructions::SystemOperationsTrait;
+use evm::stack::StackTrait;
+use evm::tests::test_utils::setup_execution_context;
+use evm::instructions::MemoryOperationTrait;
+use evm::execution::{ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionContextDestruct};
```
enitrat, line 5:
```suggestion
use evm::context::{ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionContextDestruct};
```

crates/evm/src/instructions/system_operations.cairo
enitrat: e.g. 

```rust
let mut return_data = array![];
        self
            .memory
            .load_n(
                size.low.try_into().expect('Too much return data'),
                ref return_data,
                offset.low.try_into().expect('Return data offset > u32')
            );
        self.set_return_data(return_data);
 ```

File: crates/evm/src/instructions/system_operations.cairo [hunk 3]
```diff
@@ -1,46 +1,87 @@
...
+        Result::Ok(())
+    }
 
-/// RETURN
-/// # Specification: https://www.evm.codes/#f3?fork=shanghai
-fn exec_return(ref context: ExecutionContext) {}
+    /// RETURN
+    /// # Specification: https://www.evm.codes/#f3?fork=shanghai
+    fn exec_return(ref self: ExecutionContext) -> Result<(), EVMError> {
+        let offset = self.stack.pop()?;
```
enitrat, line 37: we have a "safe cast" that returns an error if the typecast fails 

prefer something like

```rust
 let popped = self.stack.pop_n(3)?;
        let dest_offset: u32 = Into::<u256, Result<u32, EVMError>>::into((*popped[0]))?;
        let offset: u32 = Into::<u256, Result<u32, EVMError>>::into((*popped[1]))?;
        let size: u32 = Into::<u256, Result<u32, EVMError>>::into((*popped[2]))?;
```

crates/evm/src/tests/test_instructions/test_system_operations.cairo
```diff
@@ -0,0 +1,24 @@
...
+use utils::helpers::load_word;
+
+#[test]
+#[available_gas(20000000)]
+fn test_exec_return() {
```
enitrat, line 10: can you also add one test with an offset

crates/evm/src/instructions/system_operations.cairo
LucasLvy: this looks very bad, a `try_into` seems to make more sense

enitrat: why do you think so?

LucasLvy: because it's the purpose of a `try_into` output an error if it fails

enitrat: Yes, I agree, the only problem is that the corelib trait returns an Option<T> and not a Result<T, E>. Now that I think about it it's super weird that it was designed this way

Eikix: we could ask them to either transform try_into or implement `ok_or` which converts an option into a result in Rust

enitrat: @LucasLvy Can you implement the recommendation pls, this will likely be refactored later anyway

crates/evm/src/tests/test_instructions/test_system_operations.cairo
LucasLvy: no.

enitrat (approved):
LGTM 🔥

Issue #325, opened 2023-09-08
dev: add `store_byte` method to memory


enitrat:
Memory interactions happening through `MSTORE8` store a simple 1-byte value. Calling `store_n` should only be done on `XCOPY` opcodes.

We should introduce a `store_byte` method to the memory impl that stores a single 1-byte value to memory at a specific offset.

PR #329, opened 2023-09-08
feat: store_byte memory method

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #325 

## What is the new behavior?

- Adds a `store_byte` method to memory for MSTORE8 opcode
- Fixes total gas accounting in python script.
-

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -11,7 +11,7 @@ use utils::constants::{
 use cmp::{max};
 use utils::{
     helpers, helpers::SpanExtensionTrait, helpers::ArrayExtensionTrait, math::Exponentiation,
-    math::WrappingExponentiation
+    math::WrappingExponentiation, math::Bitshift
 };
 use debug::PrintTrait;
 
@@ -95,9 +95,9 @@ impl MemoryImpl of MemoryTrait {
         // As the memory words are in big-endian order, we need to convert our left-based offset
         // to a right-based one.
         let right_offset = 15 - left_offset;
-        let mask: u128 = 0xFF_u128.shl(right_offset.into() * 8);
+        let mask: u128 = 0xFF * helpers::pow2(right_offset.into() * 8);
 
-        // First erase byte value at leff, then set the new value using bitwise ops
+        // First erase byte value at offset, then set the new value using bitwise ops
         let word: u128 = self.items.get(index.into());
         let new_word = (word & ~mask) | (value.into().shl(right_offset.into() * 8));
         self.items.insert(index.into(), new_word);
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -1,6 +1,6 @@
 use integer::{
     u32_safe_divmod, u32_as_non_zero, u128_safe_divmod, u128_as_non_zero, u256_safe_div_rem,
-    u256_as_non_zero, BoundedInt
+    u256_as_non_zero
 };
 use utils::constants::{
     POW_256_0_U128, POW_256_1_U128, POW_256_2_U128, POW_256_3_U128, POW_256_4_U128, POW_256_5_U128,
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -15,7 +15,6 @@ use utils::{
 };
 use debug::PrintTrait;
 
-
 #[derive(Destruct, Default)]
 struct Memory {
     items: Felt252Dict<u128>,
@@ -90,17 +89,17 @@ impl MemoryImpl of MemoryTrait {
         self.bytes_len = cmp::max(new_min_bytes_len, self.bytes_len);
 
         // Get offset's memory word index and left-based offset of byte in word.
-        let (index, left_offset) = u32_safe_divmod(offset, u32_as_non_zero(16));
+        let (chunk_index, left_offset) = u32_safe_divmod(offset, u32_as_non_zero(16));
 
         // As the memory words are in big-endian order, we need to convert our left-based offset
-        // to a right-based one.
+        // to a right-based one.a
         let right_offset = 15 - left_offset;
         let mask: u128 = 0xFF * helpers::pow2(right_offset.into() * 8);
 
         // First erase byte value at offset, then set the new value using bitwise ops
-        let word: u128 = self.items.get(index.into());
+        let word: u128 = self.items.get(chunk_index.into());
         let new_word = (word & ~mask) | (value.into().shl(right_offset.into() * 8));
-        self.items.insert(index.into(), new_word);
+        self.items.insert(chunk_index.into(), new_word);
     }
 
 
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -77,13 +77,13 @@ impl MemoryImpl of MemoryTrait {
     }
 
 
-    /// Stores a single bytes into memory at a specified offset.
+    /// Stores a single byte into memory at a specified offset.
     ///
     /// # Arguments
     ///
-    /// * `self` - A mutable reference to the `Memory` instance to store the bytes in.
+    /// * `self` - A mutable reference to the `Memory` instance to store the byte in.
     /// * `value` - The byte value to store in memory.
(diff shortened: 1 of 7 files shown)


ClementWalter (dismissed):
see comments

enitrat:
Addressed comments

crates/utils/src/helpers.cairo
```diff
@@ -68,6 +68,269 @@ fn pow256_rev(i: usize) -> u256 {
     }
 }
 
+// COmputes 2**pow for 0 <= pow < 128.
+fn pow2(pow: usize) -> u128 {
```
Eikix, line 72: is this faster than an array! of all the powers?

```
fn pow2(pow: usize) -> Option<u128> {
   let powers = array![0x1, 0x2, ..., 0x80000000000000000000000000000000].span();
   powers.get(usize)

Comment (enitrat) [reply crates/utils/src/helpers.cairo:72]: yes

File: crates/evm/src/tests/test_memory.cairo [hunk 2]
```diff
@@ -518,3 +518,82 @@ fn test_store_padded_segment_should_add_n_elements_padded_with_offset_between_tw
...
+    assert(memory.bytes_len == 32, 'Wrong memory length');
+}
+#[test]
+#[available_gas(20000000)]
+fn test_store_byte_should_store_byte_at_offset_2() {
+    // Given
+    let mut memory = MemoryTrait::new();
+
+    // When
+    memory.store_byte(0x01, 14);
```
Eikix, line 542:
```suggestion
    memory.store_byte(0xff, 14);
```

crates/evm/src/tests/test_memory.cairo
```diff
@@ -518,3 +518,82 @@ fn test_store_padded_segment_should_add_n_elements_padded_with_offset_between_tw
...
+
+    // Then
+    assert(memory.items[0] == 0x0100, 'Wrong value for word 0');
```
Eikix, line 545:
```suggestion
    assert(memory.items[0] == 0xff00, 'Wrong value for word 0');
```

crates/evm/src/tests/test_memory.cairo
Eikix: Proposing to change the stored byte for this second test for diversity

Eikix (dismissed):
looks good to me overall

crates/utils/src/helpers.cairo
Eikix: > yes

Btw even with pow == 127? Going through all the ifs is less expensive than accessing an array?

enitrat: even if you build the array with `array![]` macro you still have the underlying cost of doing several `array.append()` behind the scenes

Eikix: And we can't build it as a const in the global scope? Like the powers of 256 that @lambda-0x did?

enitrat: What he did is useful when you know exactly which one to call. The goal here is to have a dynamic behavior that depends on runtime values, not values known at compile time

lambda-0x: agree that `if else` chain is required here. 

on side note: i think we can make all of this constants and add it to `constants.cairo` and use them here. it already contains [`POW_2_127`](https://github.com/kkrt-labs/kakarot-ssj/blob/88d529493ad82a95971e179e6ac091943bb00f02/crates/utils/src/constants.cairo#L49).

so all of this exponential constants and in one place and its easier to reuse.

crates/utils/src/math.cairo
```diff
@@ -158,3 +191,37 @@ impl U256BitshiftImpl of Bitshift<u256> {
         self / 2.pow(shift)
     }
 }
+
+impl U128BitshiftImpl of Bitshift<u128> {
+    fn shl(self: u128, shift: u128) -> u128 {
+        if shift > 127 {
+            // 2.pow(shift) for shift > 255 will panic with 'u128_mul Overflow'
+            panic_with_felt252('u128_mul Overflow');
```
ClementWalter, line 199: why not just returning `0` ?

crates/utils/src/math.cairo
```diff
@@ -158,3 +191,37 @@ impl U256BitshiftImpl of Bitshift<u256> {
...
+        }
+        self * 2.pow(shift)
+    }
+
+    fn shr(self: u128, shift: u128) -> u128 {
+        if shift > 127 {
+            // 2.pow(shift) for shift > 255 will panic with 'u128_mul Overflow'
+            panic_with_felt252('u128_mul Overflow');
```
ClementWalter, line 207: why not just returning `0` ?

crates/utils/src/math.cairo
ClementWalter: anyway, just checked the default rust behavior and it does panick, so let's panic as well since we're in a math utils.

In the EVM, it's just 0

crates/utils/src/helpers.cairo
```diff
@@ -68,6 +68,269 @@ fn pow256_rev(i: usize) -> u256 {
     }
 }
 
+// COmputes 2**pow for 0 <= pow < 128.
```
ClementWalter, line 71: typo `COmputes`

crates/utils/src/math.cairo
enitrat: We have two separate traits for bitshift WrappingBitshift (which wraps around 0) and Bitshift (which panics if we have an overflow).

For all operations that are directly EVM-related (where overflows should return 0), we will use WrappingBitshift

crates/utils/src/math.cairo
enitrat: Indeed

crates/utils/src/helpers.cairo
Eikix: can you create an issue to perform this refactor?

> agree that `if else` chain is required here.
> 
> on side note: i think we can make all of this constants and add it to `constants.cairo` and use them here. it already contains [`POW_2_127`](https://github.com/kkrt-labs/kakarot-ssj/blob/88d529493ad82a95971e179e6ac091943bb00f02/crates/utils/src/constants.cairo#L49).
> 
> so all of this exponential constants and in one place and its easier to reuse.

enitrat: tbh I'd rather not have 127 constants if they're not reused in the code in more than one place

Eikix: > i think other places we are already using constants, just need to update in this PR. Should i still open an issue for it?

yes because in some sense, we could harmonise everything to use powers of two.
POW_256_ etc. would just become powers of 2

And helpers::powers_rev becomes somewhat useless

And it's the job of the dev to do x8 to get 256 conversion

Eikix: > tbh I'd rather not have 127 constants if they're not reused in the code in more than one place

we can create them as we go

crates/evm/src/memory.cairo
```diff
@@ -75,6 +76,34 @@ impl MemoryImpl of MemoryTrait {
...
+    /// * `self` - A mutable reference to the `Memory` instance to store the byte in.
+    /// * `value` - The byte value to store in memory.
+    /// * `offset` - The offset within memory to store the byte at.
+    #[inline(always)]
+    fn store_byte(ref self: Memory, value: u8, offset: usize) {
+        let new_min_bytes_len = helpers::ceil_bytes_len_to_next_32_bytes_word(offset + 1);
+        self.bytes_len = cmp::max(new_min_bytes_len, self.bytes_len);
+
+        // Get offset's memory word index and left-based offset of byte in word.
+        let (index, left_offset) = u32_safe_divmod(offset, u32_as_non_zero(16));
```
Eikix, line 93: should we keep a consistent naming between memory utils?
chunk_index, offset_in_chunk?

or then rename those?

enitrat: renamed index to chunk_index. Kept `left_offset` as is

Eikix (approved):
lgtm

PR #339, opened 2023-09-11
chore: clean codebase

enitrat:
Cleaned some stuff in the codebase
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


crates/evm/src/instructions.cairo
```diff
@@ -1,29 +0,0 @@
-/// Sub modules.
```
Eikix, line 1: We no longer need to have this "module" file?
Can we even delete it?

enitrat: No, as we declared these modules at the root of the module tree, in `lib.cairo`. This makes it easier to navigate in the codebase and only keeps important files where actual code is written.

Eikix (dismissed):
lgtm

crates/evm/src/instructions/duplication_operations.cairo
```diff
@@ -1,21 +1,12 @@
...
-    use evm::context::ExecutionContext;
-    use evm::context::ExecutionContextTrait;
-    use evm::context::BoxDynamicExecutionContextDestruct;
+    use evm::context::{ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionContextDestruct};
     use evm::stack::StackTrait;
-    use traits::Into;
-    use result::ResultTrait;
     use evm::errors::EVMError;
 
     /// Generic DUP operation
```
ClementWalter, line 12: to be inlined everywhere?

ClementWalter: use `u8` instead of `NonZero` and remove all the `into/try_into`

crates/evm/src/lib.cairo
```diff
@@ -11,7 +11,37 @@ mod stack;
 mod interpreter;
 
 // instructions module
-mod instructions;
+mod instructions {
```
ClementWalter, line 14: I'd leave the `instructions.cairo` for now, see https://doc.rust-lang.org/book/ch07-05-separating-modules-into-different-files.html#alternate-file-paths

ClementWalter (changes requested):
please leave module file for now

ClementWalter (changes requested):
please leave module file for now

ClementWalter (approved):

PR #341, opened 2023-09-11
Fix compare_snapshot CI

ClementWalter:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

The gas report CI is not allowed to post on PR page. After some research, it seems like the only way to have such a feature would be:
- long and honest: create an app (like codecov)
- short and hacky: create a personal token and use it, the user owning the token would be the posting username

See also: https://github.com/actions/first-interaction/issues/10

## What is the new behavior?

Since the detailed output is also visible in the CI logs, I've just removed the comment step to have the job pass if there is no change and fail otherwise

- [ ] Yes
- [x] No


Eikix (approved):
lgtm

Issue #335, opened 2023-09-11
feat: add a `store_padded_segment ` method to memory


enitrat:
# Feature Request

**Describe the Feature Request**
Several opcodes have the following behavior: 
- Given an array of bytes, copy `size` bytes starting from the `offset` index into memory starting from `dest_offset` 
- If `size` is bigger than the slice to copy to memory, then fill the remaining memory elements with zeroes

It's not very efficient to duplicate this behavior. Instead, we can add a method to memory:

```rust
fn store_padded_segment(ref self: Memory, offset: usize, length:usize, source:Span<u8>);
```

This function will perform the following:

- It first checks that the length is not 0 and returns early if so.
- It resizes the memory if needed to fit the data being written.
- Copy all elements from source to the memory, starting at `offset`
- Copy `length - source.len()` zeroes afterwards

You will then refactor the `xcopy` opcodes to use this newly introduced method.
**Describe Preferred Solution**

You will add the required unit tests, covering correct execution of the function and different edge cases.
**Related Code**

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

khaeljy:
Can I do it?

khaeljy:
If I've understood correctly, parameter `offset` designates the start offset in the `source` Span. So I think we're missing a `dest_offset` parameter to define the start offset in memory.

@enitrat Is that right?

khaeljy:
Or does `offset` designate the memory start offset, and we have no offset for the source start (offset = 0)?

enitrat:
Yep second one

enitrat:
@khaeljy I will get back to this tmr, indeed I think there's a confusion but
- source is the slice of the data to copy to memory (not padded, not modified)
- offset is the memory address we start the copy at
- length is the amount of data we will copy

khaeljy:
Thanks, that's clearer. Indeed, this sentence confused me a bit: `Given an array of bytes, copy size bytes starting from the offset index into memory starting from dest_offset`

PR #343, opened 2023-09-11
feat: add a store_padded_segment method to memory

khaeljy:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #335 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


crates/evm/src/memory.cairo
```diff
@@ -124,6 +127,32 @@ impl MemoryImpl of MemoryTrait {
...
+    /// * `length` - The length of bytes to store in memory.
+    /// * `source` - A span of bytes to store in memory.
+    #[inline(always)]
+    fn store_padded_segment(ref self: Memory, offset: usize, length: usize, source: Span<u8>) {
+        if length == 0 {
+            return;
+        }
+
+        // Expand memory if needed
+        self.ensure_length(offset + length);
```
khaeljy, line 145: Is this really necessary? Isn't this already handled by the `store_n` function?

crates/evm/src/memory.cairo
```diff
@@ -124,6 +127,32 @@ impl MemoryImpl of MemoryTrait {
         self.store_last_word(final_chunk, offset_in_chunk_f, mask_f, final_bytes);
     }
 
+    /// Stores a span of N bytes into memory at a specified offset with padded out of bounds with 0s.
```
enitrat, line 130:
```suggestion
    /// Stores a span of N bytes into memory at a specified offset with padded with 0s to match the size parameter.
```

crates/evm/src/memory.cairo
```diff
@@ -124,6 +127,32 @@ impl MemoryImpl of MemoryTrait {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `self` - A mutable reference to the `Memory` instance to store the bytes in.
```
enitrat, line 134:
```suggestion
    /// * `self` - The `Memory` instance to store the bytes in.
```

I don't like to use the word reference as it's not an accurate description of what the object is

crates/evm/src/memory.cairo
```diff
@@ -124,6 +127,32 @@ impl MemoryImpl of MemoryTrait {
...
+        if length == 0 {
+            return;
+        }
+
+        // Expand memory if needed
+        self.ensure_length(offset + length);
+
+        // Get elements to store
+        let elements = if (source.len() < length) {
+            source.pad_right(length - source.len())
```
enitrat, line 149: Can you rename `pad_right` to `clone_pad_right` - we want to make it explicit

enitrat: Actually, this is not a very efficient implementation of this method. This will perform an clone of the array elements, which might be expensive if the array is big.

Could you do as previously done in the opcodes, call first `store_n ` for the source elements, then call `store_n` for the added zeroes? This prevents us from performing an explicit copy.

Add a `//TODO: optimize this with a specific function`, so that we can refactor this later with something better that only calls `store_n` once

khaeljy: Yes, I had favored readability, but you're right, it's not optimal.

crates/evm/src/memory.cairo
```diff
@@ -130,6 +133,41 @@ impl MemoryImpl of MemoryTrait {
...
+        } else {
+            length
+        };
+
+        let data_to_copy: Span<u8> = source.slice(0, slice_size);
+        self.store_n(data_to_copy, offset);
+
+        // For out of bound bytes, 0s will be copied.
+        if (slice_size < length) {
+            let out_of_bounds_bytes: Span<u8> = ArrayTrait::new()
```
enitrat, line 164: Can you create a method `append_n(ref self: Array, value: u8, n: usize)` [here](https://github.com/kkrt-labs/kakarot-ssj/blob/main/crates/utils/src/helpers.cairo#L181) that appends `n` times `value` to the ref array? And call it as `array![].append_n(0, length - source.len())` (we don't need genericity for now)

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -214,14 +214,14 @@ fn test_calldatacopy_with_offset() {
 #[available_gas(20000000)]
 fn test_calldatacopy_with_out_of_bound_bytes() {
     // For out of bound bytes, 0s will be copied.
-    test_calldatacopy(32, 0, 8, array![4, 5, 6].span().pad_right(5));
+    test_calldatacopy(32, 0, 8, array![4, 5, 6].span().clone_pad_right(5));
```
enitrat, line 217: and here you will be able to call `fill` instead of cloning the span

khaeljy: I can't find the fill function, where is it?

crates/evm/src/memory.cairo
khaeljy: I couldn't make `array![].append_n(0, length - source.len())`, but `ArrayExtensionTrait::append_n(ref out_of_bounds_bytes, 0, length - source.len())`.
Did I miss something?

crates/evm/src/tests/test_instructions/test_environment_information.cairo
enitrat: sorrryyyy i wanted to say `append_n` 😓

crates/evm/src/memory.cairo
enitrat: It must be a problem with how the ArrayExtensionTrait was implemented. Generic arguments were provided at the function level but not at the implementation level.

Can you replace it with:

```rust

#[generate_trait]
impl ArrayExtension<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayExtensionTrait<T> {
    // Concatenates two arrays by adding the elements of arr2 to arr1.
    fn concat(ref self: Array<T>, mut arr2: Span<T>) {
        loop {
            match arr2.pop_front() {
                Option::Some(elem) => self.append(*elem),
                Option::None => {
                    break;
                }
            };
        }
    }

    /// Reverses an array
    fn reverse(self: Span<T>) -> Array<T> {
        let mut counter = self.len();
        let mut dst: Array<T> = ArrayTrait::new();
        loop {
            if counter == 0 {
                break ();
            }
            dst.append(*self[counter - 1]);
            counter -= 1;
        };
        dst
    }

    // Appends n time value to the Array
    fn append_n(ref self: Array<T>, value: T, mut n: usize) {
        loop {
            if n == 0 {
                break;
            }

            self.append(value);

            n -= 1;
        };
    }
}
```

After which the following will work

```rust
  let mut zeroes: Array<u8> = array![];
zeroes.append_n(0_u8, length - source.len());
self.store_n(zeroes.span(), offset + slice_size);
 ```

khaeljy: I see, thanks!

enitrat (approved):
lgtm 💯 \

PR #351, opened 2023-09-14
docs: evm-machine design doc

enitrat:
## Pull Request type

In this design document, we dive into a refactor of the Kakarot's EVM-Machine implementation, compatible with Cairo's limitations raised in https://github.com/starkware-libs/cairo/issues/4032. 
The accepted proposed design will need to: 
- Be easily refactorable
- Have a low overhead cost compared to the one originally planned
- Not introduce development complexity in other parts of the code

I believe that the proposed one performs well in all the points mentioned above. Please discuss it extensively!

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Documentation content changes

Resolves: #

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/docs/general/machine.md
+++ b/docs/general/machine.md
@@ -26,10 +26,10 @@ To overcome the problem stated above, we have come up with the following design:
   between the different execution contexts.
 - Each execution context has its own identifier `id`, which uniquely identifies
   it.
-- Each execution context has a `parent_context` field, which is the numerical
-  identifier of its optional parent execution context.
-- Each execution context has a `child_context` field, which is the numerical
-  identifier of its optional subcontext.
+- Each execution context has a `parent_context` field, whose value is either the
+  identifier of its parent execution context or `null`.
+- Each execution context has a `child_context` field, whose value is either the
+  identifier of its child context or `null`.
 - The execution context tree is a directed acyclic graph, where each execution
   context has at most one parent, and at most one child.
 - A specific execution context is accessible by traversing the execution context
--- a/docs/general/machine.md
+++ b/docs/general/machine.md
@@ -74,8 +74,7 @@ classDiagram
         destroyed_contracts: Array~EthAddress~,
         events: Array~Event~,
         create_addresses: Array~EthAddress~,
-        reverted: bool,
-        stopped:bool,
+        status: Status,
         return_data: Array~u32~,
         parent_context: Nullable~ExecutionContext~,
         child_context: Nullable~ExecutionContext~,
@@ -95,6 +94,13 @@ classDiagram
         global_keys: Array~felt252~,
     }
 
+    class Status{
+    <<enumeration>>
+      Active,
+      Stopped,
+      Reverted
+    }
+
 
     Machine *-- Memory
     Machine *-- Stack
@@ -102,6 +108,7 @@ classDiagram
     Machine *-- Journal
     ExecutionContext *-- ExecutionContext
     ExecutionContext *-- CallContext
+    ExecutionContext *-- Status
 ```
 
 ### The Stack
@@ -192,10 +199,10 @@ Journal will have the following fields:
 
 ```rust
   struct Journal {
-      local_changes: Felt252Dict<felt252>,
-      local_keys: Array<felt252>,
-      global_changes: Felt252Dict<felt252>,
-      global_keys: Array<felt252>,
+      local_changes: Felt252Dict<u256>,
+      local_keys: Array<u256>,
+      global_changes: Felt252Dict<u256>,
+      global_keys: Array<u256>,
   }
 ```
 
--- a/docs/general/machine.md
+++ b/docs/general/machine.md
@@ -48,7 +48,7 @@ classDiagram
         root_ctx: ExecutionContext,
         stack: Stack,
         memory: Memory,
-        storage_updates: StorageUpdates,
+        storage_journal: Journal,
     }
 
     class Memory{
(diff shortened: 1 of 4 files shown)


Eikix (commented):
What are the limitations of this approach?

enitrat:
> What are the limitations of this approach?

It's a bit more expensive as we need to track each stack/memory's context-related segment length in a dict instead of as a simple variable.
Also, this imposes a limit on the maximum amount of memory space a context can allocate.

Eikix:
> > What are the limitations of this approach?
> 
> 
> 
> It's a bit more expensive as we need to track each stack/memory's context-related segment length in a dict instead of as a simple variable.
> 
> Also, this imposes a limit on the maximum amount of memory space a context can allocate.

Let's discuss it in our weekly
What is the limit on memory space per context?

enitrat:
> the gas costs required to have a memory containing
> 125000 words is above the 30M gas limit. We will use this heuristic to bound the
> maximum size of the memory to 125000 256-bits words. Therefore, we will bound
> the maximum size of the memory to 125000 256-bits words.

enitrat:
@Eikix If you're okay with it I propose to merge the current design proposal, and to apply any modifications from SW's feedback in a new PR

docs/general/model/execution_context.md
```diff
@@ -26,9 +26,11 @@ classDiagram
     stopped: bool,
     +gas_limit: u64
     +gas_price:u64
-    +memory Memory
-    +stack Stack
-    +u32 program_counter
+    +memory: Memory
```
Eikix, line 29: it seems this data structure does not match the data structure below:
- machine
- executioncontext 
etc.

enitrat: Ah, yes, indeed, thanks for pointing this out, I need to rework this section as well

enitrat:
@Eikix Reworked this:
- Changes brought to the ExecutionContext docs
- The Storage section has been entirely reworked

docs/general/machine.md
```diff
@@ -0,0 +1,222 @@
...
+
+## The Kakarot Machine design
+
+To overcome the problem stated above, we have come up with the following design:
+
+- There is only one instance of the Memory and the Stack, which is shared
+  between the different execution contexts.
+- Each execution context has its own identifier `id`, which uniquely identifies
+  it.
+- Each execution context has a `parent_context` field, which is the numerical
```
Eikix, line 29: is this true? the classDiagram says Nullable<ExecutionContext>

docs/general/machine.md
```diff
@@ -0,0 +1,222 @@
...
+  identifier of its optional parent execution context.
+- Each execution context has a `child_context` field, which is the numerical
```
Eikix, line 31: same q

docs/general/machine.md
```diff
@@ -0,0 +1,222 @@
...
+
+Considering Cairo's limitations raised previously, we will use a single data
+structure to track local and global storage changes. We will use a `Journal`
+data structure, that will track two things: the changes performed in the current
+execution context, and the changes performed in the transaction as a whole. The
+Journal will have the following fields:
+
+```rust
+  struct Journal {
+      local_changes: Felt252Dict<felt252>,
```
Eikix, line 195: what if the local changes are u256?

Eikix: https://github.com/Eikix/kakarot/blob/aec81a483c3862bd71e0d55850a64f338434b4bc/src/kakarot/accounts/contract/library.cairo#L25

docs/general/machine.md
```diff
@@ -0,0 +1,222 @@
...
+
+    class CallContext{
+        caller: EthAddress,
+        value: u256,
+        bytecode: Span~u8~,
+        calldata: Span~u8~,
+    }
+
+    class Journal{
+        local_changes: Felt252Dict~felt252~,
```
Eikix, line 92: should journal expand its logic to `created_contracts` and `destroyed_contracts`?
as well as contract account nonce (if the contract account runs many create tx, nonce should be optimistically increased, until tx is finalized, and we finally write the final nonce to storage)

docs/general/machine.md
```diff
@@ -0,0 +1,222 @@
...
+        origin: EthAddress,
+        call_ctx: CallContext,
+        gas_price: u32,
+        gas_limit: u32,
+        pc: u32,
+        read_only: bool,
+        destroyed_contracts: Array~EthAddress~,
+        events: Array~Event~,
+        create_addresses: Array~EthAddress~,
+        reverted: bool,
```
Eikix, line 77: should reverted / stopped be a status enum?

docs/general/machine.md
enitrat: Yes they're actually always u256 indeed

docs/general/machine.md
enitrat: like
```rust
enum Status{
Active,
Stopped,
Reverted
}
```
?

Eikix: Yeah

enitrat:
@Eikix Changes applied

Eikix (approved):
lgtm 🔥

Eikix (changes requested):
mb just noticed this PR does not have mentions of "dirtyStorage"? 
is dirtyStorage the equivalent of `local_changes`?
should we modify the other pr #360

enitrat:
>mb just noticed this PR does not have mentions of "dirtyStorage"?
>is dirtyStorage the equivalent of local_changes?
>should we modify the other pr https://github.com/kkrt-labs/kakarot-ssj/pull/360

dirtyStorage is a name that is unique to Geth. For us, it's the combination of `local_changes` and `global_changes`.

Eikix (approved):

Issue #260, opened 2023-09-02
feat: implement 0x5B - JUMPDEST Opcode


khaeljy:
# Feature Request

| Since | Group |
| --- | --- |
| Frontier| Stack Memory Storage and Flow Operations |

## Notes

Mark a valid destination for [JUMP](https://www.evm.codes/#56) or [JUMPI](https://www.evm.codes/#57). This operation has no effect on machine state during execution.
```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/kakarot-ssj/issues/261
- [ ] After #261 is implemented, go and implement the business logic of JUMPDEST
```

lambda-0x:
will pick up this after #250 as they are closely related.

Eikix:
Important update: we are gathering some bugs in the [Kakarot v0 codebase](https://github.com/kkrt-labs/kakarot), we need to make sure each issue and each PR in Kakarot-ssj is aware of the lists of known bugs. Look at this link everytime you take an issue and check your issue isn't targeted by a known bug.

Eikix:
Important update: we are gathering some bugs in the [Kakarot v0 codebase](https://github.com/kkrt-labs/kakarot), we need to make sure each issue and each PR in Kakarot-ssj is aware of the lists of known bugs. Look at [this tracking issue](https://github.com/kkrt-labs/kakarot-ssj/issues/324) everytime you take an issue and check your issue isn't targeted by a known bug. Will add this reminder in many places to make sure we keep track of known bugs.

lambda-0x:
@Eikix jumpdesk is a just a marker and it doesn't affect the execution state in anyway (essentially a noop). So do we need to do anything here like adding comment on `exec_jumpdest` or should this be closed?

Eikix:
> @Eikix jumpdesk is a just a marker and it doesn't affect the execution state in anyway (essentially a noop). So do we need to do anything here like adding comment on `exec_jumpdest` or should this be closed?

can you check what execution specs does?

lambda-0x:
yes i checked, it just increments pc and charge gas:

- [execution spec](https://github.com/ethereum/execution-specs/blob/e69b38986263fb1cdaf1331f05b7eacc6c9cd8ff/src/ethereum/shanghai/vm/instructions/control_flow.py#L149-L171)
[...]

PR #354, opened 2023-09-15
feat: add comment for JUMPDEST opcode

lambda-0x:
## Pull Request type

- [x] Feature

Resolves: #260 

## What is the new behavior?

- added for clarity that its supposed to be an NO-OP

- [x] No

### DIFF
--- a/crates/evm/src/instructions/memory_operations.cairo
+++ b/crates/evm/src/instructions/memory_operations.cairo
@@ -65,7 +65,7 @@ impl MemoryOperation of MemoryOperationTrait {
     /// # Specification: https://www.evm.codes/#5b?fork=shanghai
     /// 
     /// This doesn't have any affect on execution state, so we don't have
-    /// to do anything here. Its a NO-OP
+    /// to do anything here. It's a NO-OP.
     fn exec_jumpdest(ref self: ExecutionContext) -> Result<(), EVMError> {
         Result::Ok(())
     }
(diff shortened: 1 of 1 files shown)


crates/evm/src/instructions/memory_operations.cairo
```diff
@@ -63,6 +63,9 @@ impl MemoryOperation of MemoryOperationTrait {
     /// 0x5b - JUMPDEST operation
     /// Serves as a check that JUMP or JUMPI was executed correctly.
     /// # Specification: https://www.evm.codes/#5b?fork=shanghai
+    /// 
+    /// This doesn't have any affect on execution state, so we don't have
```
Eikix, line 67: Can you set in your IDE spell-check? Or maybe it didn't catch `its` vs. `it's` because its is also a correct word

lambda-0x: what spell-check do you use? i don't have it setup in my IDE. also i don't know whats difference between those lol. should it be `it's`?

Eikix: yeah should be `it's` because "it is a no op" => it's a no op

Eikix: i think @ClementWalter uses cspell extension

Eikix: we could add it in the CI

lambda-0x: fixed

Eikix (approved):

PR #355, opened 2023-09-15
feat: implement JUMPI tests

lambda-0x:
## Pull Request type

- [x] Feature

Resolves: #253 

## What is the new behavior?

- didn't find test for this in kakarot0, evm.codes, geth
- execution spec has something to test jumpi [(but it tests something very different)](https://github.com/ethereum/execution-specs/blob/master/tests/constantinople/vm/test_control_flow_operations.py) 
- out of bounds and those tests are not required because we would be using implementation of `JUMP` opcode to implement `JUMPI` which tests those cases.

- [x] No

### DIFF
--- a/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
+++ b/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
@@ -434,7 +434,7 @@ fn test_exec_jumpi_valid_zero() {
 
     // Then
     let pc = ctx.program_counter;
-    // ideally we should assert that it incremented, but incrementing is done by `decode_and_execut`
+    // ideally we should assert that it incremented, but incrementing is done by `decode_and_execute`
     // so we can assume that will be done
     assert(pc == old_pc, 'PC should be same');
 }
--- a/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
+++ b/crates/evm/src/tests/test_instructions/test_memory_operations.cairo
@@ -373,3 +373,137 @@ fn test_exec_jump_inside_pushn() {
     assert(result.is_err(), 'invalid jump dest');
     assert(result.unwrap_err() == EVMError::JumpError(INVALID_DESTINATION), 'invalid jump dest');
 }
+
+#[test]
+#[available_gas(20000000)]
+#[should_panic(expected: ('JUMPI not implemented yet',))]
+fn test_exec_jumpi_valid_non_zero_1() {
+    // Given
+    let bytecode: Span<u8> = array![0x01, 0x02, 0x03, 0x5B, 0x04, 0x05].span();
+    let mut ctx = setup_execution_context_with_bytecode(bytecode);
+    let b = 0x1;
+    ctx.stack.push(b);
+    let counter = 0x03;
+    ctx.stack.push(counter);
+    let old_pc = ctx.program_counter;
+
+    // When
+    ctx.exec_jumpi();
+
+    // Then
+    let pc = ctx.program_counter;
+    assert(pc == 0x03, 'PC should be JUMPDEST');
+}
+
+#[test]
+#[available_gas(20000000)]
+#[should_panic(expected: ('JUMPI not implemented yet',))]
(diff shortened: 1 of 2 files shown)


crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -373,3 +373,133 @@ fn test_exec_jump_inside_pushn() {
     assert(result.is_err(), 'invalid jump dest');
     assert(result.unwrap_err() == EVMError::JumpError(INVALID_DESTINATION), 'invalid jump dest');
 }
+
+#[test]
+#[available_gas(20000000)]
+#[should_panic(expected: ('JUMPI not implemented yet',))]
+fn test_exec_jumpi_valid_zero() {
```
Eikix, line 380: i think b must not be 0, for the jump to happen
if b==0 don't jump
if b!=0 jump

Eikix (commented):
a question

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
lambda-0x: oh i misunderstood it, will update.

lambda-0x: fixed

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -373,3 +373,137 @@ fn test_exec_jump_inside_pushn() {
...
+    let counter = 0x03;
+    ctx.stack.push(counter);
+    let old_pc = ctx.program_counter;
+
+    // When
+    ctx.exec_jumpi();
+
+    // Then
+    let pc = ctx.program_counter;
+    assert(pc == 0x03, 'PC should be JUMPDEST');
```
Eikix, line 416:
```suggestion
    assert(pc == 0x03, 'PC should be 0x03');
```

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -373,3 +373,137 @@ fn test_exec_jump_inside_pushn() {
...
+}
+
+#[test]
+#[available_gas(20000000)]
+#[should_panic(expected: ('JUMPI not implemented yet',))]
+fn test_exec_jumpi_valid_non_zero_2() {
```
Eikix, line 401: i think this test does not give any more information than the first one
should be able to delete it

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
```diff
@@ -373,3 +373,137 @@ fn test_exec_jump_inside_pushn() {
...
+    let counter = 0x03;
+    ctx.stack.push(counter);
+    let old_pc = ctx.program_counter;
+
+    // When
+    ctx.exec_jumpi();
+
+    // Then
+    let pc = ctx.program_counter;
+    // ideally we should assert that it incremented, but incrementing is done by `decode_and_execut`
```
Eikix, line 437:
```suggestion
    // ideally we should assert that it incremented, but incrementing is done by `decode_and_execute`
```

crates/evm/src/tests/test_instructions/test_memory_operations.cairo
lambda-0x: i added this second one to make sure `JUMPI` makes the jump on any (0x69) non-zero value and not just `0x1`.

Eikix: can't we parametrise the test then?

lambda-0x: i don't think thats possible to do, is it?

and if we make this a function call and call it twice in same function, it would make it much harder to debug.

also i think we already prefer duplication over complexity to debug in the test suite.

Eikix (approved):
lgtm

PR #363, opened 2023-09-22
Final name update

TAdev0:
@Eikix  @enitrat 
 
 it's not possible to delete a contributor with All contributors, and then add it again.
 I dit it manually then, i made a PR yesterday that was accepted but i only modified the all-contributorssrc file . I actually had to also modify the README.md file , which i did in this PR. After that, all should be ok !*
 
 sorry for the loss of time for such unimportant issue.


enitrat (approved):
no worries

Issue #358, opened 2023-09-18
dev: refactor constants to be base 2 instead of base 256


lambda-0x:
see: https://github.com/kkrt-labs/kakarot-ssj/pull/329#discussion_r1328656094

This constants needs to be updated to base 2 equivalent, uses of this variables should also be updated.

https://github.com/kkrt-labs/kakarot-ssj/blob/83c2ff4c296d9b2386a5bb595ebb78ac7198ad38/crates/utils/src/constants.cairo#L31-L46

enitrat:
I don't understand, what's "base-2 equivalent"? like POW_256_2_U128 should be POW_2_16 instead?

lambda-0x:
> I don't understand, what's "base-2 equivalent"? like POW_256_2_U128 should be POW_2_16 instead?

yes

PR #365, opened 2023-09-22
dev: refactor constants to be base 2 instead of base 256

lambda-0x:
## Pull Request type

Resolves: #358 

## What is the new behavior?

-
-
-

- [x] No


crates/utils/src/constants.cairo
```diff
@@ -28,22 +28,20 @@ const POW_256_14_U256: u256 = 0x10000000000000000000000000000;
 const POW_256_15_U256: u256 = 0x1000000000000000000000000000000;
 const POW_256_16_U256: u256 = 0x100000000000000000000000000000000;
```
lambda-0x, line 29: haven't changed this because this are only used in 1 place, which is returning power's of 256. so for readability this is better

enitrat (dismissed):

crates/utils/src/constants.cairo
enitrat: But then everything is duplicated? I thought the purpose was to unify it and only use powers of two everywhere

enitrat: I mean I'm not against it because I agree that a function returning different exponents of 256 is a lot clearer with 256-based consts, but I don't like the ideas of having multiple constants for the same thing

enitrat: Ok I just realised the type is different - but I'd rather use e.g. all u128s, and cast them into a u256 with `POW_XXX_U128.into()`. Or even ditch the `U128` suffix and just `POW_XXX.into()`, for everything that fits in a u128

lambda-0x: > but I'd rather use e.g. all u128s, and cast them into a u256 with POW_XXX_U128.into()

problem with that is `POW_256_16_U256` doesn't fit into `u128` so we would have to make this one variable `u256` and all other `u128`, which would turn into use calling into in most branches of [this function](https://github.com/kkrt-labs/kakarot-ssj/blob/ef8180d9bbc59309500ca5b16687dff5e21319e8/crates/utils/src/helpers.cairo#L30-L70) except one which is u256.

lambda-0x: also afaik `into` is not free and will consume some gas for transformation

lambda-0x: > but I don't like the ideas of having multiple constants for the same thing

i don't think its the same thing, those variables are only used at one place and thats return value of above mentioned function

lambda-0x:
@enitrat if the changes look good i will fix the conflicts, LMK.

enitrat:
@lambda-0x for constant that fit in a u128 i'd rather have just one value that we `.into()` a u256 if required. Can you rebase on main, run this change, and push it so that we can see whether it adds a lot of overhead gas? (it normally shouldn't)

lambda-0x:
@enitrat do you mean all `POW_256_XX` be changed to u128 expect for `POW_256_16` (because this cannot be stored in u128 and call into for u128 types.

or you are asking me to use base 2 variable and remove base 256 completely. I have explain above why i kept them. https://github.com/kkrt-labs/kakarot-ssj/pull/365#discussion_r1335408328

enitrat:
> do you mean all POW_256_XX be changed to u128 expect for POW_256_16 (because this cannot be stored in u128 and call into for u128 types.



this one

lambda-0x:
i see a new `pow2` function after rebase, do we want to use variables here as well (there are 127 branches)?

lambda-0x:
removed type suffix from other variables as they where not needed now.

enitrat (approved):
LFG ! 🔥

PR #372, opened 2023-09-26
docs: bytecode storage

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
+# Bytecode Storage Methods for Kakarot on Starknet
+
+The bytecode is the compiled version of your contract, and it is what the
+Kakarot EVM will execute when you call the contract. As Kakarot is developed on
+top of Starknet, you cannot really "deploy" an EVM contract on Kakarot: what
```
Eikix, line 5: i think we would benefit from rephrasing this
in a way, 
- running a constructor bytecode
- storing on-chain the deployed bytecode of a contract and making it accessible at a particular address 

is the essence of deployment

what we're not enabling are deploy txs in some sense?

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
+# Bytecode Storage Methods for Kakarot on Starknet
+
+The bytecode is the compiled version of your contract, and it is what the
+Kakarot EVM will execute when you call the contract. As Kakarot is developed on
```
Eikix, line 4: We would also benefit from being more precise, such as that Kakarot's state is embedded into Starknet's state

Eikix: That way we're not tied to saying we're deployed on Starknet L2 (and saying we're developed on top of Starknet implies that imo

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+top of Starknet, you cannot really "deploy" an EVM contract on Kakarot: what
+actually happens is that the EVM bytecode of your contract is stored on the
+blockchain, and the Kakarot EVM will be able to load it when you want to execute
+it.
```
Eikix, line 8: Lgtm, I think overall we can rephrase the content to not contain "you" and "your" but speak passively?
Maybe claude can help, or gpt

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+to make the Starknet state available, even in the case where the operator
+suddenly ceases operations. Data availability ensures that users can always
+reconstruct the state of the rollup by deriving its current state from the data
+posted by the rollup operator.
+
+Without this, users would not be able to query an L2 contract's state in case
+the operator becomes unavailable. It provides users the security of knowing that
+if the Starknet sequencer ever stops functioning, they can prove custody of
+their funds using the data posted on the DA Layer. If that DA Layer is Ethereum
+itself, then they inherit from Ethereum's security guarantees.
```
Eikix, line 43: nice

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+reconstruct the current state of Starknet. This security comes with a
+significant price, as the publication of state diffs on Ethereum accounted for
+[over 93% of the transaction fees paid on Starknet](https://community.starknet.io/t/volition-hybrid-data-availability-solution/97387).
+
+The first choice when it comes to storing contract bytecode is to store it as a
+regular storage variable, whose state diff is posted on Ethereum acting as the
+DA Layer. As outlined in our [Contract Storage](./contract_storage.md) design,
+deploying a new contract on Kakarot would not result in the deployment of a
+contract on Starknet, but rather in the storage of the contract bytecode in a
+storage variable of the KakarotCore contract.
```
Eikix, line 61: very clear, nice

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+small transactions.
+
+In our case, we would store the contract bytecode in a storage variable that is
+settled on the L2DA instead of the L1DA. This would make contract deployment
+extremely cheap on Kakarot, as we will save the cost of posting the state diff
+associated to the update of our stored bytecode on Ethereum.
+
+#### Associated Risks
+
+There are some risks that must be considered when using Volition. Consider the
```
Eikix, line 121: could someone from starkware volition team read this paragraph and give their opinion?
i still dont understand why when Starknet gets a consensus, Sequencers can sign that they had the required L2DA data to perform the state change 

which means every single sequencer in the consensus would have to go rogue and refuse to share the updated L2DA 

ofc its different for the centralized sequencer case

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+attack is finished, the honest sequencers won't have the data needed to
+reconstruct and compute the new root of the L2DA. In a such situation, not only
+the L2DA is not accessible anymore, but any execution relying on L2DA will not
+be executable and provable anymore, as sequencers won't have access the the L2DA
+state.
+
+Even though this event is unlikely to happen, it remains a possibility that must
+be taken into account as L2DA is less secure than L1DA. If an event like this
+were ever to happen, then the stored bytecode would be lost, and the deployed
+contract would not be executable anymore.
```
Eikix, line 133: people could give the bytecode themself, as they are likely to "have" it locally forever

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+code. This has the advantage of also being cheap, as this data is not stored on
+L1.
+
+On Starknet, there is a distinction between classes which is the definition of a
+contract containing the Cairo bytecode, and contracts which are instances of
+classes. When you declare a contract on Starknet, its information is added to
+the
+[Classes Tree](https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/starknet-state/#classes_tree),
+which encodes information about the existing classes in the state of Starknet by
+mapping class hashes to their. This class tree is itself a part of the Starknet
```
Eikix, line 151: mapping class hashes to their?
to their class?

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+State Commitment, which is verified on Ethereum during state updates. The class
+itself is stored in the nodes (both sequencers and full nodes) of Starknet.
+
+To implement this solution, we must declare a new class each time a contract is
+deployed via Kakarot. This new class would contain the EVM bytecode of the
+contract, exposed inside a view function that would return the entire bytecode
+when queried. To achieve that, we would need to have the RPC craft a custom
+Starknet contract that would contain this EVM bytecode, and declare it on
+Starknet - which is not ideal from security perspectives.
```
Eikix, line 160: right after this paragraph, it's not so clear that we're going with option 1

except the first sentence of the next paragraph

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+
+## Implementation details
+
+Kakarot uses the first solution, storing the bytecode inside a storage variable
+that is committed to Ethereum. This solution is the most secure one, as it
+relies on Ethereum as a DA Layer, and thus inherits from Ethereum's security
+guarantees, ensuring that the bytecode of the deployed contract is always
+available.
+
+A `deploy` transaction is identified by the `to` address being set to zero. The
```
Eikix, line 170: is this true? maybe it's if the address is not even set, or like `to: ""`
i'd have to check again 
but i think sending a tx to address(0) is valid

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+data sent to the KakarotCore contract when deploying a new contract wil be
```
Eikix, line 171:
```suggestion
data sent to the KakarotCore contract when deploying a new contract will be
```

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+formatted by the RPC to pack the bytecode into 31-bytes values, and passed as a
+`Array<felt252>` to the entrypoint `eth_call` of the KakarotCore contract. This
```
Eikix, line 173:
```suggestion
`Array<felt252>` to the entrypoint `eth_send_transaction` of the KakarotCore contract. This
```

Eikix: eth_call can only be called for view purposes atm

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+allows us to save on computation costs required to pack all byte values into the
+31-bytes values that we will store in the contract storage.
+
+The contract storage related to the deployed contract will be organized as such:
+
+```rust
+struct Storage {
+    bytecode: LegacyMap<EthAddress, List<bytes31>>,
```
Eikix, line 181: do we also need valid jumps destinations here?

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+    pending_word: LegacyMap<EthAddress, felt252>,
+    pending_word_len: LegacyMap<EthAddress, usize>,
+}
+```
+
+Each deployed contract has it's own EVM address, that we use as a key to a
+`LegacyMap` type when computing the address of each storage variable. We use the
+`List` type from Alexandria to store the bytecode, as it allows us to store up
```
Eikix, line 189:
```suggestion
`List` type from [Alexandria](https://github.com/keep-starknet-strange/alexandria/blob/main/src/storage/src/list.cairo) to store the bytecode, as it allows us to store up
```

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+struct ByteArray {
+    // Full "words" of 31 bytes each. The first byte of each word in the byte array
+    // is the most significant byte in the word.
+    data: Array<bytes31>,
+    // This felt252 actually represents a bytes31, with < 31 bytes.
+    // It is represented as a felt252 to improve performance of building the byte array.
+    // The number of bytes in here is specified in `pending_word_len`.
+    // The first byte is the most significant byte among the `pending_word_len` bytes in the word.
+    pending_word: felt252,
+    // Should be in range [0, 30].
```
Eikix, line 211: should it be a u8?

Eikix (commented):
These documents are great!

docs/general/contract_bytecode.md
enitrat: doing this after first wave of changes

docs/general/contract_bytecode.md
enitrat: > Starknet gets a consensus, Sequencers can sign that they had the required L2DA data to perform the state change

which means every single sequencer in the consensus would have to go rogue and refuse to share the updated L2DA

Not every: a malicious majority is sufficient.

docs/general/contract_bytecode.md
enitrat: The problem is not that the bytcode can't be accessed - it's that no operation using L2DA will succeed due to a failure of proving the current L2DA state.

The only migtigation would be to completely erase the L2DA state.

docs/general/contract_bytecode.md
enitrat: compiled class hash

docs/general/contract_bytecode.md
enitrat: Added this information at the end of Solution 1

docs/general/contract_bytecode.md
enitrat: fixed to: `being empty`

docs/general/contract_bytecode.md
Eikix: > > Starknet gets a consensus, Sequencers can sign that they had the required L2DA data to perform the state change
> 
> which means every single sequencer in the consensus would have to go rogue and refuse to share the updated L2DA
> 
> Not every: a malicious majority is sufficient.

Why?

If just one sequencer shares its L2DA with every full node, the state of the chain is "safe"?

docs/general/contract_bytecode.md
enitrat: I think so. I'm not sure how yet. Will be for another PR imo

docs/general/contract_bytecode.md
enitrat: This is how it is in the corelib

docs/general/contract_bytecode.md
```diff
@@ -0,0 +1,229 @@
...
+
+```rust
+struct ByteArray {
+    // Full "words" of 31 bytes each. The first byte of each word in the byte array
+    // is the most significant byte in the word.
+    data: Array<bytes31>,
+    // This felt252 actually represents a bytes31, with < 31 bytes.
+    // It is represented as a felt252 to improve performance of building the byte array.
+    // The number of bytes in here is specified in `pending_word_len`.
+    // The first byte is the most significant byte among the `pending_word_len` bytes in the word.
```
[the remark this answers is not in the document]
enitrat: This is a direct copypasta of the corelib code. I think what it means is that for example in the value 0xABCDEF, the first byte is the MSB, so 0xAB

docs/general/contract_bytecode.md
enitrat: > one sequencer shares its L2DA with every full node, the state of the chain is "safe"?

The sequencer is in minority. After the hostile takeover, the chain has agreed on a new L2DA root. If an information is hidden from honest sequencers, they can't compute the new state as they're missing information.

Eikix: But this new L2DA root cannot be proven, correct? Because it cannot result in a valid proof

docs/general/contract_bytecode.md
```diff
@@ -1,11 +1,13 @@
...
 The bytecode is the compiled version of your contract, and it is what the
-Kakarot EVM will execute when you call the contract. As Kakarot is developed on
-top of Starknet, you cannot really "deploy" an EVM contract on Kakarot: what
-actually happens is that the EVM bytecode of your contract is stored on the
-blockchain, and the Kakarot EVM will be able to load it when you want to execute
-it.
+Kakarot EVM will execute when you call the contract. As Kakarot's state is
+embedded into the Starknet chain it is deployed on, you do not really "deploy"
+an EVM contract on Kakarot: what actually happens is that the EVM bytecode of
+the deployed contract if first ran, and the returned data is then stored
```
Eikix, line 7: if -> is 
typo

docs/general/contract_bytecode.md
```diff
@@ -1,11 +1,13 @@
...
+on-chain at a particular storage address in the KakarotCore contract storage.
+The Kakarot EVM will be able to load this bytecode by querying it's own storage
```
Eikix, line 9: it's -> its

docs/general/contract_bytecode.md
```diff
@@ -1,11 +1,13 @@
...
-Kakarot EVM will execute when you call the contract. As Kakarot is developed on
-top of Starknet, you cannot really "deploy" an EVM contract on Kakarot: what
-actually happens is that the EVM bytecode of your contract is stored on the
-blockchain, and the Kakarot EVM will be able to load it when you want to execute
-it.
+Kakarot EVM will execute when you call the contract. As Kakarot's state is
+embedded into the Starknet chain it is deployed on, you do not really "deploy"
+an EVM contract on Kakarot: what actually happens is that the EVM bytecode of
+the deployed contract if first ran, and the returned data is then stored
+on-chain at a particular storage address in the KakarotCore contract storage.
```
Eikix, line 8: i think this flow is well explained, but might benefit from a mermaid diagram for clarity

docs/general/contract_bytecode.md
enitrat: The thing is that it actually can be proven (given that it's valid). You know that the state transition from S_A to S_B is valid. But if you don't make S_B available to validators, then how can they validate that the S_B -> S_C transition is valid? They can't, hence txs executing L2DA updates will fail. (As I understood it)

enitrat:
@Eikix last check and merge?

Eikix (approved):
lgtm

Issue #374, opened 2023-09-27
dev: replace all `*_div_rem` by `DivRem::div_rem`


enitrat:
This improvement makes the codebase lighter by avoiding explicit imports of libfuncs and rely on the trait system to use the correct type

khaeljy:
Can I do it, pls?

khaeljy:
Cause of the `DivRem` trait signature which requires lhs and rhs to be of the same `T` type, I was unable to replace `u512_safe_div_rem_by_u256` which is used here: `crates/evm/src/instructions/stop_and_arithmetic_operations.cairo`

The same applies to the `i256_signed_div_rem` function in `crates/utils/src/i256.cairo`, which takes a `NonZero<u256>` as its second argument, because it's not possible to have a `NonZero<i256>` since `NonZero` is a type defined in the compiler, not in Cairo ( thanks @enitrat ).

PR #381, opened 2023-09-27
Dev: replace all `*_div_rem` by `DivRem::div_rem`

khaeljy:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build-related changes

Resolves: #374 

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/memory.cairo
+++ b/crates/evm/src/memory.cairo
@@ -1,6 +1,5 @@
 use integer::{
-    u32_safe_divmod, u32_as_non_zero, u128_safe_divmod, u128_as_non_zero, u256_safe_div_rem,
-    u256_as_non_zero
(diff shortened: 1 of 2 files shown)


khaeljy:
Cause of the `DivRem` trait signature which requires lhs and rhs to be of the same `T` type, I was unable to replace `u512_safe_div_rem_by_u256` which is used here: `crates/evm/src/instructions/stop_and_arithmetic_operations.cairo`

The same applies to the `i256_signed_div_rem` function in `crates/utils/src/i256.cairo`, which takes a `NonZero<u256>` as its second argument, because it's not possible to have a `NonZero<i256>` since `NonZero` is a type defined in the compiler, not in Cairo ( thanks @enitrat ).

enitrat (approved):
lgtm

Issue #377, opened 2023-09-27
feat: update the execution context and machine to match new design 


Eikix:
## Feature Request

@enitrat created a new design for execution context and machine, in docs/general/execution_context.md and docs/general/machine.md

```mermaid
classDiagram
    class ExecutionContext{
        ctx_id: usize,
        origin: EthAddress,
        call_ctx: CallContext,
        gas_price: u32,
        gas_limit: u32,
        pc: u32,
        read_only: bool,
        destroyed_contracts: Array~EthAddress~,
        events: Array~Event~,
        create_addresses: Array~EthAddress~,
        status: Status,
        return_data: Array~u32~,
        parent_context: Nullable~ExecutionContext~,
        child_context: Nullable~ExecutionContext~,
    }

    class CallContext{
        bytecode: Span~u8~,
        calldata: Span~u8~,
        value: u256,
    }

    class Event{
        keys: Array~u256~,
        data: Array~u8~
    }

    class Status{
    <<enumeration>>
      Active,
      Stopped,
      Reverted
    }

    ExecutionContext *-- CallContext
    ExecutionContext *-- Event
    ExecutionContext *-- Status
```

```mermaid
classDiagram
    class Machine{
        current_ctx: usize,
        ctx_count: usize,
        root_ctx: ExecutionContext,
        stack: Stack,
        memory: Memory,
        storage_journal: Journal,
    }

    class Memory{
        active_segment: usize,
        items: Felt252Dict~u128~,
        bytes_len: Felt252Dict~usize~,
    }

    class Stack{
        +active_segment: usize,
        +items: Felt252Dict~Nullable~u256~~,
        +len: Felt252Dict~usize~
    }

    class ExecutionContext{
        ctx_id: usize,
        origin: EthAddress,
        call_ctx: CallContext,
        gas_price: u32,
        gas_limit: u32,
        pc: u32,
        read_only: bool,
        destroyed_contracts: Array~EthAddress~,
        events: Array~Event~,
        create_addresses: Array~EthAddress~,
        status: Status,
        return_data: Array~u32~,
        parent_context: Nullable~ExecutionContext~,
        child_context: Nullable~ExecutionContext~,
    }

    class CallContext{
        caller: EthAddress,
        value: u256,
        bytecode: Span~u8~,
        calldata: Span~u8~,
    }

    class Journal{
        local_changes: Felt252Dict~felt252~,
        local_keys: Array~felt252~,
[...]

PR #384, opened 2023-09-28
Feat/execution context update

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #377 

## What is the new behavior?

-
-
-

- [x] Yes
- [ ] No

### DIFF
--- a/crates/evm/src/context.cairo
+++ b/crates/evm/src/context.cairo
@@ -136,6 +136,7 @@ struct ExecutionContext {
     create_addresses: Array<EthAddress>,
     return_data: Array<u8>,
     parent_context: Nullable<ExecutionContext>,
+    child_context: Nullable<ExecutionContext>,
 }
 
 impl DefaultBoxExecutionContext of Default<Box<ExecutionContext>> {
@@ -157,6 +158,7 @@ impl ExecutionContextImpl of ExecutionContextTrait {
         starknet_address: ContractAddress,
         call_context: CallContext,
         parent_context: Nullable<ExecutionContext>,
+        child_context: Nullable<ExecutionContext>,
         return_data: Array<u8>,
     ) -> ExecutionContext {
         ExecutionContext {
@@ -181,6 +183,7 @@ impl ExecutionContextImpl of ExecutionContextTrait {
             create_addresses: Default::default(),
             return_data,
             parent_context,
+            child_context,
         }
     }
 
--- a/crates/evm/src/context.cairo
+++ b/crates/evm/src/context.cairo
@@ -245,6 +245,7 @@ impl ExecutionContextImpl of ExecutionContextTrait {
     #[inline(always)]
     fn revert(ref self: ExecutionContext, revert_reason: Span<u8>) {
         self.status = Status::Reverted;
+        ArrayExtensionTrait::concat(ref self.return_data, revert_reason);
     }
 
     // *************************************************************************
@@ -261,25 +262,6 @@ impl ExecutionContextImpl of ExecutionContextTrait {
         *self.starknet_address
     }
 
-    #[inline(always)]
-    fn caller(self: @ExecutionContext) -> EthAddress {
-        (*self.call_context).unbox().caller()
-    }
-
-    #[inline(always)]
-    fn read_only(self: @ExecutionContext) -> bool {
-        (*self.call_context).unbox().read_only()
-    }
-
-    #[inline(always)]
-    fn gas_limit(self: @ExecutionContext) -> u64 {
-        (*self.call_context).unbox().gas_limit()
-    }
-
-    #[inline(always)]
-    fn gas_price(self: @ExecutionContext) -> u64 {
-        (*self.call_context).unbox().gas_price()
-    }
 
     // *************************************************************************
     //                          ExecutionContext methods
--- a/crates/evm/src/context.cairo
+++ b/crates/evm/src/context.cairo
@@ -15,7 +15,6 @@ enum Status {
     Reverted
 }
 
-
 // *************************************************************************
 //                              CallContext
 // *************************************************************************
@@ -117,37 +116,6 @@ impl DefaultBoxCallContext of Default<Box<CallContext>> {
     }
 }
 
-// *************************************************************************
-//                              DynamicContext
-// *************************************************************************
-
-#[derive(Drop, Default)]
-struct DynamicContext {
-    destroyed_contracts: Array<EthAddress>,
-    events: Array<Event>,
-    create_addresses: Array<EthAddress>,
-    return_data: Array<u8>,
-}
-
-impl DefaultBoxDynamicContext of Default<Box<DynamicContext>> {
-    fn default() -> Box<DynamicContext> {
-        let dynamic_context: DynamicContext = Default::default();
-        BoxTrait::new(dynamic_context)
-    }
-}
-
-#[generate_trait]
-impl DynamicContextImpl of DynamicContextTrait {
-    #[inline(always)]
-    fn new(return_data: Array<u8>) -> DynamicContext {
-        DynamicContext {
-            destroyed_contracts: Default::default(),
-            events: Default::default(),
-            create_addresses: Default::default(),
-            return_data,
-        }
-    }
-}
 
 // *************************************************************************
 //                              ExecutionContext
@@ -163,9 +131,18 @@ struct ExecutionContext {
     program_counter: u32,
     status: Status,
     call_context: Box<CallContext>,
-    dynamic_context: Box<DynamicContext>,
+    destroyed_contracts: Array<EthAddress>,
+    events: Array<Event>,
+    create_addresses: Array<EthAddress>,
+    return_data: Array<u8>,
     parent_context: Nullable<ExecutionContext>,
-    child_context: Nullable<ExecutionContext>,
+}
+
+impl DefaultBoxExecutionContext of Default<Box<ExecutionContext>> {
+    fn default() -> Box<ExecutionContext> {
+        let context: ExecutionContext = Default::default();
+        BoxTrait::new(context)
+    }
 }
 
 
@@ -199,9 +176,11 @@ impl ExecutionContextImpl of ExecutionContextTrait {
                     call_context.gas_price,
                 )
             ),
-            dynamic_context: BoxTrait::new(DynamicContextTrait::new(return_data)),
+            destroyed_contracts: Default::default(),
+            events: Default::default(),
+            create_addresses: Default::default(),
+            return_data,
             parent_context,
-            child_context: Default::default(),
         }
     }
 
@@ -210,7 +189,7 @@ impl ExecutionContextImpl of ExecutionContextTrait {
     // *************************************************************************
 
     #[inline(always)]
-    fn reverted(ref self: ExecutionContext) -> bool {
+    fn reverted(self: @ExecutionContext) -> bool {
         match self.status {
             Status::Active => false,
             Status::Stopped => false,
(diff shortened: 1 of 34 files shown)


crates/evm/src/context.cairo
```diff
@@ -7,38 +7,77 @@ use utils::traits::{SpanDefault, EthAddressDefault, ContractAddressDefault};
 use starknet::{EthAddress, ContractAddress};
 use starknet::get_caller_address;
 
+#[derive(Drop, Default, Copy)]
```
Eikix, line 10: adding a status for the execution context,

one question comes to mind: when you revert an execution context,
is the "stopped" var true as well?

crates/evm/src/context.cairo
```diff
@@ -7,38 +7,77 @@ use utils::traits::{SpanDefault, EthAddressDefault, ContractAddressDefault};
...
+//                              CallContext
+// *************************************************************************
+// We should not directly access the fields of the call context;
+// instead we should use the methods defined in the trait.
+// This is not enforced until there are `pub` and `priv` visibility on struct fields.
+
 /// The call context.
 #[derive(Drop, Copy, Default)]
 struct CallContext {
+    caller: EthAddress,
```
Eikix, line 28: adding the "from" field here

crates/evm/src/context.cairo
```diff
@@ -7,38 +7,77 @@ use utils::traits::{SpanDefault, EthAddressDefault, ContractAddressDefault};
...
     /// The bytecode to execute.
     bytecode: Span<u8>,
     /// The call data.
     calldata: Span<u8>,
     /// Amount of native token to transfer.
     value: u256,
+    // If the call is read only (cannot modify the state of the chain)
+    read_only: bool,
```
Eikix, line 36: adding call related fields here as well -> deleting static execution context

crates/evm/src/context.cairo
```diff
@@ -53,188 +92,152 @@ impl CallContextImpl of CallContextTrait {
     fn value(self: @CallContext) -> u256 {
         *self.value
     }
-}
-
-// *************************************************************************
-//                              StaticExecutionContext
-// *************************************************************************
-
-#[derive(Drop, Copy, Default)]
```
Eikix, line 62: deleting static context as it's not really relevant anymore, too heavy to manipulate

crates/evm/src/context.cairo
```diff
@@ -53,188 +92,152 @@ impl CallContextImpl of CallContextTrait {
...
-        DynamicExecutionContext {
-            destroyed_contracts: Default::default(),
-            events: Default::default(),
-            create_addresses: Default::default(),
-            revert_contract_state: Default::default(),
-            return_data,
-            reverted: false,
-            stopped: false
-        }
+impl DefaultBoxCallContext of Default<Box<CallContext>> {
```
Eikix, line 112: we need a default on the box itself to derive default on the execution context

crates/evm/src/context.cairo
```diff
@@ -53,188 +92,152 @@ impl CallContextImpl of CallContextTrait {
...
 
+
 // *************************************************************************
 //                              ExecutionContext
 // *************************************************************************
 
 /// The execution context.
 /// Stores all data relevant to the current execution context.
-#[derive(Destruct)]
+#[derive(Drop, Default)]
```
Eikix, line 126: we can use Drop now, as there is no Dict

crates/evm/src/context.cairo
```diff
@@ -53,188 +92,152 @@ impl CallContextImpl of CallContextTrait {
...
 struct ExecutionContext {
-    static_context: Box<StaticExecutionContext>,
-    dynamic_context: Box<DynamicExecutionContext>,
+    context_id: usize,
+    evm_address: EthAddress,
```
Eikix, line 129: evm_address is the current called contract's address
starknet_address is its sn address, to prevent from using compute_starknet_address everytime (or to check the mapping)

crates/evm/src/context.cairo
```diff
@@ -53,188 +92,152 @@ impl CallContextImpl of CallContextTrait {
...
     program_counter: u32,
-    stack: Stack,
-    memory: Memory,
-// TODO: refactor using smart pointers
-// once compiler supports it
-//calling_context: Nullable<ExecutionContext>,
-//sub_context: Nullable<ExecutionContext>,
+    status: Status,
+    call_context: Box<CallContext>,
+    destroyed_contracts: Array<EthAddress>,
```
Eikix, line 134: the fields of dynamic context has been inlined/flatenned into execution context, it was too heavy to manipulate as well in that box

crates/evm/src/context.cairo
```diff
@@ -293,34 +280,16 @@ impl ExecutionContextImpl of ExecutionContextTrait {
...
         // Copy code slice from [pc, pc+len]
-        let code = self
-            .static_context
-            .unbox()
-            .call_context
-            .bytecode()
-            .slice(self.program_counter, len);
+        let code = self.call_context.unbox().bytecode().slice(self.program_counter, len);
 
         self.program_counter += len;
```
Eikix, line 285: still not sure why we increment the pc inside read_code, it feels weird to me

crates/evm/src/context.cairo
```diff
@@ -293,34 +280,16 @@ impl ExecutionContextImpl of ExecutionContextTrait {
...
     #[inline(always)]
-    fn is_root(self: @ExecutionContext) { //TODO: implement this (returns a bool)
-    // self.calling_context.is_null()
-    // true
-    }
-    #[inline(always)]
-    fn is_leaf(self: @ExecutionContext) { //TODO implement this(returns a bool)
-    // self.sub_context.is_null()
+    fn is_root(self: @ExecutionContext) -> bool {
+        *self.context_id == 0
```
Eikix, line 292: root context MUST have id 0

crates/evm/src/context.cairo
```diff
@@ -345,21 +312,6 @@ impl ExecutionContextImpl of ExecutionContextTrait {
     }
 }
 
-impl DefaultExecutionContext of Default<ExecutionContext> {
```
Eikix, line 348: we can delete this and derive default instead

crates/evm/src/errors.cairo
```diff
@@ -21,6 +21,7 @@ enum EVMError {
     TypeConversionError: felt252,
     ReturnDataError: felt252,
     JumpError: felt252,
+    NotImplemented,
```
Eikix, line 24: adding this and making every unimplemented opcode return an error instead of Ok for clarity

crates/evm/src/execution.cairo
```diff
@@ -1,25 +1,17 @@
...
 
 /// Execute EVM bytecode.
-fn execute(
-    call_context: CallContext,
-    starknet_address: ContractAddress,
-    evm_address: EthAddress,
-    gas_limit: u64,
-    gas_price: u64,
-) {
+fn execute() {
```
Eikix, line 8: trimming to the minimum execute as we don't know what it looks like yet

crates/evm/src/instructions/duplication_operations.cairo
```diff
@@ -1,117 +1,118 @@
 //! Duplication Operations.
 
 // Internal imports
-use evm::context::{ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionContextDestruct};
 use evm::errors::EVMError;
+use evm::machine::Machine;
 
 mod internal {
-    use evm::context::{ExecutionContext, ExecutionContextTrait, BoxDynamicExecutionContextDestruct};
+    use evm::context::{ExecutionContext, ExecutionContextTrait,};
```
Eikix, line 8: I wonder when Scarb will warn us about unused imports

crates/evm/src/machine.cairo
```diff
@@ -0,0 +1,246 @@
...
+            current_context: Default::default(),
+            ctx_count: 1,
+            stack: Default::default(),
+            memory: Default::default(),
+            storage_journal: Default::default(),
+        }
+    }
+}
+
+/// A set of getters and setters for the current context
```
Eikix, line 40: lmk if this documentation is enough

crates/evm/src/instructions/duplication_operations.cairo
enitrat: would be nice yeah

crates/evm/src/context.cairo
enitrat: Revert ⊆ Stop so yes

crates/evm/src/context.cairo
```diff
@@ -7,38 +7,77 @@ use utils::traits::{SpanDefault, EthAddressDefault, ContractAddressDefault};
 use starknet::{EthAddress, ContractAddress};
 use starknet::get_caller_address;
 
+#[derive(Drop, Default, Copy)]
+enum Status {
+    #[default]
```
enitrat, line 12: What is the #[default] for? I think it's just for Storage, so doesn't really matter for us

crates/evm/src/context.cairo
```diff
@@ -53,188 +92,152 @@ impl CallContextImpl of CallContextTrait {
...
     // *************************************************************************
 
     #[inline(always)]
-    fn reverted(ref self: ExecutionContext) -> bool {
-        let dyn_ctx = self.dynamic_context.unbox();
-        let reverted = dyn_ctx.reverted;
-        self.dynamic_context = BoxTrait::new(dyn_ctx);
-        reverted
+    fn reverted(self: @ExecutionContext) -> bool {
+        match self.status {
```
enitrat, line 196: why not just one if statement?

```rust
if self.status == Status::Reverted {
    return true;
}
false
```

( you would need to derive PartialEq on Status)

crates/evm/src/context.cairo
```diff
@@ -53,188 +92,152 @@ impl CallContextImpl of CallContextTrait {
...
     }
 
     #[inline(always)]
-    fn stopped(ref self: ExecutionContext) -> bool {
-        let dyn_ctx = self.dynamic_context.unbox();
-        let stopped = dyn_ctx.stopped;
-        self.dynamic_context = BoxTrait::new(dyn_ctx);
-        stopped
+    fn stopped(self: @ExecutionContext) -> bool {
+        match self.status {
```
enitrat, line 205: same

crates/evm/src/context.cairo
enitrat: PUSH_N opcodes should increment PC by N, not `read_code`, as read code should only READ the code. Can you modify this in this PR?

crates/evm/src/machine.cairo
```diff
@@ -0,0 +1,246 @@
...
+/// Rebox the modified current ExecutionContext
+///
+/// Limitations:
+/// 1. We're not able to use @Machine as an argument for getters, as the ExecutionContext struct does not derive the Copy trait.
+/// 2. We must use a box reference to the current context, as the changes made during execution must be applied
+/// to only one ExecutionContext struct instance. Using a pointer ensures we never duplicate structs and thus changes.
+#[generate_trait]
+impl MachineCurrentContextImpl of MachineCurrentContext {
+    #[inline(always)]
+    fn current_ctx_pc(ref self: Machine) -> usize {
```
enitrat, line 60: I'm not sure we should prefix `current_ctx` to all these methods give that we will always access the current_ctx anyway.

Eikix: it's just when i thought about it, for me the Machine has a current_ctx, and then we access the pc of the current ctx I guess, not the pc of the overarching machine

enitrat: yeah I get it but since we always access the current_ctx, is it worth precising or not

crates/evm/src/context.cairo
enitrat: ah ok nice 👍

enitrat (approved):
lgtm

PR #396, opened 2023-10-04
feat: sload opcode

enitrat:
- Implements the `sload` opcode
- Tests for `sload`
- Introduces a `pop_sba` stack method to convert a u256 stack value to `StorageBaseAddress` upon retrieval
- Tests for `pop_sba`
- Implements journal methods allowing to retrieve, write, and finalize values
- Tests for journal methods
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


crates/evm/src/errors.cairo
```diff
@@ -22,7 +25,8 @@ enum EVMError {
     ReturnDataError: felt252,
     JumpError: felt252,
     NotImplemented,
-    UnknownOpcode: u8
+    UnknownOpcode: u8,
+    SyscallFailed: felt252
```
Eikix, line 29: Is this an EVM error?
Should we create a new type of Error, such as StarknetError and implement from trait?

Eikix: imo, this should be a StarknetError (linked to syscall), and be converted to 
EVMError::InternalError

Eikix: Maybe this can be done in a later issue, but essentially have:
EVMError only hold official (from execution-specs) errors and then intermediary Error types that can hold different business logic for errors

crates/utils/src/traits.cairo
```diff
@@ -40,3 +42,41 @@ impl EthAddressIntoU256 of Into<EthAddress, u256> {
         intermediate.into()
     }
 }
+
+//TODO remove once merged in corelib
+impl StorageBaseAddressIntoFelt252 of Into<StorageBaseAddress, felt252> {
```
Eikix, line 47: What is the difference between this file and helpers file? it doens't seem we're separating traits and impl here, 

for instance TryIntoResult lives in helpers, maybe we should find more accurate naming or file structure?

crates/evm/src/storage_journal.cairo
```diff
@@ -0,0 +1,86 @@
...
+use starknet::{StorageBaseAddress, Store, storage_base_address_from_felt252};
+use utils::helpers::ArrayExtensionTrait;
+use utils::traits::{StorageBaseAddressPartialEq, StorageBaseAddressIntoFelt252};
+/// The Journal tracks the changes applied to storage during the execution of a transaction.
+/// Local changes tracks the changes applied inside a single execution context.
+/// Global changes tracks the changes applied in the entire transaction.
+/// Upon exiting an execution context, local changes must be finalized into global changes
+/// Upon exiting the transaction, global changes must be finalized into storage updates.
+#[derive(Destruct, Default)]
+struct Journal {
```
Eikix, line 11: reflecting on it, I wonder what is the value of having local_changes and global_changes,
since the machine is the overarching data structure in our context, it always has access to global_changes in some sense 


-> what if we just had global_changes, and simply, just once, when trying to finalize the root context, we finalize_global on the global_changes of the journal
I think the end result is exactly the same

crates/utils/src/helpers.cairo
```diff
@@ -479,43 +479,28 @@ impl ArrayExtension<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayExtensi
...
-    /// # Arguments
-    ///
-    /// * `self` - The `Span<u8>` instance to pad with zeroes.
-    /// * `n_zeroes` - The number of zeroes to add to the right of the span.
-    ///
-    /// # Returns
-    ///
-    /// A new `Span<u8>` instance which has a length equal to the length of the input
-    /// span plus the number of zeroes specified.
-    fn clone_pad_right(self: Span<u8>, n_zeroes: usize) -> Span<u8> {
```
Eikix, line 500: why don't we need this anymore?

crates/utils/src/helpers.cairo
```diff
@@ -479,43 +479,28 @@ impl ArrayExtension<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayExtensi
             n -= 1;
         };
     }
+
+    // Appends an item only if it is not already in the array.
+    fn append_unique<+PartialEq<T>>(ref self: Array<T>, value: T) {
```
Eikix, line 484: cool idea

Eikix (commented):
beautiful PR! I like the layers of abstraction and the general direction

My main question is:
- do we need local and global changes in journal? My understanding is that having only global_changes is enough,
- we'd call finalize global changes at the end of each execution context (one storage syscall)

=> if the tx reverts, all sstores revert
=> if the tx does not revert, then writing directly from global_changes into storage multiple times at the same slot won't be charged in the L1 calldata, so it's merely a storage syscall cost tradeoff vs. maintaining 2 dicts instead of 1 and doing many dicts writes/accesses between each execution contexts

crates/evm/src/errors.cairo
Eikix: Mhm problem would be: our "?" operator usage will not be as smooth
So I think this conversion can be done inside the outter `run` context, in case we return an error message

Eikix:
Update: after thinking about it for a while, I guess the debate between having local and global vs. only global can be summarized as:
- is it more expensive to hold two dicts and write from one into the other at every sub context finalization
- or is it more expensive to hold just one dict, and write into storage at every sub context finalization

Essentially in my challenge I wonder about the flow:
- just one global change dict
- at every end of sub context: write into storage and clean keys
- inside any context, can use global_changes dict as cache layer
- at the end, in root context, can just write into storage global_changes as if it were any sub context
- if a sub context reverts, then have to clear entries drop keys in the array of key

ERRATUM: this is not so good as in case of sub context reverting, we can't recover the past state of global_changes

OKPM pour local + global

We just need to implement what happens on revert -> clear keys but also overwrite local_changes dict by a fresh empty one

crates/utils/src/traits.cairo
enitrat: I was thinking that ` traits.cairo`. would be for additional impls of existing traits, while `helpers`. would be for our other stuff. but now that I think abt TryIntoResult should be moved in ` traits`

enitrat:
> can you add logic in machine.revert() to wipe the current local changes?

Wouldn't it be for another PR ? I can add a todo

crates/evm/src/errors.cairo
Eikix: created #402

crates/utils/src/traits.cairo
Eikix: i think those names are not so expressive though

traits feels like it's our definition of new traits
helpers is just general

could we rename traits.cairo into traits_extension.cairo
and make sure helpers does not grow too much, if it does we should split it

Eikix:
> > can you add logic in machine.revert() to wipe the current local changes?
> 
> Wouldn't it be for another PR ? I can add a todo

fine with me

Eikix:
lgtm

Eikix (approved):
lgtm

PR #405, opened 2023-10-05
Implementation of rlp::decode()

Quentash:
Implementation of rlp::decode() as it was benchmarked and tested here (https://github.com/danilowhk/cairo-lib/blob/main/src/encoding/rlp.cairo).

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

-
-
-

- [ ] Yes
- [X] No


crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,159 @@
...
+        }
+    }
+}
 
+// @notice Represent a RLP item
+#[derive(Drop)]
+enum RLPItem {
+    Bytes: Bytes,
+    // Should be Span<RLPItem> to allow for any depth/recursion, not yet supported by the compiler
+    List: Span<Bytes>
```
Eikix, line 46: Have you tried to see if self-referencing Enum is supported by the compiler?

Eikix:
The scope of this PR is too large. Could you break it down into two PRs, one for the RLP trait & RLP decode
And the second one for the EOA logic: decode_tx?

crates/utils/src/bitwise.cairo
```diff
@@ -0,0 +1,78 @@
+use utils::math::{pow, pow_felt252};
+use math::Oneable;
+use zeroable::Zeroable;
+
+// @notice Bitwise left shift
+// @param num The number to be shifted
+// @param shift The number of bits to shift
+// @return The left shifted number
+fn left_shift<
```
Eikix, line 9: I think you don't need this, you can use the existing bitshift trait in Math.cairo

crates/utils/src/bitwise.cairo
```diff
@@ -0,0 +1,78 @@
+use utils::math::{pow, pow_felt252};
```
Eikix, line 1: note that we already have a bitshift trait and library in math.cairo,

crates/utils/src/bitwise.cairo
```diff
@@ -0,0 +1,78 @@
...
+    impl TDrop: Drop<T>
+// TODO refactor shift type from T to usize
+>(
+    num: T, shift: T
+) -> T {
+    let two = TOneable::one() + TOneable::one();
+    num * pow(two, shift)
+}
+
+fn left_shift_felt252(num: felt252, shift: felt252) -> felt252 {
```
Eikix, line 26: same afaik

crates/utils/src/bitwise.cairo
```diff
@@ -0,0 +1,78 @@
...
+    num: T, shift: T
+) -> T {
+    let two = TOneable::one() + TOneable::one();
+    num / pow(two, shift)
+}
+
+// @notice Bit length of a number
+// @param num The number to be measured
+// @return The number of bits in the number
+fn bit_length<
```
Eikix, line 55: this looks very expensive, why do we need this? I don't see it being used

crates/utils/src/eth_transaction.cairo
```diff
@@ -30,22 +42,54 @@ trait EthTransaction {
...
-    fn decode_tx(tx_data: Span<u8>) -> EthereumTransaction;
+    fn decode_tx(tx_data: Span<u8>) -> EthereumTransaction {
+        let mut eth_transaction: EthereumTransaction = Default::default();
+
+        let (items, size) = rlp::rlp_decode(tx_data).unwrap();
+
+        match items {
+            rlp::RLPItem::Bytes(value) => {},
+            rlp::RLPItem::List(list) => {
+                let nonce: u256 = (*list[0]).try_into().unwrap();
```
Eikix, line 53: we must remove every unwrap(), and handle the error gracefully using a Result<>

1. Create a new error enum (like we have for EVMError), call it `InternalError`, and add a TypeConversionError alternative in the enum,
2. modify the function signature to return a Result<EthereumTransaction, InternalError>
3. modify the signature of the TryInto() trait for Span<Bytes> so that it returns a result rather than an option. Since you can't do that on TryInto trait, do it with TryIntoResult trait
4. instead of unwrap(), use "?" operator

crates/utils/src/bitwise.cairo
Eikix: but good idea to make it generic, i'll create an issue for this

crates/utils/src/types/byte.cairo
```diff
@@ -0,0 +1,16 @@
+use utils::bitwise::right_shift;
+
+type Byte = u8;
+
+#[generate_trait]
+impl ByteImpl of ByteTrait {
+    // @notice Extracts the high and low nibbles from a byte
+    // @return (high, low)
+    fn extract_nibbles(self: Byte) -> (Byte, Byte) {
```
Eikix, line 9: we don't use this, we can delete it

crates/utils/src/types/byte.cairo
```diff
@@ -0,0 +1,16 @@
+use utils::bitwise::right_shift;
+
+type Byte = u8;
```
Eikix, line 3: I think we can do without this type alias

Let's replace every occurence of Byte by u8
Let's replace every occurence of Bytes by Span<u8>

crates/utils/src/types/bytes.cairo
```diff
@@ -0,0 +1,104 @@
...
+            result += left_shift_felt252(byte, 8 * (offset - i.into()));
+
+            i += 1;
+        };
+
+        Option::Some(result)
+    }
+}
+
+impl BytesPartialEq of PartialEq<Bytes> {
```
Eikix, line 64: Do we need this? When do we have snapshots of Spans?

crates/utils/src/helpers.cairo
```diff
@@ -504,6 +505,34 @@ impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtensionTrait<T> {
...
+
+        let offset = len.into() - 1;
+        let mut result: u256 = 0;
+        let mut i: usize = 0;
+        loop {
+            if i >= len {
+                break ();
+            }
+            let byte: u256 = (*self.at(i)).into();
+            result += Bitshift::shl(byte, 8 * (offset - i.into()));
```
Eikix, line 527:
```suggestion
            result +=byte.shl(8 * (offset - i.into())));
```

crates/utils/src/math.cairo
```diff
@@ -231,3 +232,28 @@ impl U128WrappingBitshiftImpl of WrappingBitshift<u128> {
     }
 }
 
+fn pow<
```
Eikix, line 235: why do we need this? we already have a pow fn

crates/utils/src/math.cairo
```diff
@@ -231,3 +232,28 @@ impl U128WrappingBitshiftImpl of WrappingBitshift<u128> {
...
+    base: T, mut exp: T
+) -> T {
+    if exp.is_zero() {
+        TOneable::one()
+    } else {
+        base * pow(base, exp - TOneable::one())
+    }
+}
+
+fn pow_felt252(base: felt252, exp: felt252) -> felt252 {
```
Eikix, line 253: I don't think we need this either

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+use array::{Array, ArrayTrait, Span, SpanTrait};
+use clone::Clone;
+use traits::{Into, TryInto};
+use utils::helpers::SpanU8TryIntoU256;
 
-// We can reuse https://github.com/HerodotusDev/cairo-lib/blob/main/src/encoding/rlp.cairo 
-// Check before reusing that it works soundly
+// @notice Enum with all possible RLP types
+#[derive(Drop, PartialEq)]
+enum RLPType {
```
Eikix, line 10: you can now write:

```
#[derive(Drop, PartialEq)]
enum RLPType {
    String,
    StringShort,
    StringLong,
    ListShort,
    ListLong,
}
```

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+            Result::Ok(RLPType::String(()))
+        } else if byte <= 0xb7 {
+            Result::Ok(RLPType::StringShort(()))
+        } else if byte <= 0xbf {
+            Result::Ok(RLPType::StringLong(()))
+        } else if byte <= 0xf7 {
+            Result::Ok(RLPType::ListShort(()))
+        } else if byte <= 0xff {
+            Result::Ok(RLPType::ListLong(()))
+        } else {
```
Eikix, line 34: can we ever reach this case?

Eikix: if not we can remove the result fn signature type

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+            Result::Ok((RLPItem::Bytes(arr.span()), 1))
+        },
+        RLPType::StringShort(()) => {
+            let len = prefix.into() - 0x80;
+            let res = input.slice(1, len);
+
+            Result::Ok((RLPItem::Bytes(res), 1 + len))
+        },
+        RLPType::StringLong(()) => {
+            let len_len = prefix.into() - 0xb7;
```
Eikix, line 68: why len_len name?

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+            let len_span = input.slice(1, len_len);
+
+            // Bytes => u256 => u32
+            let len256: u256 = len_span.try_into().unwrap();
```
Eikix, line 72: why do we unwrap here?

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+            let len: u32 = len256.try_into().unwrap();
```
Eikix, line 73: you can just implement a try_into() for Span<u8> to u32 imo

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+        } else if byte <= 0xff {
+            Result::Ok(RLPType::ListLong(()))
+        } else {
+            Result::Err('Invalid byte')
+        }
+    }
+}
 
+// @notice Represent a RLP item
+#[derive(Drop)]
```
Eikix, line 41: You can derive PartialEq here

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+                input = input.slice(decoded_len, input.len() - decoded_len);
+            },
+            RLPItem::List(_) => { panic_with_felt252('Recursive list not supported'); }
+        }
+        i += decoded_len;
+    };
+    output.span()
+}
+
+impl RLPItemPartialEq of PartialEq<RLPItem> {
```
Eikix, line 121: Why do you need this? We can't derive PartialEq on RLPItem?

crates/utils/src/tests/test_math.cairo
```diff
@@ -1,6 +1,6 @@
 use integer::{u256_overflowing_add, BoundedInt, u512};
 use utils::math::{
-    Exponentiation, WrappingExponentiation, u256_wide_add, Bitshift, WrappingBitshift
+    Exponentiation, WrappingExponentiation, u256_wide_add, Bitshift, WrappingBitshift, pow
```
Eikix, line 3: why do we import pow

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -0,0 +1,1583 @@
...
+
+    arr.pop_front();
+    let expected_item = RLPItem::Bytes(arr.span());
+
+    assert(res == expected_item, 'Wrong value');
+}
+
+#[test]
+#[available_gas(99999999)]
+fn test_rlp_decode_long_string_len_of_len_1() {
```
Eikix, line 104: cannot understand with the naming

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -0,0 +1,1583 @@
...
+    let mut arr = array![
```
Eikix, line 105: could you add links of where you found the test cases next to test case? i think it'd help maintainability

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -0,0 +1,1583 @@
...
+        0x86,
+        0x19,
+        0x7d,
+        0xd9
+    ];
+
+    let (res, len) = rlp_decode(arr.span()).unwrap();
+    assert(len == 1 + (0xb8 - 0xb7) + 0x3c, 'Wrong len');
+
+    arr.pop_front();
```
Eikix, line 173: can you add a comment why we need to pop twice?

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -0,0 +1,1583 @@
...
+    arr.pop_front();
+    let expected_item = RLPItem::Bytes(arr.span());
+
+    assert(res == expected_item, 'Wrong value');
+}
+
+#[test]
+#[available_gas(99999999)]
+fn test_rlp_decode_long_string_len_of_len_2() {
```
Eikix, line 182: cannot understand with this naming

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+    Bytes: Span<u8>,
+    // Should be Span<RLPItem> to allow for any depth/recursion, not yet supported by the compiler
+    List: Span<Span<u8>>
+}
+
+// @notice RLP decodes a rlp encoded byte array
+// @param input RLP encoded bytes
+// @return Result with RLPItem and size of the decoded item
+fn rlp_decode(input: Span<u8>) -> Result<(RLPItem, usize), felt252> {
+    let prefix = *input.at(0);
```
Eikix, line 52: we can check if input.is_empty() and return an error in case it is empty
we can also probably create an enum RLPError for this library with all the different error cases encoded as felt252 (like EVMError)

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+enum RLPItem {
+    Bytes: Span<u8>,
+    // Should be Span<RLPItem> to allow for any depth/recursion, not yet supported by the compiler
+    List: Span<Span<u8>>
+}
+
+// @notice RLP decodes a rlp encoded byte array
+// @param input RLP encoded bytes
+// @return Result with RLPItem and size of the decoded item
+fn rlp_decode(input: Span<u8>) -> Result<(RLPItem, usize), felt252> {
```
Eikix, line 51: why do we need to know the size of the decoded item?

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,158 @@
...
+    // Unwrap is impossible to panic here
+    let rlp_type = RLPTypeTrait::from_byte(prefix).unwrap();
+    match rlp_type {
+        RLPType::String(()) => {
+            let mut arr = array![prefix];
+            Result::Ok((RLPItem::Bytes(arr.span()), 1))
+        },
+        RLPType::StringShort(()) => {
+            let len = prefix.into() - 0x80;
+            let res = input.slice(1, len);
```
Eikix, line 63: do you know what will happen in case the RLP is malformed, and has 
input.len() < prefix.into() - 0x80?

will slice panic? 

in this case, we can add a check for this case and return an Error with `RlpInconsistentLengthAndData`

Kind of like here:
https://github.com/paritytech/parity-common/blob/master/rlp/src/error.rs

adapt as you need

crates/utils/src/rlp.cairo
Quentash: This is how it was implemented in Herodotus ( https://github.com/HerodotusDev/cairo-lib/blob/main/src/encoding/rlp.cairo ).
`rlp_decode` and `rlp_decode_list` functions seem to be used recursively to decode the "whole" RLP input and `rlp_decode_list` seems to use the len returned by decode_rlp().
Do you want me to find an other way to implement it so that it doesn't return the size ?

crates/utils/src/tests/test_rlp.cairo
Quentash: Those tests have been written by a Herodotus member. I noticed that they are slightly written differently from here (https://github.com/HerodotusDev/cairo-lib/blob/main/src/encoding/tests/test_rlp.cairo), but the tests cases looks identicals.
I ignore the reason tests have different inputs tho, do you want me to change our tests to be exactly the same than Herodotus ? I could also try to reach the tests author to understand why the inputs are different.

crates/utils/src/rlp.cairo
Quentash: I don't think you reach such a malformation. To make slice panic, `prefix` should be lower than 0x80. Which would have therefore given an other `rlp_type` from `let rlp_type = RLPTypeTrait::from_byte(prefix);`.
It would indeed be matched to a string which wouldn't make this slice panic.

crates/utils/src/tests/test_rlp.cairo
Eikix: I think we can have the same tests yes,
As well as the same tests as geth,
Official Execution-specs https://github.com/ethereum/execution-specs/tree/master

as we probably don't want 100 tests, you can pick a few from herodotus, some from geth, and most from eels

crates/utils/src/rlp.cairo
Eikix: I think it's fine for now, for our knowledge and culture, you could research various implementations of RLP and forge your opinion on your favorite

crates/utils/src/rlp.cairo
Eikix: yes but it can happen that an RLP payload is malformed:
- prefix is some value and the underlying data is something else -> prefix is short string, payload is long list or idk

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,121 @@
...
+    let mut i = 0;
+    let len = input.len();
+    let mut output = ArrayTrait::new();
+
+    loop {
+        if i >= len {
+            break ();
+        }
+
+        let (decoded, decoded_len) = rlp_decode(input).unwrap();
```
enitrat, line 110: this unwrap is unsafe, what happens if the decoding fails?

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -0,0 +1,1590 @@
...
+use result::ResultTrait;
+use utils::errors::{RLPError, RLP_INVALID_LENGTH};
+
+#[test]
+#[available_gas(9999999)]
+fn test_rlp_types() {
+    let mut i = 0;
+    loop {
+        if i <= 0x7f {
+            assert(RLPTypeTrait::from_byte(i) == RLPType::String(()), 'Parse type String');
```
enitrat, line 12:
```suggestion
            assert(RLPTypeTrait::from_byte(i) == RLPType::String, 'Parse type String');
```

crates/utils/src/tests/test_rlp.cairo
Quentash: My very very very bad, the tests are actually identical, but Herodotus' have the inputs encoded in u64 little endian instead of u8. I am sorry for the misinformation.

crates/utils/src/rlp.cairo
Quentash: I've added verification on the inputs length so that slice() functions will never panic because we now make sure there are enough bytes to slice.

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,121 @@
...
+            RLPType::ListLong(())
+        }
+    }
+}
 
+// @notice Represent a RLP item
+#[derive(Drop, PartialEq)]
+enum RLPItem {
+    Bytes: Span<u8>,
+    // Should be Span<RLPItem> to allow for any depth/recursion, not yet supported by the compiler
```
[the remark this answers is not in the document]
Quentash: To be fair, I imported this as it was here https://github.com/danilowhk/cairo-lib/blob/main/src/encoding/rlp.cairo
@danilowhk could you help me explaning this part ?

crates/utils/src/rlp.cairo
Quentash: I implemented an error verification and propagate the rlp_decode.
I couldn' t "return" from inside a loop so it doesn't look very appealing, should it panic instead ?

crates/utils/src/helpers.cairo
```diff
@@ -505,35 +506,36 @@ impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtensionTrait<T> {
     }
 }
 
-impl SpanU8TryIntoU32 of TryInto<Span<u8>, u32> {
-    fn try_into(self: Span<u8>) -> Option<u32> {
-        let len = self.len();
+trait BytesSerde<T> {
+    /// Serialize/deserialize bytes into/from
+    /// an array of bytes.
+    fn serialize(self: @T, ref output: Array<u8>);
```
enitrat, line 512: remove serialize from the traits as it's never used. We don't want to mistakenly end up using it if it's not implemented

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,153 @@
...
+        match decoded {
+            RLPItem::Bytes(b) => {
+                output.append(b);
+                input = input.slice(decoded_len, input.len() - decoded_len);
+            },
+            RLPItem::List(_) => { panic_with_felt252('Recursive list not supported'); }
+        }
+        i += decoded_len;
+    };
+    if decode_error != Option::None {
```
enitrat, line 149:
```suggestion
    if decode_error.is_some() {
```

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,153 @@
...
+            Result::Ok((RLPItem::List(res), 1 + len_of_len + len))
+        }
+    }
+}
+
+fn rlp_decode_list(ref input: Span<u8>) -> Result<Span<Span<u8>>, RLPError> {
+    let mut i = 0;
+    let len = input.len();
+    let mut output = ArrayTrait::new();
+    let mut decode_error: Option<RLPError> = Option::None;
```
enitrat, line 127: remove this variable

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,153 @@
...
+
+    loop {
```
enitrat, line 129: let err = loop {...}

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,153 @@
...
+        if i >= len {
+            break ();
+        }
+
+        let res = rlp_decode(input);
+        if res.is_err() {
+            decode_error = Option::Some(res.unwrap_err());
```
enitrat, line 136:
```suggestion
        break Option::Some(res.unwrap_err());
```

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,153 @@
...
+
+fn rlp_decode_list(ref input: Span<u8>) -> Result<Span<Span<u8>>, RLPError> {
+    let mut i = 0;
+    let len = input.len();
+    let mut output = ArrayTrait::new();
+    let mut decode_error: Option<RLPError> = Option::None;
+
+    loop {
+        if i >= len {
+            break ();
```
enitrat, line 131:
```suggestion
            break None
```

crates/utils/src/rlp.cairo
```diff
@@ -1,7 +1,153 @@
...
+        }
+
+        let res = rlp_decode(input);
+        if res.is_err() {
+            decode_error = Option::Some(res.unwrap_err());
+            break;
+        }
+        let (decoded, decoded_len) = res.unwrap();
```
enitrat, line 139: best would still be to avoid using an unwrap and just use the match control flow on `res` directly

enitrat (commented):
almos lgtm now wdyt @eikix

Eikix:
> almos lgtm now wdyt @Eikix

agreed

crates/utils/src/rlp.cairo
enitrat: you break with a value from the loop and then return depending on this value

crates/utils/src/helpers.cairo
```diff
@@ -505,35 +506,34 @@ impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtensionTrait<T> {
     }
 }
 
-impl SpanU8TryIntoU32 of TryInto<Span<u8>, u32> {
-    fn try_into(self: Span<u8>) -> Option<u32> {
-        let len = self.len();
+trait BytesSerde<T> {
+    /// Serialize/deserialize bytes into/from
+    /// an array of bytes.
+    fn deserialize(self: @Span<u8>) -> Option<T>;
```
enitrat, line 512: my bad. We don't want to pass it by snapshot

```suggestion
    fn deserialize(self: Span<u8>) -> Option<T>;
```

crates/utils/src/helpers.cairo
```diff
@@ -505,35 +506,34 @@ impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtensionTrait<T> {
...
+}
+
+impl BytesSerdeU32Impl of BytesSerde<u32> {
+    fn deserialize(self: @Span<u8>) -> Option<u32> {
+        let len = (*self).len();
```
enitrat, line 517: if not passed by snapshot then we can just

```suggestion
        let len = self.len();
```

crates/utils/src/helpers.cairo
```diff
@@ -505,35 +506,34 @@ impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtensionTrait<T> {
...
-        let mut i: usize = 0;
+        let mut i: u32 = 0;
         loop {
-            if i >= len {
+            if i == len {
                 break ();
             }
-            let byte: u32 = (*self.at(i)).into();
-            result += byte.shl(8 * (offset - i.into()));
+            let byte: u32 = (*(*self).at(i)).into();
```
enitrat, line 528: and this can be simplified as well back to
```suggestion
            let byte: u32 = (*self.at(i)).into();
```

crates/utils/src/rlp.cairo
```diff
@@ -16,98 +16,125 @@ enum RLPType {
...
     match rlp_type {
-        RLPType::String(()) => {
+        RLPType::String => {
             let mut arr = array![prefix];
             Result::Ok((RLPItem::Bytes(arr.span()), 1))
         },
-        RLPType::StringShort(()) => {
+        RLPType::StringShort => {
             let len = prefix.into() - 0x80;
+            if input.len() < len + 1 {
```
enitrat, line 70:
```suggestion
            if input.len() <= len {
```

crates/utils/src/rlp.cairo
```diff
@@ -16,98 +16,125 @@ enum RLPType {
...
             let res = input.slice(1, len);
 
             Result::Ok((RLPItem::Bytes(res), 1 + len))
         },
-        RLPType::StringLong(()) => {
-            // Extract the amount of bytes representing the data payload length
-            let len_len = prefix.into() - 0xb7;
-            let len_span = input.slice(1, len_len);
+        RLPType::StringLong => {
+            let len_of_len = prefix.into() - 0xb7;
```
enitrat, line 78:
```suggestion
            let len_bytes_count = prefix.into() - 0xb7;
```

count of how many bytes encode the length

crates/utils/src/rlp.cairo
```diff
@@ -16,98 +16,125 @@ enum RLPType {
...
+            if input.len() < len_of_len + 1 {
+                return Result::Err(RLPError::RlpInputTooShort(RLP_INPUT_TOO_SHORT));
+            }
+
+            let len_in_bytes = input.slice(1, len_of_len);
+            let len: u32 = len_in_bytes.deserialize().unwrap();
```
enitrat, line 84:
```suggestion
            let string_len: u32 = len_in_bytes.deserialize().unwrap();
```

crates/utils/src/rlp.cairo
```diff
@@ -16,98 +16,125 @@ enum RLPType {
...
-            // Extract the amount of bytes representing the data payload length
-            let len_len = prefix.into() - 0xb7;
-            let len_span = input.slice(1, len_len);
+        RLPType::StringLong => {
+            let len_of_len = prefix.into() - 0xb7;
+            if input.len() < len_of_len + 1 {
+                return Result::Err(RLPError::RlpInputTooShort(RLP_INPUT_TOO_SHORT));
+            }
+
+            let len_in_bytes = input.slice(1, len_of_len);
```
enitrat, line 83:
```suggestion
            let string_len_bytes = input.slice(1, len_of_len);
```

crates/utils/src/rlp.cairo
```diff
@@ -16,98 +16,125 @@ enum RLPType {
...
+            let len: u32 = len_in_bytes.deserialize().unwrap();
+            if input.len() < len + len_of_len + 1 {
+                return Result::Err(RLPError::RlpInputTooShort(RLP_INPUT_TOO_SHORT));
+            }
 
-            // Bytes => u32
-            let len: u32 = len_span.try_into().unwrap();
-            let res = input.slice(1 + len_len, len);
+            let res = input.slice(1 + len_of_len, len);
```
enitrat, line 89:
```suggestion
            let decoded_string = input.slice(1 + len_of_len, len);
```

Also call it `decoded_string` instead of `res` in other branches

crates/utils/src/rlp.cairo
```diff
@@ -16,98 +16,125 @@ enum RLPType {
...
-            let len_len = prefix.into() - 0xf7;
-            let len_span = input.slice(1, len_len);
-
-            // Bytes => u32
-            let len: u32 = len_span.try_into().unwrap();
-            let mut in = input.slice(1 + len_len, len);
-            let res = rlp_decode_list(ref in);
-            Result::Ok((RLPItem::List(res), 1 + len_len + len))
+        RLPType::ListLong => {
+            let len_of_len = prefix.into() - 0xf7;
```
enitrat, line 104: same names as in StringLong

enitrat (commented):
lgtm, proposed variable name changes to make it easier to understand on first read.

crates/utils/src/rlp.cairo
Quentash: My bad for this one, I assumed  "<" and a "+" would be cheaper than a "<="

Issue #408, opened 2023-10-06
dev: make Exponentiation in math.cairo completely generic


Eikix:
```[tasklist]
### Tasks
- [ ] refactor Exponentiation (not WrappingExponentiation) to have trait bounds on T, T should implement addition, multiplication and all the traits necessary to perform exponentiation. Can take reference in [Herodotus' cairo-lib](https://github.com/HerodotusDev/cairo-lib/blob/main/src/utils/math.cairo)
- [ ] then remove implementations of Exponentiation for u256 and u128 
- [ ] everything should still work, tests should pass
```

akhercha:
Hi, happy to try this one!

PR #410, opened 2023-10-06
dev: Exponentiation & Bitshift generics

akhercha:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #408 

## What is the new behavior?

- Added `ExponentiationImpl<T, ...> of Exponentiation<T>`
    - then removed `U256ExpImpl` & `U128ExpImpl`

- Added `BitshiftImpl<T, ...> of Bitshift<T>`
    - then removed `U256BitshiftImpl` & `U128BitshiftImpl`

- [ ] Yes
- [x] No


crates/utils/src/math.cairo
```diff
@@ -129,21 +125,25 @@ trait Bitshift<T> {
...
+    impl TCopy: Copy<T>,
+    impl TDrop: Drop<T>
+> of Bitshift<T> {
+    fn shl(self: T, shift: T) -> T {
+        let two = TOneable::one() + TOneable::one();
+        self * two.pow(shift)
     }
 
-    fn shr(self: u256, shift: u256) -> u256 {
-        if shift > 255 {
```
Eikix, line 142: I'm sad that we loose this^^

crates/utils/src/math.cairo
```diff
@@ -9,6 +10,27 @@ trait Exponentiation<T> {
     fn pow(self: T, exponent: T) -> T;
 }
 
+impl ExponentiationImpl<
+    T,
+    impl TZeroable: Zeroable<T>,
```
Eikix, line 15: please use the new syntax:
`impl TZeroable: Zeroable<T>,` => `+Zeroable<T>`


Make sure you have the latest scarb (2.3.0-rc1)

crates/utils/src/math.cairo
```diff
@@ -1,6 +1,7 @@
 use integer::{
     u256, u256_overflow_mul, u256_overflowing_add, u512, BoundedInt, u128_overflowing_mul
 };
+use math::{Zeroable, Oneable};
```
Eikix, line 4: We were told by Ori, lead of the Starkware compiler team that we should not use Zeroable and Oneable (although they're publicly exported).

Could you implement a Zero trait and One trait in this very file, math.cairo?
It'd have the same API afaik, @enitrat may have additional thoughts

Eikix (commented):
Nice PR overall,
Good catch that wrappingbitshift and wrappingpow are not able to be generic over T

crates/utils/src/math.cairo
enitrat: I started working on this this morning and created an issue on the Cairo compiler side.

In the meantime, let's do as usual and implement it in our codebase under utils/traits following the solution proposed here https://github.com/starkware-libs/cairo/issues/4236

akhercha: Alright, should be good then!

Also regarding the error I had, I just did `rm -rf ~/.cache/scarb` and my code compiles now. :D

enitrat:
Regular `shr` is supposed to panic indeed

akhercha:
Should be good then!
Let me know if I missed something. 🙏

crates/utils/src/mem.cairo
enitrat, line 1: I know it's inside std::mem in Rust but since we won't be doing any memory operations, perhaps it would be better somewhere else. I'd rather have Zero, One, and SizeOf inside a `num` module

enitrat (commented):
Great work

crates/utils/src/math.cairo
```diff
@@ -1,6 +1,223 @@
...
+trait Zero<T> {
+    /// Returns the multiplicative identity element of Self, 0.
+    fn zero() -> T;
+    /// Returns whether self is equal to 0, the multiplicative identity element.
+    fn is_zero(self: @T) -> bool;
+    /// Returns whether self is not equal to 0, the multiplicative identity element.
+    fn is_non_zero(self: @T) -> bool;
+}
+
+impl U8Zero of Zero<u8> {
```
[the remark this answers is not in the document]
akhercha: That would be:
```rust
impl ZeroImpl<T, +Zeroable<T>, +PartialEq<T>, +Drop<T>, +Copy<T>> of Zero<T> {
    fn zero() -> T {
        Zeroable::zero()
    }
    fn is_zero(self: @T) -> bool {
        *self == ZeroImpl::zero()
    }

    fn is_non_zero(self: @T) -> bool {
        !self.is_zero()
    }
}

impl OneImpl<T, +Oneable<T>, +PartialEq<T>, +Drop<T>, +Copy<T>> of One<T> {
    fn one() -> T {
        Oneable::one()
    }

    fn is_one(self: @T) -> bool {
        *self == OneImpl::one()
    }

    fn is_non_one(self: @T) -> bool {
        !self.is_one()
    }
}
```

But regarding of what Eikix said,
> We were told by Ori, lead of the Starkware compiler team that we should not use Zeroable and Oneable

Is it still fine to use them here? Or did I get it wrong?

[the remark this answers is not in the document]
enitrat: > That would be:

Yes, exactly!

> We were told by Ori, lead of the Starkware compiler team that we should not use Zeroable and Oneable

To be honest I don't really understand this decision - if it's available _publically_ in the _core library_ in and it does what we want it to do - then I would say let's use it until we have a better solution. 😎 
I find it a lot more elegant to have this generic impl then just reinventing the wheel.

The only problem is for Felt252Oneable which is required for exponentiation and not implemented in the corelib (and the associated PR was refused). So let's add our own manual `Felt252One` trait impl. And whenever `One` is merged in the core library, then we will just need to change the import path from our local repo to whatever the path in the corelib is.

Thoughts @Eikix ?

enitrat:
Also @akhercha I think your local git is not synced to your GH account - not sure if it's on purpose or not

crates/utils/src/math.cairo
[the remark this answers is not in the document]
Eikix: Alright, sounds fine to me!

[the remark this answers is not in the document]
akhercha: 🤝 Updated!

akhercha:
Ah no indeed, something exploded in my config I guess, gonna check that thanks!

enitrat:
@akhercha I just pushed a change with more optimized gas usage - basically since Oneable and Zeroable fns are not inlined there's a small overhead in gas consumption if we rely on them for genericity. Therefore I changed my mind about my previous comment, and explicitly implemented Zero and One for u128, u256, and felt252 ( we won't need other impls, at least for now).

This nullifies any increase in gas consumption for `pow`-dependent functions; however there's still an increase for `shr`. and `shl`, I'm looking for the cause

Eikix (approved):
lgtm 🔥

Issue #402, opened 2023-10-05
dev: create two new error enums: VM error and InternalError


Eikix:
```[tasklist]
### Tasks
- [ ] Create the StarknetError enum and refactor all Starknet specific errors inside it
- [ ] Create the VMError enum and it should be an enum of sub error types: EVMError and StarknetError
- [ ] Refactor opcodes return signature to return a VMError instead of an EVMError
```

enitrat:
are `TypeConversionError` and else StarknetError? If yes the name doesn't seem very adapted to me. I'd prefer something more generic, like `InternalError`

Eikix:
Ahh could be StarknetError for syscalls, 
TypeConversionError should be something else I agree

enitrat:
enumception

azurwastaken:
Hello !
are you working on it @Eikix or can i do it ?

Eikix:
> Hello !
> 
> are you working on it @Eikix or can i do it ?

Hey:), thanks for your swift response, after discussing with @enitrat, we're waiting a bit to implement this feature to see if it makes sense in our current architecture,

For now let's keep using EVMError everywhere as a global error type for our repository

azurwastaken:
> > Hello !
> > are you working on it @Eikix or can i do it ?
> 
> Hey:), thanks for your swift response, after discussing with @enitrat, we're waiting a bit to implement this feature to see if it makes sense in our current architecture,
> 
> For now let's keep using EVMError everywhere as a global error type for our repository

Oh ok I see !
I was working on a solution yesterday that would have allowed to have more specified errors type and also allowed to had type of errors without breaking any existing code. So if you change your mind feel free to assign me to this task :)

Eikix:
> > > Hello !
> 
> > > are you working on it @Eikix or can i do it ?
> 
> > 
> 
> > Hey:), thanks for your swift response, after discussing with @enitrat, we're waiting a bit to implement this feature to see if it makes sense in our current architecture,
> 
> > 
> 
> > For now let's keep using EVMError everywhere as a global error type for our repository
> 
> 
> 
> Oh ok I see !
> 
> I was working on a solution yesterday that would have allowed to have more specified errors type and also allowed to had type of errors without breaking any existing code. So if you change your mind feel free to assign me to this task :)
> 
> 

I'm open to seeing your design proposal, feel free to open a draft pr and we can discuss it

Eikix:
> > > Hello !
> 
[...]

PR #412, opened 2023-10-09
dev: enums VM error and InternalError

azurwastaken:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #402 

## What is the new behavior?

- Added a new enum that act as Interface for Errors which add the possibility for new errors type in the future without breaking existing code
- Added two enums for InternalError and EVMError (their content must be discussed/modified)

- [ ] Yes
- [X] No

## More Informations

@Eikix Here is my proposal for your issue about the errors handling, feel free to give me feedback/improvement.


Eikix (commented):
Interesting PR! Let me think about it for a bit,
Summoning @enitrat and @ClementWalter who'll also have an opinion 

I'll check again how Rust handles Error management (i usually use `thiserror`)

crates/evm/src/errors.cairo
```diff
@@ -21,32 +21,57 @@ const WRITE_IN_STATIC_CONTEXT: felt252 = 'KKT: WriteInStaticContext';
 const READ_SYSCALL_FAILED: felt252 = 'KKT: read syscall failed';
 
 #[derive(Drop, Copy, PartialEq)]
-enum EVMError {
+enum EVMErrorEnum {
```
enitrat, line 24: If it's an `enum` then calling it `XEnum` is not a good practice imo. It's already defined as an enum, we wouldn't say `PersonStruct` or `balance_u256` if we have type information

crates/evm/src/instructions/block_information.cairo
```diff
@@ -13,48 +13,48 @@ impl BlockInformation of BlockInformationTrait {
...
+        Result::Err(Errors::EVMError(EVMErrorEnum::NotImplemented))
     }
 
     /// 0x41 - COINBASE
     /// Get the block's beneficiary address.
     /// # Specification: https://www.evm.codes/#41?fork=shanghai
-    fn exec_coinbase(ref self: Machine) -> Result<(), EVMError> {
-        Result::Err(EVMError::NotImplemented)
+    fn exec_coinbase(ref self: Machine) -> Result<(), Errors> {
+        Result::Err(Errors::EVMError(EVMErrorEnum::NotImplemented))
```
enitrat, line 24:
```suggestion
        Result::Err(Errors::EVMError::NotImplemented))
```

I'd rather have something like this for internal error as well

```rust
Errors::InternalError::TypeConversionError(TYPE_CONVERSION_ERROR)
```

thought idk if it's possible

enitrat (commented):
Few remarks

- This PR only brings one extra layer of abstraction, and I don't feel like there is any benefits to this migration. The logic is still handled the same and the separation between EVM errors and Internal errors doesn't add any value since both end up merged into the `Errors` type which is used in EVM functions

- Overall it makes the code very verbose 

Perhaps a better approach would be to keep it as it currently is, and just implement the an error management logic that acts based on enum variants

azurwastaken:
> Few remarks
> 
> * This PR only brings one extra layer of abstraction, and I don't feel like there is any benefits to this migration. The logic is still handled the same and the separation between EVM errors and Internal errors doesn't add any value since both end up merged into the `Errors` type which is used in EVM functions
> * Overall it makes the code very verbose
> 
> Perhaps a better approach would be to keep it as it currently is, and just implement the an error management logic that acts based on enum variants

I agree with the verbosity, maybe there is a way to shorten the way its called.

To me, those change would help to understand trully the cause of an error and would make the debug process faster. Also, the Errors enum work like a pure virtual interface that allow implementation of new types of errors without having to change a single prototype which is, imo, a good things for sustainability.

Eikix:
Ideally I'd like our error management to be the same as Rust, 
That's not possible for now as the ? operator does not implicitly try to apply the "Into" trait

But ideally I'd like for utils error enum to be decoupled from any other crate

EVMError would also be completely self comprised.

If in evm crate, a function uses a util to compute some information, and that util can return an Result::err, then I'd like for the evm function to perform some conversion in a light way.

How about map_err()?

We could remain rustic while waiting for "?" operator to support our desired behaviour?

Eikix:
Closing this for now, we can revisit the error handling, using the knowledge we developed thanks to this PR and our discussion

Issue #383, opened 2023-09-28
bug: returndatacopy, returndatasize should use return data from subctx


enitrat:
## Bug Report

**Kakarot version:**

**Current behavior:**
Copies the returndata of the current context instead of the subcontext.

**Expected behavior:**
Copies the returndata of the last executed subcontext

**Steps to reproduce:**

**Related code:**

**Other information:**

PR #419, opened 2023-10-11
fix: returndata

enitrat:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #383

## What is the new behavior?

- Correct behavior of returndatasize/returndatacopy
- Got rid of the `child_return_data` field
- `return_data` now refers to the return_data of the last executed subcontext
- renamed machine `revert` and `stop` to `set_reverted` and `set_stopped` to avoid confusion with `revert` and `stop` opcodes
- Unexpected revert (not the REVERT Opcode) reason is not appended to the parent's return data field.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/context.cairo
+++ b/crates/evm/src/context.cairo
@@ -140,9 +140,9 @@ struct ExecutionContext {
     destroyed_contracts: Array<EthAddress>,
     events: Array<Event>,
     create_addresses: Array<EthAddress>,
-    return_data: Array<u8>,
+    // Return data of a child context.
+    return_data: Span<u8>,
     parent_ctx: Nullable<ExecutionContext>,
-    child_return_data: Option<Span<u8>>,
 }
 
 impl DefaultBoxExecutionContext of Default<Box<ExecutionContext>> {
@@ -165,8 +165,7 @@ impl ExecutionContextImpl of ExecutionContextTrait {
         starknet_address: ContractAddress,
         call_ctx: CallContext,
         parent_ctx: Nullable<ExecutionContext>,
-        child_return_data: Option<Span<u8>>,
-        return_data: Array<u8>,
+        return_data: Span<u8>,
     ) -> ExecutionContext {
         ExecutionContext {
             id,
@@ -180,7 +179,6 @@ impl ExecutionContextImpl of ExecutionContextTrait {
             create_addresses: Default::default(),
             return_data,
             parent_ctx,
-            child_return_data,
         }
     }
 
@@ -232,12 +230,12 @@ impl ExecutionContextImpl of ExecutionContextTrait {
 
     #[inline(always)]
     fn return_data(self: @ExecutionContext) -> Span<u8> {
-        self.return_data.span()
+        *self.return_data
     }
 
     /// Stops the current execution context.
     #[inline(always)]
-    fn stop(ref self: ExecutionContext) {
+    fn set_stopped(ref self: ExecutionContext) {
         self.status = Status::Stopped;
     }
 
@@ -247,9 +245,8 @@ impl ExecutionContextImpl of ExecutionContextTrait {
     /// (it is stopped) and contract creation and contract storage writes are
     /// reverted on its finalization.
     #[inline(always)]
-    fn revert(ref self: ExecutionContext, revert_reason: Span<u8>) {
+    fn set_reverted(ref self: ExecutionContext) {
         self.status = Status::Reverted;
-        ArrayExtensionTrait::concat(ref self.return_data, revert_reason);
     }
 
     // *************************************************************************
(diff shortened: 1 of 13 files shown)


crates/evm/src/context.cairo
```diff
@@ -301,11 +298,6 @@ impl ExecutionContextImpl of ExecutionContextTrait {
         'print debug'.print();
     }
 
-    #[inline(always)]
-    fn set_return_data(ref self: ExecutionContext, value: Array<u8>) {
```
Eikix, line 305: you probably still need this, 
when you call exec_return, and context is_root =>
then set_return_data as the correct slice in memory and exit the tx

crates/evm/src/machine.cairo
```diff
@@ -258,20 +259,23 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
         is_root
     }
 
+    /// Sets the `return_data` field of the parent context of the current
+    /// context to the `value` passed as parameter. If the current context is
+    /// the root context, does nothing.  Should be called upon returning from a
+    /// child context.
     #[inline(always)]
-    fn set_return_data(ref self: Machine, value: Array<u8>) {
```
Eikix, line 262: you probably still need this, 
when you call exec_return, and context is_root =>
then set_return_data as the correct slice in memory and exit the tx

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -32,9 +32,9 @@ impl SystemOperations of SystemOperationsTrait {
     fn exec_return(ref self: Machine) -> Result<(), EVMError> {
         let offset = self.stack.pop_usize()?;
         let size = self.stack.pop_usize()?;
-        let mut return_data = array![];
+        let mut return_data = Default::default();
         self.memory.load_n(size, ref return_data, offset);
-        self.set_return_data(return_data);
+        self.set_parent_return_data(return_data.span());
```
Eikix, line 37: I think we need a way to differentiate if it is the root context 
In which case, write return data

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -32,9 +32,9 @@ impl SystemOperations of SystemOperationsTrait {
...
         self.stop();
```
Eikix, line 38: why keep .stop for the machine, but not the ctx

enitrat: missed it

enitrat:
- Updated `set_return_data` to 

```rust
    /// Sets the `return_data` field of the appropriate execution context,
    /// taking into acount EVM specs: If the current context is the root
    /// context, sets the return_data field of the root context.  If the current
    /// context is a subcontext, sets the return_data field of the parent.
    /// Should be called when returning from a context.
```

- renamed machine method to `set_stopped`

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -32,10 +32,10 @@ impl SystemOperations of SystemOperationsTrait {
     fn exec_return(ref self: Machine) -> Result<(), EVMError> {
         let offset = self.stack.pop_usize()?;
         let size = self.stack.pop_usize()?;
-        let mut return_data = array![];
+        let mut return_data = Default::default();
         self.memory.load_n(size, ref return_data, offset);
-        self.set_return_data(return_data);
-        self.stop();
+        self.set_return_data(return_data.span());
+        self.stopped();
```
Eikix, line 38: I think you're trying to set_stopped

Maybe the exec_return test should make sure:
- the context is stopped
- the correct return data is written

enitrat: thanks

crates/evm/src/tests/test_utils.cairo
```diff
@@ -244,3 +226,35 @@ fn deploy_kakarot_core() -> IExtendedKakarotCoreDispatcher {
 
     IExtendedKakarotCoreDispatcher { contract_address }
 }
+
+
+// Simulate return of subcontext where
+/// 1. Set `return_data` field of parent context
+/// 2. make `parent_ctx` of `current_ctx` the current ctx
+fn return_from_subcontext(ref self: Machine, return_data: Span<u8>) {
```
Eikix, line 234: I wonder if we'll need this when we have the finalize_context method?
it could be used in the if (stopped) branch of the interpreter, and in the tests that need to go up one level?

this might be good though, to only target the behaviour we want,

in any case, i think its ok to have it rn, we just need to make sure it does exactly what is expected when a context is stopped

crates/evm/src/tests/test_instructions/test_system_operations.cairo
```diff
@@ -21,14 +23,15 @@ fn test_exec_return() {
     assert(machine.exec_return().is_ok(), 'Exec return failed');
 
     // Then
-    assert(1000 == load_word(32, machine.return_data()), 'Wrong return_data');
+    assert(1000 == load_word(32, parent_ctx_return_data(ref machine)), 'Wrong return_data');
```
Eikix, line 26: can we call machine.parent_ctx_return_data?

PR #425, opened 2023-10-13
feat: contract account bytecode storage

enitrat:
Currently implemented:
- feat: storage of bytecode (full bytes31 words)
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #423 

## What is the new behavior?

- #423 
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -335,8 +335,8 @@ fn pow2(pow: usize) -> u128 {
 /// Splits a u256 into `len` bytes, big-endian, and appends the result to `dst`.
 fn split_word(mut value: u256, mut len: usize, ref dst: Array<u8>) {
     let word_le = split_word_le(value, len);
-    let word_be = ArrayExtensionTrait::reverse(word_le.span());
-    ArrayExtensionTrait::concat(ref dst, word_be.span());
+    let word_be = ArrayExtTrait::reverse(word_le.span());
+    ArrayExtTrait::concat(ref dst, word_be.span());
 }
 
 fn split_u128_le(ref dest: Array<u8>, mut value: u128, mut len: usize) {
@@ -442,7 +442,7 @@ fn u256_to_bytes_array(mut value: u256) -> Array<u8> {
 }
 
 #[generate_trait]
-impl ArrayExtension<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayExtensionTrait<T> {
+impl ArrayExtension<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayExtTrait<T> {
     // Concatenates two arrays by adding the elements of arr2 to arr1.
     fn concat(ref self: Array<T>, mut arr2: Span<T>) {
         loop {
@@ -490,7 +490,7 @@ impl ArrayExtension<T, impl TCopy: Copy<T>, impl TDrop: Drop<T>> of ArrayExtensi
 }
 
 #[generate_trait]
-impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtensionTrait<T> {
+impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtTrait<T> {
     // Returns true if the array contains an item.
     fn contains<+PartialEq<T>>(mut self: Span<T>, value: T) -> bool {
         loop {
@@ -523,3 +523,19 @@ impl U256Impl of U256Trait {
     }
 }
 
+#[generate_trait]
+impl ByteArrayExt of ByteArrayExTrait {
+    fn from_bytes(mut bytes: Span<u8>) -> ByteArray {
+        //TODO(eni): optimize deserialization of Span<u8> to ByteArray;
+        // we can just deserialize bytes by chunks of 31, skipping pending_word
+        // checks
+        let mut arr: ByteArray = Default::default();
+        loop {
+            match bytes.pop_front() {
+                Option::Some(byte) => { arr.append_byte(*byte); },
+                Option::None => { break; }
+            }
+        };
+        arr
+    }
+}
--- a/crates/utils/src/storage.cairo
+++ b/crates/utils/src/storage.cairo
@@ -1,0 +1,31 @@
+use starknet::{
+    StorageBaseAddress, storage_base_address_from_felt252, Store, EthAddress, SyscallResult,
+    storage_write_syscall, storage_address_from_base, storage_read_syscall,
+    storage_address_from_base_and_offset
+};
+use poseidon::PoseidonTrait;
+use hash::{HashStateTrait, HashStateExTrait};
+
+
+/// Computes the StorageBaseAddress of a storage variable.  The address is
+/// computed by applying hashing the variable selector and the keys with
+/// Poseidon.
+///  # Arguments
+///  * `selector` - The selector of the storage variable.
+///  * `keys` - The keys of the storage variable.
+///  # Returns
+///  * The StorageBaseAddress of the storage variable, calculated as
+///    the sequential hashes of the selector and the keys.
+fn compute_storage_base_address(selector: felt252, mut keys: Span<felt252>) -> StorageBaseAddress {
+    //TODO: if we want compatibility with LegacyMaps, we should use pedersen
+    // it might not be required.
+    let mut state = PoseidonTrait::new().update(selector);
+    let hash = loop {
+        match keys.pop_front() {
+            Option::Some(val) => { state = state.update(*val); },
+            Option::None => { break state.finalize(); }
+        };
+    };
+
+    storage_base_address_from_felt252(hash)
+}
--- a/crates/utils/src/tests/test_helpers.cairo
+++ b/crates/utils/src/tests/test_helpers.cairo
@@ -1,7 +1,6 @@
 use utils::helpers;
-use utils::helpers::{
-    SpanExtension, SpanExtensionTrait, ArrayExtension, ArrayExtensionTrait, U256Trait
-};
+use utils::helpers::{SpanExtension, SpanExtTrait, ArrayExtension, ArrayExtTrait, U256Trait};
+use utils::helpers::{ByteArrayExTrait};
 use debug::PrintTrait;
 
 #[test]
@@ -200,4 +199,73 @@ fn test_split_u256_into_u64_little() {
     assert(low_h == 0xE5000000FFFFFFFA, 'split mismatch');
     assert(high_l == 0x0000450000DEFA00, 'split mismatch');
     assert(low_l == 0x00200400000000AD, 'split mismatch');
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_pack_bytes_le_bytes31() {
+    let mut arr = array![0x11, 0x22, 0x33, 0x44];
+    let res = ByteArrayExTrait::from_bytes(arr.span());
+
+    // Ensure that the result is complete and keeps the same order
+    let mut i = 0;
+    loop {
+        if i == arr.len() {
+            break;
+        };
+        assert(*arr[i] == res[i], 'byte mismatch');
+        i += 1;
+    };
+}
+
+#[test]
+#[available_gas(20000000)]
+fn test_pack_bytes_ge_bytes31() {
+    let mut arr = array![
+        0x01,
+        0x02,
+        0x03,
+        0x04,
+        0x05,
+        0x06,
+        0x07,
+        0x08,
+        0x09,
+        0x0A,
+        0x0B,
+        0x0C,
+        0x0D,
+        0x0E,
+        0x0F,
+        0x10,
+        0x11,
+        0x12,
+        0x13,
+        0x14,
+        0x15,
+        0x16,
+        0x17,
+        0x18,
+        0x19,
+        0x1A,
+        0x1B,
+        0x1C,
+        0x1D,
+        0x1E,
+        0x1F,
+        0x20,
+        0x21 // 33 elements
+    ];
+
+    let res = ByteArrayExTrait::from_bytes(arr.span());
+
+    // Ensure that the result is complete and keeps the same order
+    let mut i = 0;
+    loop {
+        if i == arr.len() {
+            break;
+        };
+        assert(*arr[i] == res[i], 'byte mismatch');
+        i += 1;
+    };
 }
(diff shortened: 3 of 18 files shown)


crates/core_contracts/src/contract_account.cairo
```diff
@@ -0,0 +1,86 @@
...
+            Option::Some(val) => { state = state.update(*val); },
+            Option::None => { break state.finalize(); }
+        };
+    };
+
+    storage_base_address_from_felt252(hash)
+}
+
+impl Storebytes31 of Store<bytes31> {
+    fn read(address_domain: u32, base: StorageBaseAddress) -> SyscallResult<bytes31> {
```
Eikix, line 55: what are those utils used for?

enitrat: read and write bytes31

crates/utils/src/helpers.cairo
```diff
@@ -523,3 +523,17 @@ impl U256Impl of U256Trait {
     }
 }
 
+fn pack_bytes(mut bytes: Span<u8>) -> ByteArray {
```
enitrat, line 526: #426

crates/utils/src/storage.cairo
```diff
@@ -0,0 +1,31 @@
...
+///  # Arguments
+///  * `selector` - The selector of the storage variable.
+///  * `keys` - The keys of the storage variable.
+///  # Returns
+///  * The StorageBaseAddress of the storage variable, calculated as
+///    the sequential hashes of the selector and the keys.
+fn compute_storage_base_address(selector: felt252, mut keys: Span<felt252>) -> StorageBaseAddress {
+    //TODO: if we want compatibility with LegacyMaps, we should use pedersen
+    // it might not be required.
+    let mut state = PoseidonTrait::new().update(selector);
```
enitrat, line 22: Using Poseidon makes more sense than Pedersen. We're likely not going to access the storage variables from the `Storage` structs anyway, so might as well go for the zk-optimized hash function

Eikix: how many steps do we save with poseidon compared to pedersen?
because in some sense, we can use the Unsafecontractstate trick in some opcodes to get the bytecode of a contract account 
at least in my PR it made code more clear

crates/core_contracts/Scarb.toml
```diff
@@ -7,6 +7,8 @@ version = "0.1.0"
 [dependencies]
 starknet.workspace = true
 evm = { path = "../evm" }
+utils = { path = "../utils" }
+alexandria_storage = { git = "https://github.com/keep-starknet-strange/alexandria.git", rev = "3356bf0c5c1a089167d7d3c28d543e195325e596" }
```
Eikix, line 11: it looks like we're importing the whole of alexandria 
why not name the package alexandria

crates/core_contracts/src/contract_account.cairo
```diff
@@ -0,0 +1,46 @@
+///! Module containing functions related to Contract Accounts
```
Eikix, line 1: this document comment feels a bit bad, 
- `Module containing` is already implied
- the name is contract_accounts.cairo

Could we have something like
```
Contract Account utils functions, including bytecode storage
```

crates/utils/src/helpers.cairo
```diff
@@ -523,3 +523,17 @@ impl U256Impl of U256Trait {
...
+    let mut byte_array: ByteArray = Default::default();
```
Eikix, line 527: this line is redundant afaik

crates/utils/src/tests/test_helpers.cairo
```diff
@@ -201,3 +201,72 @@ fn test_split_u256_into_u64_little() {
...
+
+    let res = helpers::pack_bytes(arr.span());
+
+    // Ensure that the result is complete and keeps the same order
+    let mut i = 0;
+    loop {
+        if i == arr.len() {
+            break;
+        };
+        assert(*arr[i] == res[i], 'byte mismatch');
```
Eikix, line 269: i'm not sure I understand this test
the type ByteArray is a byte optimized (byte packing) type, but can be accessed normally by indexed?

so theoretically, res is a ByteArray, of "length" 2 -> two bytes31 
but you can access it with 0 < index < 33?

crates/utils/src/traits.cairo
```diff
@@ -96,3 +96,38 @@ impl U256TryIntoResultU32 of TryIntoResult<u256, usize> {
         }
     }
 }
+
+
+impl StoreBytes31 of Store<bytes31> {
```
Eikix, line 101: where is this used? I still don't understand

crates/core_contracts/src/contract_account.cairo
```diff
@@ -0,0 +1,46 @@
...
+    let data_address = compute_storage_base_address(
+        selector!("contract_account_bytecode"), array![evm_address.into()].span()
+    );
+    // We start storing the full 31-byte words of bytecode data at address
+    // `data_address`.  The `pending_word` and `pending_word_len` are stored at
+    // address `data_address-2` and `data_address-1` respectively.
+    let mut stored_list: List<bytes31> = List {
+        address_domain: 0, base: data_address, len: 0, storage_size: 31
+    };
+    let pending_word_addr: felt252 = data_address.into() - 2;
```
Eikix, line 31: why did we choose -2 as a place to store the pending word?

crates/core_contracts/src/contract_account.cairo
```diff
@@ -0,0 +1,46 @@
...
+    let pending_word_len_addr: felt252 = pending_word_addr + 1;
```
Eikix, line 32: why did we choose -1 as a place to store the pending word len?

crates/utils/src/traits.cairo
Eikix: erratum: https://github.com/keep-starknet-strange/alexandria/blob/3356bf0c5c1a089167d7d3c28d543e195325e596/src/storage/src/list.cairo#L15

ListTrait asks for impl TStore: Store<T>> of ListTrait<T>

crates/utils/src/storage.cairo
enitrat: > because in some sense, we can use the Unsafecontractstate trick in some opcodes to get the bytecode of a contract account

didn't understand

crates/core_contracts/Scarb.toml
enitrat: this is just importing the `alexandria_storage` crate, because alexandria's root is a workspace

crates/core_contracts/src/contract_account.cairo
enitrat: We want the StorageBaseAddress computed to only be used for the list, because the internal calculations consider that from the base offset you can store up to 256 values of size one.
 If we started by storing `pending_word` and `pending_word_len` at offsets 0 and 1, we would encounter clashes when instantiating the `List`

 This leaves us two choices:
 - Compute a new hash and store `pending_word{_len}` at some other place in storage
 - Just use the two previous addresses value to store `pending_word{_len}`. This saves us unnecessary hashes.
 
  If we use `-1` and `-2` for the pending word and the pending word_len,

crates/utils/src/tests/test_helpers.cairo
enitrat: res is a ByteArray containing 33 bytes that can be accessed by index - internally, they're split into one full-packed 31-byte word and one 2 byte word

crates/utils/src/storage.cairo
enitrat: > how many steps do we save with poseidon compared to pedersen?

```rust
❯ scarb test
     Running cairo-test hashes
testing hashes ...
running 2 tests
test hashes::tests::test_poseidon ... ok (gas usage est.: 26550)
test hashes::tests::test_pedersen ... ok (gas usage est.: 40800)
test result: ok. 2 passed; 0 failed; 0 ignored; 0 filtered out;
```

~50% when hashing 4 values

crates/core_contracts/Scarb.toml
Eikix: interesting i didn't know naming a certain import was a way to only import a certain crate

crates/core_contracts/src/contract_account.cairo
Eikix: > This leaves us two choices:

I agree on this finding, my understanding is that starknet has chosen as an acceptable risk for storing elements of size up to 256 to offset the base address with 256, 

so we have [base_address, base_address + 255] as an interval,
are we kinda breaking this design choice by extending slightly this interval for our specific usecase to [base_interval - 2, base_address + 255]? 

Instinctively, the collision risk has not become that much greater than before, but still

My understanding of the tradeoff:
- another hash ==> more gas
- -2 offset => marginally more risk of collision

crates/core_contracts/src/contract_account.cairo
```diff
@@ -0,0 +1,47 @@
...
+use starknet::{
+    StorageBaseAddress, storage_base_address_from_felt252, Store, EthAddress, SyscallResult,
+    storage_write_syscall, storage_address_from_base, storage_read_syscall,
+    storage_address_from_base_and_offset
+};
+use utils::helpers::pack_bytes;
+use utils::storage::{compute_storage_base_address};
+use utils::traits::{StorageBaseAddressIntoFelt252, StoreBytes31};
+
+/// Stores the bytecode of a contract account.  The bytecode is first packed
```
Eikix, line 15:
```suggestion
/// Stores the EVM bytecode of a contract account in Kakarot Core's contract storage.  The bytecode is first packed
```

crates/core_contracts/src/contract_account.cairo
```diff
@@ -0,0 +1,47 @@
...
+    // We start storing the full 31-byte words of bytecode data at address
+    // `data_address`.  The `pending_word` and `pending_word_len` are stored at
+    // address `data_address-2` and `data_address-1` respectively.
+    let mut stored_list: List<bytes31> = List {
+        address_domain: 0, base: data_address, len: 0, storage_size: 31
+    };
+    let pending_word_addr: felt252 = data_address.into() - 2;
+    let pending_word_len_addr: felt252 = pending_word_addr + 1;
+
+    // Store the `ByteArray` in the contract storage.
```
Eikix, line 35: is this an accurate comment?

crates/utils/src/tests/test_helpers.cairo
Eikix: ahh yes ok, in Cairo you can overload the index view : []
```
impl ByteArrayIndexView of IndexView<ByteArray, usize, u8> {
    fn index(self: @ByteArray, index: usize) -> u8 {
        self.at(index).expect('Index out of bounds')
    }
}
```

Eikix: note to self: we should be careful in using indexview as it can panic

crates/core_contracts/src/contract_account.cairo
```diff
@@ -0,0 +1,47 @@
...
+use utils::storage::{compute_storage_base_address};
+use utils::traits::{StorageBaseAddressIntoFelt252, StoreBytes31};
+
+/// Stores the bytecode of a contract account.  The bytecode is first packed
+/// into a ByteArray and then stored in the contract storage.
+/// # Arguments
+/// * `evm_address` - The EVM address of the contract account
+/// * `bytecode` - The bytecode to store
+fn store_bytecode(evm_address: EthAddress, bytecode: Span<u8>) {
+    let packed_bytecode: ByteArray = pack_bytes(bytecode);
```
Eikix, line 21: I wonder if we should implement pack_bytes as a trait on Span<u8>? to write
bytecode.pack_bytes?

crates/core_contracts/src/contract_account.cairo
enitrat: > -2 offset => marginally more risk of collision

This doesn't increase the risk of collision at all since you just remove 2 unused values from the unused address set. Whether these addresses are calculated using hashes or randomly doesn't matter.

crates/core_contracts/src/contract_account.cairo
enitrat: We could, but we could also leave it as is - I don't mind. I don't see a lot of benefits of having an impl with a single function as they're normally used to pack logic together.

crates/core_contracts/src/contract_account.cairo
enitrat: yes

crates/utils/src/helpers.cairo
```diff
@@ -523,3 +523,16 @@ impl U256Impl of U256Trait {
     }
 }
 
+fn pack_bytes(mut bytes: Span<u8>) -> ByteArray {
```
Eikix, line 526: I think it'd be clean to implement this as a ByteArray::From function, wdyt

crates/utils/src/storage.cairo
```diff
@@ -0,0 +1,31 @@
...
+/// Computes the StorageBaseAddress of a storage variable.  The address is
+/// computed by applying hashing the variable selector and the keys with
+/// Poseidon.
+///  # Arguments
+///  * `selector` - The selector of the storage variable.
+///  * `keys` - The keys of the storage variable.
+///  # Returns
+///  * The StorageBaseAddress of the storage variable, calculated as
+///    the sequential hashes of the selector and the keys.
+fn compute_storage_base_address(selector: felt252, mut keys: Span<felt252>) -> StorageBaseAddress {
```
Eikix, line 19: how cool would it be if we could type keys as any type that implement the trait hash?

Eikix: because if we type it T, +Hash<T>, it forces us to use the same key type?

enitrat: yeah, can do, or we serialize the keys before. Let's do yours

Eikix:
Good morrow, kind sir! Pray forgive the entanglements yon recent merge hath wrought upon thy most noble PR. Verily, the fault is mine, for thy pull request was as splendid as a midsummer's day! Yet, in this digital Globe Theatre of codes and commits, sometimes our scripts do cross and tangle. Fear not, for together we shall untwine the knot and let thy code shine once more! Huzzah to thee and thy patience, good sire! 🎭📜

enitrat:
> Good morrow, kind sir! Pray forgive the entanglements yon recent merge hath wrought upon thy most noble PR. Verily, the fault is mine, for thy pull request was as splendid as a midsummer's day! Yet, in this digital Globe Theatre of codes and commits, sometimes our scripts do cross and tangle. Fear not, for together we shall untwine the knot and let thy code shine once more! Huzzah to thee and thy patience, good sire! 🎭📜


Aye aye, matey! I be makin' the necessary changes to merge this here pull request without a hitch! Yarrr!

Eikix (approved):
lgtm

PR #430, opened 2023-10-17
feat: execute

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #424

## What is the new behavior?

- Remove `starknet_address` field from the context
- Implements the `execute` function
- #424 
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -26,15 +26,13 @@ mod KakarotCore {
     use core::pedersen::{HashState, PedersenTrait};
     use core::starknet::SyscallResultTrait;
     use core::zeroable::Zeroable;
-    use core_contracts::components::ownable::ownable_component::InternalTrait;
-    use core_contracts::components::ownable::{ownable_component};
     use evm::errors::EVMError;
     use evm::execution::execute;
-    use evm::storage::ContractAccountStorage;
+    use evm::context::Status;
+    use super::ContractAccountStorage;
     use starknet::{
         EthAddress, ContractAddress, ClassHash, get_tx_info, get_contract_address, deploy_syscall
     };
-    use super::ContractAccountStorage;
     use super::INVOKE_ETH_CALL_FORBIDDEN;
     use utils::constants::{CONTRACT_ADDRESS_PREFIX, MAX_ADDRESS};
     use utils::traits::U256TryIntoContractAddress;
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -1,7 +1,6 @@
 use starknet::{ContractAddress, EthAddress, ClassHash};
 
 const INVOKE_ETH_CALL_FORBIDDEN: felt252 = 'KKT: Cannot invoke eth_call';
-const INVOKE_ETH_CALL_ERROR: felt252 = 'KKT: Cannot invoke eth_call';
 
 
 #[derive(Copy, Drop, Serde, starknet::Store)]
@@ -26,13 +25,13 @@ mod KakarotCore {
     use core::pedersen::{HashState, PedersenTrait};
     use core::starknet::SyscallResultTrait;
     use core::zeroable::Zeroable;
+    use evm::context::Status;
     use evm::errors::EVMError;
     use evm::execution::execute;
-    use evm::context::Status;
-    use super::ContractAccountStorage;
     use starknet::{
         EthAddress, ContractAddress, ClassHash, get_tx_info, get_contract_address, deploy_syscall
     };
+    use super::ContractAccountStorage;
     use super::INVOKE_ETH_CALL_FORBIDDEN;
     use utils::constants::{CONTRACT_ADDRESS_PREFIX, MAX_ADDRESS};
     use utils::traits::U256TryIntoContractAddress;
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -1,6 +1,7 @@
 use starknet::{ContractAddress, EthAddress, ClassHash};
 
 const INVOKE_ETH_CALL_FORBIDDEN: felt252 = 'KKT: Cannot invoke eth_call';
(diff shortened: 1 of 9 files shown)


crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -267,6 +269,27 @@ mod KakarotCore {
...
             Result::Ok(())
         }
+
+
+        /// Deploys a contract account for a particular EVM address
+        fn deploy_contract_account(
+            ref self: ContractState, evm_address: EthAddress, value: u256, bytecode: Span<u8>
+        ) -> bool {
+            let (ret_status, ret_data) = execute(
+                :evm_address,
```
Eikix, line 279: why :evm_address?

crates/evm/src/execution.cairo
```diff
@@ -1,16 +1,43 @@
...
+        id: 0, :evm_address, :call_ctx, parent_ctx: Default::default(), return_data: array![].span()
+    );
+    let mut machine: Machine = MachineCurrentContextTrait::new(ctx);
 
     let mut interpreter = EVMInterpreterTrait::new();
-    // Execute the transaction.
-    interpreter.run(ref machine)
+    // Execute the bytecode.
+    interpreter.run(ref machine);
+    (machine.status(), machine.return_data())
```
Eikix, line 41: I think in Cairo0 we return a summary of the stack and memory as well, it seems like its the behaviour in evm.codes as well?
How about EELS? https://github.com/ethereum/execution-specs/tree/master/src/ethereum/shanghai/vm

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -267,6 +269,27 @@ mod KakarotCore {
             }
             Result::Ok(())
         }
+
+
+        /// Deploys a contract account for a particular EVM address
+        fn deploy_contract_account(
```
Eikix, line 275: From what I understand, the responsibility to compute the evm_address is left to the caller of this function (handled upstream)
source:
https://github.com/ethereum/execution-specs/blob/0fdcaa003910127165e59f46e781349fe192d019/src/ethereum/shanghai/utils/message.py#L78C1-L78C1

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -267,6 +269,27 @@ mod KakarotCore {
...
+            ref self: ContractState, evm_address: EthAddress, value: u256, bytecode: Span<u8>
+        ) -> bool {
+            let (ret_status, ret_data) = execute(
```
Eikix, line 278: Are we checking enough invariants before entering this context?
https://github.com/ethereum/execution-specs/blob/master/src/ethereum/shanghai/vm/interpreter.py#L147

Or should we do those upstream the call of this function

crates/evm/src/execution.cairo
```diff
@@ -1,16 +1,43 @@
...
+    gas_price: u64,
+    gas_limit: u64,
+) -> (Status, Span<u8>) {
+    let call_ctx = CallContextTrait::new(
+        caller: evm_address, :bytecode, :calldata, :value, read_only: false, :gas_limit, :gas_price
+    );
     // Create new execution context.
-    let mut machine: Machine = Default::default();
+    let ctx = ExecutionContextTrait::new(
+        id: 0, :evm_address, :call_ctx, parent_ctx: Default::default(), return_data: array![].span()
```
Eikix, line 34:
```suggestion
        id: 0, :evm_address, :call_ctx, parent_ctx: Default::default(), return_data: Default::default()
```

crates/contracts/src/kakarot_core/kakarot.cairo
enitrat: named parameter

crates/evm/src/execution.cairo
enitrat: > we return a summary of the stack and memory as well,

What for? Isn't it just the squashed data structures? I don't see where we would do that. When the execution is over, why would we access the stack and the memory? We just need `return_data`

> it seems like its the behaviour in evm.codes as well?

Can you point me to this?

> How about EELS? https://github.com/ethereum/execution-specs/tree/master/src/ethereum/shanghai/vm

What about it? 😅

crates/contracts/src/kakarot_core/kakarot.cairo
enitrat: This is my mistake - it shouldn't be a part of this PR. 
I will delete this and take into accounts comments for the next PR

crates/evm/src/execution.cairo
Eikix: I think we may have to return self-destruct contract list, contract-creation list, as well as event array? that way we separate responsibility and allow another function to handle the finalization of those

crates/evm/src/execution.cairo
```diff
@@ -16,28 +16,44 @@ use starknet::{EthAddress, ContractAddress};
...
+
+    // Initiate the Machine with the root context
     let mut machine: Machine = MachineCurrentContextTrait::new(ctx);
 
     let mut interpreter = EVMInterpreterTrait::new();
     // Execute the bytecode.
     interpreter.run(ref machine);
-    (machine.status(), machine.return_data())
+    ExecutionResult {
+        status: machine.status(),
```
Eikix, line 52: it's not really a problem but let me think out loud here:
status can be reverted, stopped or running,

in this particular line, it can only be either stopped or reverted,

which amounts to being a success field, 

do we need the outer context to "match" this inner type to figure out what to do? or must we transform it here into a `success` field?

Eikix (commented):
Can we test this new function?

enitrat:
> Can we test this new function?

Should we? I think it should be part of a bigger testing strategy. Or should I just send an empty bytecode?
Not sure what can be unit tested here.

Eikix:
> > Can we test this new function?
> 
> Should we? I think it should be part of a bigger testing strategy. Or should I just send an empty bytecode? Not sure what can be unit tested here.

I see, may make more sense to test interpreter.run then

Eikix (approved):
lgtm

Issue #426, opened 2023-10-16
dev: optimize `pack_bytes` by skipping checks for initial `n` chunks with more than 31 bytes


enitrat:
https://github.com/kkrt-labs/kakarot-ssj/pull/425/files#diff-8ab8f65f6e1ee9e453913c37b065a35340c0de8ac765d0649c53f5dc83d200d5R526-R539

Refactor the `pack_bytes` function and optimize deserialization of Span<u8> to ByteArray. Instead of calling `append_byte` byte by byte, we can simply pack the full 31-bytes chunks ourselves and append them directly to the `data` field. 
For the last incomplete chunk, we know that it's length is `len() % 31`. we can pack these together and modify the `pending_word` and `pending_word_len` fields directly.

chachaleo:
Can I take it ?

enitrat:
it's now under BytesArrayExTrait - `from_bytes`

PR #432, opened 2023-10-17
dev: refactor from_byte

chachaleo:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #426

## What is the new behavior?

optimize deserialization of Span to ByteArray

### DIFF
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -529,7 +529,7 @@ impl U256Impl of U256Trait {
 impl ByteArrayExt of ByteArrayExTrait {
     fn from_bytes(mut bytes: Span<u8>) -> ByteArray {
         let mut arr: ByteArray = Default::default();
-        let (mut nb_full_words, mut pending_word_len) = DivRem::div_rem(
+        let (nb_full_words, pending_word_len) = DivRem::div_rem(
             bytes.len(), 31_u32.try_into().unwrap()
         );
         let mut i = 0;
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -7,6 +7,8 @@ use utils::constants::{
     POW_256_16,
 };
 use keccak::u128_split;
+use traits::DivRem;
+use integer::U32TryIntoNonZero;
 
 
 /// Ceils a number of bits to the next word (32 bytes)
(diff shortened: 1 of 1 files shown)


crates/utils/src/helpers.cairo
```diff
@@ -530,12 +532,39 @@ impl ByteArrayExt of ByteArrayExTrait {
         // we can just deserialize bytes by chunks of 31, skipping pending_word
         // checks
         let mut arr: ByteArray = Default::default();
+        let (mut q, mut r) = DivRem::div_rem(bytes.len(), 31_u32.try_into().unwrap());
```
enitrat, line 535:
```suggestion
        let (mut nb_full_words, mut bytes_last_word) = DivRem::div_rem(bytes.len(), 31_u32.try_into().unwrap());
```

crates/utils/src/helpers.cairo
```diff
@@ -530,12 +532,39 @@ impl ByteArrayExt of ByteArrayExTrait {
...
         loop {
-            match bytes.pop_front() {
-                Option::Some(byte) => { arr.append_byte(*byte); },
-                Option::None => { break; }
-            }
+            if q == 0 {
```
enitrat, line 537: I prefer that you loop from `i=0` to `nb_full_words` rather than mutating this value. If you apply this change, change the internal counter name to `j`

crates/utils/src/helpers.cairo
```diff
@@ -530,12 +532,39 @@ impl ByteArrayExt of ByteArrayExTrait {
...
+
+        if r == 0 {
+            return arr;
+        };
+
+        arr.pending_word_len = r;
+        let mut pending_word: felt252 = 0;
+
+        loop {
+            if r == 0 {
```
enitrat, line 561: same as above.

Eikix:
Damn!!!!!!! 
Nice first PR! Welcome;D

crates/utils/src/helpers.cairo
```diff
@@ -526,16 +528,44 @@ impl U256Impl of U256Trait {
 #[generate_trait]
 impl ByteArrayExt of ByteArrayExTrait {
     fn from_bytes(mut bytes: Span<u8>) -> ByteArray {
-        //TODO(eni): optimize deserialization of Span<u8> to ByteArray;
-        // we can just deserialize bytes by chunks of 31, skipping pending_word
-        // checks
         let mut arr: ByteArray = Default::default();
+        let (mut nb_full_words, mut pending_word_len) = DivRem::div_rem(
```
enitrat, line 532: one last thing: since we don't mutate these variables anymore, we can remove the `mut `keyword

```suggestion
        let (mut nb_full_words, mut pending_word_len) = DivRem::div_rem(
```

enitrat:
fyi: I think the word_not_full test has higher consumption, because the price of the divrem probably outweighs the cost of simply running the `append_bytes` fn three times.

It's interesting to see how this cost shifts when the number of bytes goes up. In real-world scenario, the bytecode will be a lot heavier, so it's going to save a lot!

enitrat (approved):
LGTM 💯 
image

Issue #391, opened 2023-10-04
feat: create a `generic_call` util to abstract the logic of the call opcodes


Eikix:
## Feature
- Implement a util function `generic_call` that abstracts the logic of call opcode

Reference:
- https://github.com/ethereum/execution-specs/blob/8b8c7c1a7437510f614272392275855de26209d3/src/ethereum/berlin/vm/instructions/system.py#L309

- https://github.com/kkrt-labs/kakarot/blob/main/src/kakarot/instructions/system_operations.cairo

PR #441, opened 2023-10-19
feat: exec call

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #391 #392 

- [ ] Yes
- [ ] No


Eikix:
Still need to implement tests

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,131 @@
...
+        // Case 1: `to` address is a precompile
+        // Handle precompile logic
+        if is_precompile(call_args.to) {
+            return Result::Err(EVMError::NotImplemented);
+        }
+
+        // Case 2: `to` address is not a precompile
+        // We enter the standard flow
+        // TODO(elias): be able to fetch bytecode
+        let bytecode = array![].span();
```
enitrat, line 70: no strong opinion on this

```suggestion
        let bytecode = Default::default().span();
```

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,131 @@
...
+        self.current_ctx = BoxTrait::new(child_ctx);
+
+        Result::Ok(())
+    }
+}
+
+/// Check whether a `to` address for a call-family opcode is a precompile
+/// Since range check is expensive in Cairo, we proceed with checking equality
+fn is_precompile(to: EthAddress) -> bool {
+    if to.into() == 0x01_felt252 {
```
enitrat, line 103: is it better by checking that `to.into() <= 9_felt252` ?

crates/utils/src/traits.cairo
```diff
@@ -54,6 +54,16 @@ impl U256TryIntoContractAddress of TryInto<u256, ContractAddress> {
     }
 }
 
+impl U256TryIntoEthAddress of TryInto<u256, EthAddress> {
```
enitrat, line 57: used where?

crates/evm/src/bytecode.cairo
```diff
@@ -0,0 +1,23 @@
...
+    let eoa_starknet_address = kakarot_state.eoa_starknet_address(evm_address);
+
+    // Case 1: EOA is deployed
+    if !eoa_starknet_address.is_zero() {
+        return Result::Ok(Default::default().span());
+    }
+
+    // Case 2: EOA is not deployed and CA is deployed
+    let ca = ContractAccountTrait::new(evm_address);
+    return Result::Ok(Default::default().span());
```
enitrat, line 21: you can use ca.load_bytecode() here

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,156 @@
...
+        let value = if with_value {
+            self.stack.pop()?
+        } else {
+            0
+        };
+
+        let args_offset = self.stack.pop_usize()?;
+        let args_size = self.stack.pop_usize()?;
+
+        let output_offset = self.stack.pop_usize()?;
```
enitrat, line 49: for consistency I with evm.codes I prefer `ret_offset` and `ret_size`

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,156 @@
...
+
+    /// Finalize the calling context by:
+    /// - Pushing the execution status to the Stack
+    /// - Set the return data of the parent context
+    /// - Store the return data in Memory
+    /// - Return to parent context and decrease the ctx_count.
+    fn finalize_calling_context(ref self: Machine) -> () {
+        // Put the status of the call on the stack.
+        let status = self.status();
+        let status = match status {
```
enitrat, line 117:
```suggestion
        let success = match status {
```

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,156 @@
...
+            Status::Active => 1,
```
enitrat, line 118: Just out of curiosity but can this ever happen?

crates/evm/src/machine.cairo
```diff
@@ -129,6 +144,23 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
         call_ctx
     }
 
+    /// Returns from the sub context by setting the current context
+    /// to the parent context.
+    #[inline(always)]
+    fn return_to_parent_ctx(ref self: Machine) {
+        let mut current_ctx = self.current_ctx.unbox();
+        let maybe_parent_ctx = current_ctx.parent_ctx;
+        let maybe_parent_ctx = match match_nullable(maybe_parent_ctx) {
```
enitrat, line 153:
```suggestion
        match match_nullable(maybe_parent_ctx) {
```

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -48,7 +59,29 @@ impl SystemOperations of SystemOperationsTrait {
...
     /// # Specification: https://www.evm.codes/#f1?fork=shanghai
     fn exec_call(ref self: Machine) -> Result<(), EVMError> {
-        Result::Err(EVMError::NotImplemented)
+        // Prepare the call arguments with with_value = true for a call.
+        let call_args = self.prepare_call(true)?;
+        let read_only = self.read_only();
+        let value = call_args.value;
+
+        // Check if current context is read only that value == 0.
+        if read_only & (value != 0) {
```
enitrat, line 68: short circuiting AND 
```suggestion
        if read_only && (value != 0) {
```

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -48,7 +59,29 @@ impl SystemOperations of SystemOperationsTrait {
...
+        // 0 on the stack to indicate call failure.
+        let caller_address = self.evm_address();
+        let sender_balance = balance(caller_address)?;
+        if sender_balance < value {
+            self.stack.push(0);
+            return Result::Ok(());
+        }
+
+        // Initialize the sub context.
+        self.init_sub_call_ctx(call_args, read_only,);
```
enitrat, line 82:
```suggestion
        self.init_sub_call_ctx(call_args, read_only,)
```

Propagate eventual error by returning the output of this function call - delete line below

crates/evm/src/interpreter.cairo
```diff
@@ -67,7 +72,7 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
     fn decode_and_execute(ref self: EVMInterpreter, ref machine: Machine) -> Result<(), EVMError> {
         // Retrieve the current program counter.
         let pc = machine.pc();
-        let bytecode = machine.call_ctx().bytecode();
+        let bytecode = CallContextTrait::bytecode(@machine.call_ctx());
```
enitrat, line 75: was it not working?

crates/evm/src/context.cairo
```diff
@@ -141,9 +141,19 @@ struct ExecutionContext {
     create_addresses: Array<EthAddress>,
     // Return data of a child context.
     return_data: Span<u8>,
+    // Output of the current context.
+    output: Span<u8>,
     parent_ctx: Nullable<ExecutionContext>,
 }
 
+#[derive(Drop, Default, Copy, PartialEq)]
+enum ExecutionContextId {
```
enitrat, line 150: can you add code doc

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,131 @@
...
+    /// The Machine will change its `current_ctx` to point to the
+    /// newly created sub-context.
+    /// Then, the EVM execution loop will start on this new execution context.
+    fn init_sub_ctx(ref self: Machine, with_value: bool, read_only: bool) -> Result<(), EVMError> {
+        let call_args = self.prepare_call(with_value)?;
+
+        // Case 1: `to` address is a precompile
+        // Handle precompile logic
+        if is_precompile(call_args.to) {
+            return Result::Err(EVMError::NotImplemented);
```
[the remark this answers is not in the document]
Eikix: ah yes sounds better

crates/evm/src/call_helpers.cairo
Eikix: idk why, i think the reason was that i tried writing:
```
let bytecode: Span<u8>  = Default::default();
```
and it didnt work

crates/evm/src/call_helpers.cairo
Eikix: in terms of gas?

crates/utils/src/traits.cairo
Eikix: actually i think not used, I initially needed to convert stack item to evm_address, but I ended using `pop_eth_address`

crates/evm/src/context.cairo
```diff
@@ -141,9 +141,19 @@ struct ExecutionContext {
...
+    #[default]
+    Root: usize,
```
Eikix, line 152: root `must` be 0, is there a way to enforce this with the type system?

in rust you'd do enum {
Root = 0,
} afaik

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -30,11 +35,17 @@ impl SystemOperations of SystemOperationsTrait {
...
         let size = self.stack.pop_usize()?;
+
+        // Load the data from memory
         let mut return_data = Default::default();
         self.memory.load_n(size, ref return_data, offset);
-        self.set_return_data(return_data.span());
+
+        // Set the memory data to the context output
+        // and halt the context.
+        self.set_output(return_data.span());
```
Eikix, line 48: afaik we don't need this

Eikix: we don't need the output field iirc

Eikix: for the entire lifetime of the execution context, the return_data field holds the child's return data (enabling RETURNDATASIZE and RETURNDATACOPY)

when return is called, we know nothing can be done afterwards,
we overwrite the return data with the current ctx return data,

later on, during finalization, this return data will be copied to his parent's return data field

crates/evm/src/context.cairo
```diff
@@ -34,29 +34,17 @@ struct CallContext {
...
-    fn value(self: @CallContext) -> u256;
-    fn read_only(self: @CallContext) -> bool;
-    fn gas_limit(self: @CallContext) -> u64;
-    fn gas_price(self: @CallContext) -> u64;
+    // Evm gas limit for the call
+    gas_limit: u128,
+    // Evm gas price for the call
+    gas_price: u128,
+    // The offset in memory to store the context output
+    output_offset: usize,
```
Eikix, line 42: FYI in KKT0 we don't need output_offset and output_size to live in a called context's callcontext as we use a trick:
- when CALL* family opcode is invoked, output_offset and output_size are pushed onto the stack
- we leave them on the stack on purpose (pop less items than CALL no. of arguments)
- when we finalize, we know they're the stack top of the parent context of the called context we're finalizing, so we're able to grab them.
- we perform the correct behaviour of the RETURN opcode

Eikix: For clarity and ease of reading/refactoring, I'm okay with leaving this as is
@ClementWalter wdyt

crates/evm/src/call_helpers.cairo
enitrat: `let bytecode: Span<u8>  = Default::default().span();`

crates/evm/src/call_helpers.cairo
enitrat: ye

crates/utils/src/traits.cairo
enitrat: ok, lets remove it

crates/evm/src/call_helpers.cairo
greged93: I think it can, actually that might be something that we need to check because not sure this will work on our VM currently. 

Tried it with anvil and cast if you want to try, this code seems to run:
- `anvil`
- `cast rpc anvil_setCode 0x00000000000000000000000000000000deadbeef 0x604260005260206000` sets the bytecode to PUSH1 42 PUSH1 0 MSTORE PUSH1 20 PUSH1 0, meaning there is no return and no stop.
- `cast send 0x00000000000000000000000000000000deadbeef --value 1 --private-key any_private_key_from_anvil`
- Should return status `1`
- You can also check that the tx ran with `cast run tx_hash`

Eikix: so if i understand correctly, having a pc that reaches the end of the byte code amounts to a stop??

crates/evm/src/context.cairo
greged93: tried a couple of combinations but I don't think you can inforce this

crates/evm/src/interpreter.cairo
greged93: it gave me an error, but for some reason now not, fixed.

crates/evm/src/context.cairo
greged93: I'm not 100% sure, maybe @ClementWalter can confirm, but we are able to do this on KKRT0 because each context has his own stack. When we init a subcontext, it comes with an empty stack that only it can use. However, in SSJ, there is not separation of Stack, just a global stack for all contexts. Meaning if we do this and leave elements on the stack, we could have the subcontext calling `pop` on it and removing these return offset et return size.

crates/evm/src/call_helpers.cairo
greged93: Exactly that yes
Screenshot 2023-10-20 at 10 29 10

crates/evm/src/context.cairo
greged93: Ok it seems we also use separate stacks for each context in ssj, so we could implement the same as kkrt0, @enitrat and @ClementWalter your thoughts?

crates/evm/src/call_helpers.cairo
greged93: https://github.com/kkrt-labs/kakarot-ssj/issues/446

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,153 @@
...
+        // The difference being the "value" parameter in CALL and CALLCODE.
+        let gas = self.stack.pop_u128()?;
+        let to = self.stack.pop_eth_address()?;
+
+        // CALL and CALLCODE expect value to be on the stack
+        // for STATICCALL and DELEGATECALL, the value is the calling call context's value
+        let value = if with_value {
+            self.stack.pop()?
+        } else {
+            0
```
Eikix, line 43: I guess we could technically put here:
`self.value()`, i.e. the parent context's value

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,153 @@
...
+        // We enter the standard flow
+        let bytecode = bytecode(call_args.to)?;
+        // The caller in the subcontext is the current context's current address
+        let caller = self.evm_address();
+
+        let call_ctx = CallContextTrait::new(
+            caller,
+            bytecode,
+            call_args.calldata,
+            call_args.value,
```
Eikix, line 81: linking https://github.com/kkrt-labs/kakarot-ssj/pull/441/files#r1368066868

crates/evm/src/call_helpers.cairo
Eikix: Should we instead -> have the PC reach the last index -> set status to stopped, 
so this case should never happen?

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,153 @@
...
+    /// - Return to parent context and decrease the ctx_count.
+    fn finalize_calling_context(ref self: Machine) -> () {
+        // Put the status of the call on the stack.
+        let status = self.status();
+        let success = match status {
+            Status::Active => 1,
+            Status::Stopped => 1,
+            Status::Reverted => 0,
+        };
+        self.stack.push(success);
```
Eikix, line 120:
```suggestion
        self.stack.push(success)?;
```

Eikix: Pushing an element to the stack is a Result

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,153 @@
...
+        let return_data = return_data.slice(0, return_data_len);
+        self.memory.store_n(return_data, call_ctx.ret_offset);
+
+        // Return from the current sub ctx by setting the execution context
+        // to the parent context.
+        self.return_to_parent_ctx();
+
+        // Machine logic
+        // Decrement the total context count
+        self.ctx_count -= 1;
```
Eikix, line 140: The context count should never be decreased afaik,
we use the context count to choose the ID of a sub context,
Each context ID should be unique, otherwise their stack and memory might collide.

We can't have
Context 1 -> Subcalls -> Context 2 -> Context 2 finishes -> Return to Context 1 -> Subcalls -> creates a new context "context 2"

This new Id will make it so that dicts will have lingering data.

Eikix: Maybe the name of the variable does not reflect this properly -> it's a overall context count

crates/evm/src/context.cairo
Eikix: How about just
```
Root,
Call: usize,
Create: usize,
```

crates/evm/src/interpreter.cairo
```diff
@@ -50,6 +51,10 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
                         machine.storage_journal.finalize_local();
                         if machine.is_root() {
                             machine.storage_journal.finalize_global();
+                        } else if machine.is_call() {
```
Eikix, line 54: how about 

```
match self.current_ctx().execution_id() {
...
}
```

crates/evm/src/machine.cairo
```diff
@@ -129,6 +129,23 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
...
     }
 
+    /// Returns from the sub context by setting the current context
+    /// to the parent context.
+    #[inline(always)]
+    fn return_to_parent_ctx(ref self: Machine) {
+        let mut current_ctx = self.current_ctx.unbox();
+        let maybe_parent_ctx = current_ctx.parent_ctx;
+        match match_nullable(maybe_parent_ctx) {
+            FromNullableResult::Null => current_ctx = Default::default(),
```
Eikix, line 139: Why do we have to return Default::default() from this match arm?

crates/evm/src/machine.cairo
```diff
@@ -129,6 +129,23 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
...
+            FromNullableResult::NotNull(parent_ctx) => {
+                let parent_ctx = parent_ctx.unbox();
+                current_ctx.parent_ctx = Default::default();
```
Eikix, line 142: Why do we need to do this?

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,153 @@
...
+        let success = match status {
+            Status::Active => 1,
+            Status::Stopped => 1,
+            Status::Reverted => 0,
+        };
+        self.stack.push(success);
+
+        // Set the return_data of the parent context if a call, or of the
+        // root.
+        let return_data = self.parent_ctx_return_data();
```
Eikix, line 124: This gets the return_data from the parent ctx,
But I don't understand why we need it here

crates/evm/src/bytecode.cairo
```diff
@@ -0,0 +1,25 @@
+use contracts::contract_account::{ContractAccount, ContractAccountTrait};
+use contracts::kakarot_core::interface::{IKakarotCore};
+use contracts::kakarot_core::{KakarotCore};
+use evm::errors::{EVMError};
+use starknet::EthAddress;
+use utils::helpers::ByteArrayExTrait;
+
+/// Returns the bytecode of the EVM account (EOA or CA)
+fn bytecode(evm_address: EthAddress) -> Result<Span<u8>, EVMError> {
```
enitrat, line 9: refactor with #449

crates/evm/src/call_helpers.cairo
greged93: We should for sure fix this but you still have to handle every variant in the enum right? So doesn't change this code.

crates/evm/src/call_helpers.cairo
greged93: because the return data is set to the parent ctx in return

crates/evm/src/context.cairo
greged93: ah yeah nice that works

crates/evm/src/call_helpers.cairo
Eikix: Yeah you're right! I just wonder if the Active variant should return an Error, as it's an invariant and if the code is active, we have a problem

crates/evm/src/context.cairo
Eikix: I'm okay with either

crates/evm/src/machine.cairo
greged93: what would you set the parent ctx to instead? Should we keep the parent ctx?

crates/evm/src/machine.cairo
greged93: if I don't do anything with it I get `Variable was previously moved.` Do you want to keep the current ctx? Or return an error?

Eikix: I think we can return an error and introduce a result based return signature

crates/evm/src/machine.cairo
Eikix: We don't have to change it yeah

crates/evm/src/call_helpers.cairo
greged93: ahhhhhh ok ok mmmm yeah that makes sense.

crates/evm/src/machine.cairo
Eikix: Ahh, you think we need to set it to be "nothing" or say, Default, so  that the machine can take ownership of it?

crates/evm/src/call_helpers.cairo
Eikix: I don't understand

Eikix: like this?
https://github.com/kkrt-labs/kakarot-ssj/blob/72efecc9ab0168d130d05011709e3851f5a558dc/crates/evm/src/machine.cairo#L271

greged93: https://github.com/kkrt-labs/kakarot-ssj/issues/458

crates/evm/src/bytecode.cairo
Eikix: i think we can erase this and rebase it on the latest abstraction: AccountTrait

crates/evm/src/context.cairo
Eikix: Lmk if you think that's too verbose, otherwise it's fine

crates/evm/src/machine.cairo
```diff
@@ -251,17 +276,33 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
         code
     }
 
+    /// Returns whether the current execution context id.
```
Eikix, line 279: wrong comment i think

crates/evm/src/machine.cairo
```diff
@@ -251,17 +276,33 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
...
+    #[inline(always)]
+    fn execution_ctx_id(ref self: Machine) -> ExecutionContextId {
```
Eikix, line 281: i think we can rename this function to simply be `.id()` wdyt

Eikix (commented):
almost ready to merge imo

crates/evm/src/machine.cairo
greged93: I already made an `id` function that returns the actual id :/

crates/evm/src/machine.cairo
```diff
@@ -278,7 +278,7 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
 
     /// Returns whether the current execution context type.
     #[inline(always)]
-    fn execution_ctx_id(ref self: Machine) -> ExecutionContextId {
+    fn id(ref self: Machine) -> ExecutionContextId {
         let current_execution_ctx = self.current_ctx.unbox();
         let id = current_execution_ctx.execution_ctx_id();
```
Eikix, line 283: we should rename the underlying util execution_ctx_id as well

crates/evm/src/tests/test_utils.cairo
```diff
@@ -267,8 +267,10 @@ fn parent_ctx_return_data(ref self: Machine) -> Span<u8> {
     value
 }
 
-/// Sets the contract account bytecode at the provide ethereum address
+/// Sets the contract account bytecode at the provide ethereum address. Also
+/// increments the nonce in order to activate the contract.
 fn set_code(eth_address: EthAddress, bytecode: Span<u8>) -> Result<(), EVMError> {
     let mut ca = ContractAccountTrait::new(eth_address);
+    ca.increment_nonce()?;
```
Eikix, line 274: are we sure this is the place where the nounce should be incremented?
im not against but it seems a bit weird that set_code would increment the nonce

Eikix (commented):
Ready to merge imo and move onto #458

greged93:
@enitrat can you merge after CI is done? Neither me nor Elias can

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,157 @@
...
+        // to the parent context.
+        self.return_to_parent_ctx()
+    }
+}
+
+/// Check whether a `to` address for a call-family opcode is a precompile
+/// Since range check is expensive in Cairo, we proceed with checking equality
+fn is_precompile(to: EthAddress) -> bool {
+    let to: felt252 = to.into();
+    if to.into() <= 0x9_u256 {
```
enitrat, line 153:
```suggestion
    if to.into() < 10_u256 {
```

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,157 @@
...
+                return Result::Err(
+                    EVMError::InvalidMachineState(ACTIVE_MACHINE_STATE_IN_CALL_FINALIZATION)
+                );
+            },
+            Status::Stopped => 1,
+            Status::Reverted => 0,
+        };
+        self.stack.push(success)?;
+
+        // Set the return_data of the parent context if a call, or of the
```
enitrat, line 131: fix doc

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,157 @@
...
+        let child_ctx = ExecutionContextTrait::new(
+            ExecutionContextType::Call(self.ctx_count),
+            call_args.to,
+            call_ctx,
+            parent_ctx,
+            Default::default().span()
+        );
+
+        // Machine logic
+        // Increment the total context count
```
enitrat, line 104:
```suggestion
```

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,157 @@
...
+        self.ctx_count += 1;
+        // Set the current context to be the newly created child context
```
enitrat, line 106:
```suggestion
```

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,157 @@
...
+        self.memory.store_n(return_data, call_ctx.ret_offset);
+
+        // Return from the current sub ctx by setting the execution context
+        // to the parent context.
+        self.return_to_parent_ctx()
+    }
+}
+
+/// Check whether a `to` address for a call-family opcode is a precompile
+/// Since range check is expensive in Cairo, we proceed with checking equality
```
enitrat, line 150:
```suggestion
```

crates/utils/src/tests/test_helpers.cairo
```diff
@@ -289,6 +289,58 @@ fn test_bytearray_append_span_bytes() {
...
+        0x1D,
+        0x1E,
+        0x1F,
+        0x20,
+        0x21 // 33 elements
+    ];
+    let byte_array = ByteArrayExTrait::from_bytes(input.span());
+    let res = ByteArrayExTrait::into_bytes(byte_array);
+
+    // Ensure that the elements are correct
```
enitrat, line 333: I think you can Span has `PartialEq` so you can just check input.span() == res

crates/utils/src/helpers.cairo
```diff
@@ -1,3 +1,4 @@
+use array::ArrayTrait;
```
enitrat, line 1:
```suggestion
```

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -48,7 +58,31 @@ impl SystemOperations of SystemOperationsTrait {
     /// CALL
     /// # Specification: https://www.evm.codes/#f1?fork=shanghai
     fn exec_call(ref self: Machine) -> Result<(), EVMError> {
-        Result::Err(EVMError::NotImplemented)
+        // Prepare the call arguments with with_value = true for a call.
```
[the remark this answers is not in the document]
greged93: Given the fact that you pass kw args in cairo and that the linting is quite bad, I would tend to keep this one.

crates/evm/src/call_helpers.cairo
```diff
@@ -0,0 +1,157 @@
...
+    to: EthAddress,
+    gas: u128,
+    value: u256,
+    calldata: Span<u8>,
+    ret_offset: usize,
+    ret_size: usize,
+}
+
+#[generate_trait]
+impl MachineCallHelpersImpl of MachineCallHelpers {
```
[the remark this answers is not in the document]
greged93: It makes sense to be in call_helpers.cairo for me, can rename to `CallHelpersImpl of CallHelpers`.

enitrat (approved):

Issue #440, opened 2023-10-19
feat: create RLP::encode method


Eikix:
## Feature Request

For computing the eth address of a contract, we need to compute the RLP::encode(deployer_address, deployer_nonce) and then hash it, and take the 20 least significant bytes.

### Describe the Feature Request

Implement RLP::encode

### Related Code

The RLP library made by @Quentash 

### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

Noeljarillo:
Can I get this one?

PR #445, opened 2023-10-20
feat: rlp string encoding

enitrat:
- Refactored a bit how RLP is called introducing the RLPTrait
- Tiny improvements in RLP Type decoding by using __range checks only__ and not __range checks + equality checks__
- Implemented RLP string encoding.
- Added utils functions to know how many bytes a u32 uses and convert it to bytes
- Thoroughly tested the added utils and rlp_encoding
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #440

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -511,12 +511,12 @@ impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtTrait<T> {
 impl U32Impl of U32Trait {
     /// Packs 4 bytes into a u32
     /// # Arguments
-    /// * `self` a Span<u8> of len <=4
+    /// * `input` a Span<u8> of len <=4
     /// # Returns
     /// * Option::Some(u32) if the operation succeeds
     /// * Option::None otherwise
-    fn from_bytes(self: Span<u8>) -> Option<u32> {
-        let len = self.len();
+    fn from_bytes(input: Span<u8>) -> Option<u32> {
+        let len = input.len();
         if len > 4 {
             return Option::None(());
         }
@@ -527,12 +527,56 @@ impl U32Impl of U32Trait {
             if i == len {
                 break ();
             }
-            let byte: u32 = (*self.at(i)).into();
+            let byte: u32 = (*input.at(i)).into();
             result += byte.shl(8 * (offset - i));
 
             i += 1;
         };
         Option::Some(result)
+    }
+
+    /// Unpacks a u32 into an array of bytes
+    /// # Arguments
+    /// * `self` a `u32` value.
+    /// # Returns
+    /// * The bytes array representation of the value.
+    fn to_bytes(mut self: u32) -> Span<u8> {
+        let bytes_used: u32 = self.bytes_used().into();
+        let mut reversed_res: Array<u8> = Default::default();
+        let mut i = 0;
+        loop {
+            if i == bytes_used {
+                break ();
+            }
+            reversed_res.append((self & 0xFF).try_into().unwrap());
+            self = self.shr(8);
+            i += 1;
+        };
+
+        reversed_res.span().reverse().span()
+    }
+
+    /// Returns the number of bytes used to represent a `u64` value.
+    /// # Arguments
+    /// * `val` - The value to check.
+    /// # Returns
+    /// The number of bytes used to represent the value.
+    fn bytes_used(self: usize) -> u8 {
+        if self < 0x10000 { // 256^2
+            if self < 0x100 { // 256^1
+                if self == 0 {
+                    return 0;
+                } else {
+                    return 1;
+                };
+            }
+            return 2;
+        } else {
+            if self < 0x1000000 { // 256^6
+                return 3;
+            }
+            return 4;
+        }
     }
 }
 
@@ -555,6 +599,17 @@ impl U256Impl of U256Trait {
 }
 #[generate_trait]
 impl ByteArrayExt of ByteArrayExTrait {
+    fn append_span_bytes(ref self: ByteArray, mut bytes: Span<u8>) {
+        let mut i = 0;
+        loop {
+            if i == bytes.len() {
+                break;
+            }
+            self.append_byte(*bytes.at(i));
+            i += 1;
+        }
+    }
+
     fn from_bytes(mut bytes: Span<u8>) -> ByteArray {
         let mut arr: ByteArray = Default::default();
         let (nb_full_words, pending_word_len) = DivRem::div_rem(
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -542,18 +542,18 @@ impl U32Impl of U32Trait {
     /// * The bytes array representation of the value.
     fn to_bytes(mut self: u32) -> Span<u8> {
         let bytes_used: u32 = self.bytes_used().into();
-        let mut reversed_res: Array<u8> = Default::default();
+        let mut bytes: Array<u8> = Default::default();
         let mut i = 0;
         loop {
             if i == bytes_used {
                 break ();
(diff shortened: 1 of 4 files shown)


crates/utils/src/helpers.cairo
```diff
@@ -527,13 +527,57 @@ impl U32Impl of U32Trait {
...
+
+        reversed_res.span().reverse().span()
+    }
+
+    /// Returns the number of bytes used to represent a `u64` value.
+    /// # Arguments
+    /// * `val` - The value to check.
+    /// # Returns
+    /// The number of bytes used to represent the value.
+    fn bytes_used(self: usize) -> u8 {
```
Eikix, line 564: self is a u32

crates/utils/src/helpers.cairo
```diff
@@ -527,13 +527,57 @@ impl U32Impl of U32Trait {
...
+            }
+            reversed_res.append((self & 0xFF).try_into().unwrap());
+            self = self.shr(8);
+            i += 1;
+        };
+
+        reversed_res.span().reverse().span()
+    }
+
+    /// Returns the number of bytes used to represent a `u64` value.
```
Eikix, line 559:
```suggestion
    /// Returns the number of bytes used to represent a `u32` value.
```

crates/utils/src/helpers.cairo
```diff
@@ -527,13 +527,57 @@ impl U32Impl of U32Trait {
...
+    /// # Arguments
+    /// * `val` - The value to check.
+    /// # Returns
+    /// The number of bytes used to represent the value.
+    fn bytes_used(self: usize) -> u8 {
+        if self < 0x10000 { // 256^2
```
Eikix, line 565: nice dichotomy lul i like

crates/utils/src/rlp.cairo
```diff
@@ -24,101 +24,136 @@ enum RLPItem {
...
+    /// string data in Cairo.  Since RLP encoding is only used for eth_address
+    /// computation by calculating the RLP::encode(deployer_address, deployer_nonce)
+    /// and then hash it, the input is a ByteArray and not a Span<u8>
+    /// # Arguments
+    /// * `input` - ByteArray to encode
+    /// # Returns
+    /// * `ByteArray - RLP encoded ByteArray
+    /// # Errors
+    /// * RLPError::RlpEmptyInput - if the input is empty
+    fn encode_string(input: ByteArray) -> Result<ByteArray, RLPError> {
```
Eikix, line 59: why is the argument here a ByteArray?
will this be a drop-in replacement for Span<u8> overall in the codebase?

our specific usecase [for encode is to encode [deployer_address, deployer_nonce]](https://github.com/ethereum/execution-specs/blob/3fe6514f2d9d234e760d11af883a47c1263eff51/src/ethereum/paris/utils/address.py#L59)

crates/utils/src/rlp.cairo
```diff
@@ -24,101 +24,136 @@ enum RLPItem {
...
+        let len = input.len();
+        if len == 0 {
+            return Result::Err(RLPError::RlpEmptyInput(RLP_EMPTY_INPUT));
```
Eikix, line 62: I wonder if you can RLP something empty?
How do we encode the empty byte -> 0x80,

as @greged93 pointed out, this is the only way to encode a to==None

crates/utils/src/rlp.cairo
```diff
@@ -24,101 +24,136 @@ enum RLPItem {
...
+        } else if len == 1 && input.at(0).unwrap() < 0x80 {
+            return Result::Ok(input);
+        } else if len <= 55 {
+            let mut prefixes: ByteArray = Default::default();
+            prefixes.append_byte(0x80 + len.try_into().unwrap());
+            let encoding = ByteArrayTrait::concat(@prefixes, @input);
```
Eikix, line 68: should we ape in to the overload of concat?
```suggestion
            let encoding = prefixes + input;
```

crates/utils/src/rlp.cairo
```diff
@@ -24,101 +24,136 @@ enum RLPItem {
...
+            return Result::Ok(encoding);
+        } else {
+            let mut prefixes: ByteArray = Default::default();
+            let len_as_bytes = len.to_bytes();
+            let len_bytes_count = len_as_bytes.len();
```
Eikix, line 73: nice naming

crates/utils/src/rlp.cairo
enitrat: > will this be a drop-in replacement for Span overall in the codebase?

I would opt for, except for transaction data received from the RPC which will not be packed. In this specific situation It's either we split [address,nonce] into 2 span<u8> that we concatenate, or we can build a ByteArray in a probably more optimized way.

We need to run perf tests for that!

crates/utils/src/rlp.cairo
enitrat: > f a string is 0-55 bytes long, the RLP encoding consists of a single byte with value 0x80

An empty string would be encoded as 0x80, is that correct? In that case you would always be able to encode something, 🤔 which I think is the intended behavior, good catch

crates/utils/src/rlp.cairo
enitrat: what? that works? I'm honestly not sure I want to go this way

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -31,12 +32,145 @@ fn test_rlp_types() {
...
-    let res = rlp_decode(ArrayTrait::new().span());
+    let res = RLPTrait::decode(ArrayTrait::new().span());
 
     assert(res.is_err(), 'should return an error');
     assert(res.unwrap_err() == RLPError::RlpEmptyInput(RLP_EMPTY_INPUT), 'err != RlpInvalidLength');
 }
 
+#[test]
+#[available_gas(20000000)]
+fn test_rlp_encode_string_empty_input() {
```
Quentash, line 43: Should return a ByteArray[0x80]

Quentash (commented):
An empty input should return a ByteArray[0x80] instead of failing.
Looking good beside that 💪

crates/utils/src/helpers.cairo
```diff
@@ -555,6 +599,17 @@ impl U256Impl of U256Trait {
 }
 #[generate_trait]
 impl ByteArrayExt of ByteArrayExTrait {
+    fn append_span_bytes(ref self: ByteArray, mut bytes: Span<u8>) {
```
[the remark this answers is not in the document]
enitrat: refactored to 

```rust
    fn append_span_bytes(ref self: ByteArray, mut bytes: Span<u8>) {
        loop {
            match bytes.pop_front() {
                Option::Some(val) => { self.append_byte(*val) },
                Option::None => { break; }
            }
        }
    }
```

crates/utils/src/rlp.cairo
```diff
@@ -24,101 +24,136 @@ enum RLPItem {
...
+    /// * `ByteArray - RLP encoded ByteArray
+    /// # Errors
+    /// * RLPError::RlpEmptyInput - if the input is empty
+    fn encode_string(input: ByteArray) -> Result<ByteArray, RLPError> {
+        let len = input.len();
+        if len == 0 {
+            return Result::Err(RLPError::RlpEmptyInput(RLP_EMPTY_INPUT));
+        } else if len == 1 && input.at(0).unwrap() < 0x80 {
+            return Result::Ok(input);
+        } else if len <= 55 {
```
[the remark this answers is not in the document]
enitrat: aha yes

Eikix (approved):
lgtm

PR #447, opened 2023-10-20
Feat/extcodesize

Eikix:
## Pull Request type

implement extcodesize

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #127 

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/evm/src/instructions/environmental_information.cairo
+++ b/crates/evm/src/instructions/environmental_information.cairo
@@ -170,7 +170,21 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
     /// Get size of an account's code.
     /// # Specification: https://www.evm.codes/#3b?fork=shanghai
     fn exec_extcodesize(ref self: Machine) -> Result<(), EVMError> {
-        Result::Ok(())
+        let evm_address = self.stack.pop_eth_address()?;
+
+        let maybe_account = AccountTrait::account_at(evm_address)?;
+        let account = match maybe_account {
+            Option::Some(account) => account,
+            Option::None => { return self.stack.push(0); },
+        };
+
+        match account {
+            Account::EOA(eoa) => { return self.stack.push(0); },
+            Account::ContractAccount(ca) => {
+                let mut bytecode = ca.load_bytecode()?;
+                self.stack.push(bytecode.len().into())
+            }
+        }
     }
 
     /// 0x3C - EXTCODECOPY
--- a/crates/evm/src/tests/test_instructions/test_environment_information.cairo
+++ b/crates/evm/src/tests/test_instructions/test_environment_information.cairo
@@ -537,8 +537,76 @@ fn test_gasprice() {
 }
 
 // *************************************************************************
-// 0x3D: RETURNDATASIZE
+// 0x3B - EXTCODESIZE
 // *************************************************************************
+#[test]
+#[available_gas(20000000)]
+fn test_exec_extcodesize_eoa() {
+    // Given
+    let evm_address = evm_address();
+    let mut machine = setup_machine();
+    let kakarot_core = deploy_kakarot_core(native_token());
+    let expected_eoa_starknet_address = kakarot_core.deploy_eoa(evm_address);
+    machine.stack.push(evm_address.into());
+    set_contract_address(kakarot_core.contract_address);
+
+    // When
+    machine.exec_extcodesize().unwrap();
+
+    // Then
+    assert(machine.stack.peek().unwrap() == 0, 'expected code size 0');
+}
+
+
+#[test]
+#[available_gas(20000000)]
+fn test_exec_extcodesize_ca_empty() {
+    // Given
+    let evm_address = evm_address();
+    let mut machine = setup_machine();
+    let kakarot_core = deploy_kakarot_core(native_token());
+    set_contract_address(kakarot_core.contract_address);
+
+    // The bytecode remains empty, and we expect the empty hash in return
+    let mut contract_account = ContractAccountTrait::new(evm_address);
+
+    machine.stack.push(evm_address.into());
+
+    // When
+    machine.exec_extcodesize().unwrap();
+
+    // Then
+    assert(machine.stack.peek().unwrap() == 0, 'expected code size 0');
+}
+
+
+#[test]
+#[available_gas(20000000000)]
+fn test_exec_extcodesize_ca_with_bytecode() {
+    // Given
+    let evm_address = evm_address();
+    let mut machine = setup_machine();
+    let kakarot_core = deploy_kakarot_core(native_token());
+    set_contract_address(kakarot_core.contract_address);
+
+    // Set nonce of CA to 1 so that it appears as an existing account
+    // The bytecode stored is the bytecode of a Counter.sol smart contract
+    let mut contract_account = ContractAccountTrait::new(evm_address);
+    contract_account.increment_nonce().unwrap();
+    contract_account.store_bytecode(counter_evm_bytecode());
+
+    machine.stack.push(evm_address.into());
+    // When
+    machine.exec_extcodesize().unwrap();
+
+    // Then
+    assert(
+        machine.stack.peek() // extcodesize(Counter.sol) := 487 (source: remix)
+        .unwrap() == 275,
+        'expected counter SC code size'
+    );
+}
+
 
 #[test]
 #[available_gas(20000000)]
--- a/crates/evm/src/tests/test_instructions/test_environment_information.cairo
+++ b/crates/evm/src/tests/test_instructions/test_environment_information.cairo
@@ -601,7 +601,7 @@ fn test_exec_extcodesize_ca_with_bytecode() {
 
     // Then
     assert(
-        machine.stack.peek() // extcodesize(Counter.sol) := 487 (source: remix)
+        machine.stack.peek() // extcodesize(Counter.sol) := 275 (source: remix)
         .unwrap() == 275,
         'expected counter SC code size'
     );


Eikix:
Mhmm one of my tests is falsy, let me investigate

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -161,7 +167,20 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
     /// Get size of an account's code.
     /// # Specification: https://www.evm.codes/#3b?fork=shanghai
     fn exec_extcodesize(ref self: Machine) -> Result<(), EVMError> {
-        Result::Ok(())
+        let evm_address = self.stack.pop_eth_address()?;
+
+        // Case 1: The address corresponds to a EOA. If so, the account exists but has no code.
+        // We return size == 0.
+        let kakarot_state = KakarotCore::unsafe_new_contract_state();
```
enitrat, line 174: I would rather abstract this behavior, especially given that we replicate it a lot in the codebase. - I will think about it in subsequent PRs, but perhaps a Account enum with ContractAccount and EOA fields would be good.

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -208,8 +227,108 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+        let eoa_starknet_address = kakarot_state.eoa_starknet_address(evm_address);
+        if !eoa_starknet_address.is_zero() {
+            return self.stack.push(EMPTY_KECCAK);
+        }
+
+        let contract_account = ContractAccountTrait::new(evm_address);
+        // Case 2: The address corresponds to a non-existing or destroyed contract account.
+        // Checking if the nonce of a contrat account equals zero is a sufficient condition,
+        // as upon deployment, its nonce is increment to 1.
```
enitrat, line 249: All contract accounts have a nonce of 1 after deployment? So deploying a CA sets its nonce to 1?

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -208,8 +227,108 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+        }
+
+        // Case 3: The address corresponds to an existing contract account.
+        let mut bytecode = contract_account.load_bytecode()?;
+        // If the bytecode is empty, return the empty keccak hash.
+        if bytecode.is_zero() {
+            return self.stack.push(EMPTY_KECCAK);
+        }
+
+        // Else hash the bytecode and push the hash on the stack
```
enitrat, line 262: This should be a function on its own

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -208,8 +227,108 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+        // Since bytecode is a ByteArray, we need to perform a couple of transformations.
+        // `cairo_keccak` takes in an array of little-endian u64s
+        // First we compute how many full u64s there are in bytecode as well as the keccak `last_word`
+        let (full_u64_word_count, last_input_num_bytes) = DivRem::div_rem(
+            bytecode.len(), u32_as_non_zero(8)
+        );
+
+        // The `keccak_input` parameter of `cairo_keccak`
```
enitrat, line 270: I feel like there's too many comments here. Only keep the required ones, comments are dangerous because they don't throw errors when you refactor stuff :)

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -208,8 +227,108 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+        let mut keccak_input: Array<u64> = Default::default();
+        // A counter for the 8 bytes in a u64, runs in the range [0..8].
+        let mut byte_counter: u8 = 0;
+        // The temporary variable where we accumulate full u64 words
+        let mut tmp: u64 = 0;
```
enitrat, line 275:
```suggestion
        let mut pending_word: u64 = 0;
```

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -208,8 +227,108 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+                byte_counter += 1;
+            };
+
+        // Fill the last input word
+        let mut last_input_word: u64 = 0;
+        let mut byte_counter: u8 = 0;
+
+        // We enter a second loop for clarity.
+        // O(2n) should be okay
+        // We might want to regroup every computation into a single loop with appropriate `if` branching
```
enitrat, line 315: add TODO

crates/evm/src/instructions/environmental_information.cairo
enitrat: Can even be a ByteArray method on `ByteArrayExTrait` `to_u64_words`

It would return something like
```rust
{
    full_words: Array<u64>,
    pending_word:u64,
    pending_word_len:usize
}
```

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -1,15 +1,19 @@
-use contracts::kakarot_core::interface::IExtendedKakarotCoreDispatcherTrait;
+use contracts::contract_account::ContractAccountTrait;
+use contracts::kakarot_core::{interface::IExtendedKakarotCoreDispatcherImpl, KakarotCore};
+use contracts::tests::utils::counter_evm_bytecode;
 use contracts::tests::utils::{
     deploy_kakarot_core, deploy_native_token, fund_account_with_native_token
 };
+use debug::U256PrintImpl;
```
enitrat, line 7:
```suggestion
use debug::PrintTrait;
```

(if you still need it)

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -534,8 +538,77 @@ fn test_gasprice() {
...
+#[test]
+#[available_gas(20000000)]
+fn test_exec_extcodesize_ca_empty() {
+    // Given
+    let evm_address = evm_address();
+    let mut machine = setup_machine();
+    let kakarot_core = deploy_kakarot_core(native_token());
+    set_contract_address(kakarot_core.contract_address);
+
+    // Set nonce of CA to 1 so that it appears as an existing account
```
enitrat, line 571: where did you set it to 1?

crates/utils/src/traits.cairo
```diff
@@ -132,6 +132,20 @@ impl StoreBytes31 of Store<bytes31> {
     }
 }
 
+impl ByteArrayZero of Zero<ByteArray> {
```
enitrat, line 135: I'm against this! Zero represents the multiplicative identify element. Semantically an array cannot be zero. 

It can be empty, but not zero.

crates/evm/src/instructions/environmental_information.cairo
Eikix: but there is no reason to have a pending_word and a pending_word_len, ever,
if you convert a ByteArray to Array<u64>, I don't see why you would ever need to have "full" words and then one pending word

crates/utils/src/traits.cairo
Eikix: Does ByteArray have a "is_empty" method

crates/evm/src/instructions/environmental_information.cairo
Eikix: > I would rather abstract this behavior, especially given that we replicate it a lot in the codebase. - I will think about it in subsequent PRs, but perhaps a Account enum with ContractAccount and EOA fields would be good.

Actually good idea! Let's leverage the algebraic Enum system
Can you create an issue?

crates/evm/src/instructions/environmental_information.cairo
enitrat: > if you convert a ByteArray to Array, I don't see why you would ever need to have "full" words and then one pending word

What would happen if you had n bytes where $n \mod 8 \neq 0$ then? How do you feed the cairo_keccak the correct input if you don't know the size of the last word?

crates/utils/src/traits.cairo
enitrat: no but you can create it on `ByteArrayExTrait`

crates/evm/src/instructions/environmental_information.cairo
enitrat: #448

crates/evm/src/instructions/environmental_information.cairo
Eikix: Apologies, I meant, i feel the ByteArray type's philosophy is to pack bytes in an efficient way, so it only makes sense to exist with bytes31, not u64

crates/utils/src/traits.cairo
Eikix: I felt as though: zero is the additive neutral element, and since + is defined for bytearrays as concatenation, 
it made sense then to have zero trait,
i'm fine with is_empty

crates/evm/src/instructions/environmental_information.cairo
Eikix: Upon re-reading, I can delete one or two lines that are self-explanatory, but I feel documenting well how keccak related work is very important for clarity.

crates/evm/src/instructions/environmental_information.cairo
enitrat: > {
    full_words: Array<u64>,
    pending_word:u64,
    pending_word_len:usize
}

This is just what the `to_u64_words` return type could be: a conversion from bytes31 words to u64 words, and if the total amount of bytes is not a multiple of 8, then you need a field that tracks the pending word and the size of the pending word. It's not a redifinition of the ByteArray type, it's just the minimal amount of information you can output after converting your bytes31 words to u64 words

crates/evm/src/instructions/environmental_information.cairo
enitrat: > // The `keccak_input` parameter of `cairo_keccak`

The variable is named `keccak_input` so we already know it's  going to be the input for the keccak hash. Reading `cairo_keccak` we see that it's used as the first parameter, so it gives 0 information here

> // The temporary variable where we accumulate full u64 words
        let mut tmp: u64 = 0;

If you give it a better name like `pending_word` then you can deduce that it's going to store words that are not fully packed yet

> // Append to the keccak_input the accumulated u64
                    keccak_input.append(tmp);

redundant

> // Reset the counter and tmp
                    byte_counter = 0;

> // Increase the u64 counter
                    u64_word_counter += 1;
                    
Very easily deducible from the code itself

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -161,7 +167,20 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
     /// Get size of an account's code.
     /// # Specification: https://www.evm.codes/#3b?fork=shanghai
     fn exec_extcodesize(ref self: Machine) -> Result<(), EVMError> {
-        Result::Ok(())
+        let evm_address = self.stack.pop_eth_address()?;
+
+        // Case 1: The address corresponds to a EOA. If so, the account exists but has no code.
```
enitrat, line 172: refactor with #449

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -208,8 +227,107 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
+    // If the account has no code (or is a precompile), return the empty hash:
+    // `0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`
+    // If the account does not exist of was destroyed (SELFDESTRUCT), return 0
+    // Else return, the hash of the account's code
     /// # Specification: https://www.evm.codes/#3f?fork=shanghai
     fn exec_extcodehash(ref self: Machine) -> Result<(), EVMError> {
-        Result::Ok(())
+        let evm_address = self.stack.pop_eth_address()?;
+
+        // Case 1: The address corresponds to a EOA. If so, the account exists but has no code.
```
enitrat, line 238: refactor with #449

Q: why is `exec_extcodehash` implemented in this PR?

crates/evm/src/instructions/environmental_information.cairo
Eikix: Okay

crates/evm/src/instructions/environmental_information.cairo
Eikix: rebased on my last PR to reuse the utils/ tests,
though we should merge them in order, and I'll rebase to make it cleaner

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -170,7 +170,21 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
     /// Get size of an account's code.
     /// # Specification: https://www.evm.codes/#3b?fork=shanghai
     fn exec_extcodesize(ref self: Machine) -> Result<(), EVMError> {
-        Result::Ok(())
+        let evm_address = self.stack.pop_eth_address()?;
+
+        let maybe_account = AccountTrait::account_at(evm_address)?;
```
enitrat, line 175: so nice and smooth!

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -537,8 +537,76 @@ fn test_gasprice() {
...
+#[test]
+#[available_gas(20000000000)]
+fn test_exec_extcodesize_ca_with_bytecode() {
+    // Given
+    let evm_address = evm_address();
+    let mut machine = setup_machine();
+    let kakarot_core = deploy_kakarot_core(native_token());
+    set_contract_address(kakarot_core.contract_address);
+
+    // Set nonce of CA to 1 so that it appears as an existing account
```
enitrat, line 592: FYI this will change with my next PR but we can merge this first and I will rebase

crates/evm/src/tests/test_instructions/test_environment_information.cairo
```diff
@@ -537,8 +537,76 @@ fn test_gasprice() {
...
+    contract_account.store_bytecode(counter_evm_bytecode());
+
+    machine.stack.push(evm_address.into());
+    // When
+    machine.exec_extcodesize().unwrap();
+
+    // Then
+    assert(
+        machine.stack.peek() // extcodesize(Counter.sol) := 487 (source: remix)
+        .unwrap() == 275,
```
enitrat, line 605: what is 487 then?

Eikix: mb left by mistake

crates/evm/src/tests/test_instructions/test_environment_information.cairo
Eikix: okk!

crates/evm/src/instructions/environmental_information.cairo
Eikix: yess

enitrat (approved):
lgtm

Issue #448, opened 2023-10-21
feat: Create an Account enum and associated traits


enitrat:
### Describe the Feature Request
Modeling accounts and associated methods properly allows us to avoid duplicating code to check whether an account is an EOA or a CA. 
### Describe Preferred Solution
- Create an `Account` enum, with `EOA` and `ContractAccount` variants. 
- Create associated methods that return the variant corresponding to an account at address `evm_address`
- TODO: determine other useful methods to add.
### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

Eikix:
I can take this today or tomorrow 🔭🏎️

enitrat:
I'm almost done @Eikix

PR #449, opened 2023-10-23
feat: account model

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #448 

## What is the new behavior?

- Account enum model allowing us to fetch the correct type of account deployed at an address
- Refactored `balance` to be an account method and call the correct logic based on the account type

- [ ] Yes
- [ ] No


crates/eoa/src/tests/test_externally_owned_account.cairo
```diff
@@ -14,12 +14,15 @@ mod test_external_owned_account {
     fn deploy_eoa() -> IExternallyOwnedAccountDispatcher {
```
Eikix, line 14: come to think of it, this function should be in a utils file, since it's reused throughout the codebase

enitrat: agreed

crates/evm/src/model.cairo
```diff
@@ -13,3 +20,89 @@ struct ExecutionResult {
...
+
+
+/// An EVM Account is either an EOA or a Contract Account.  In both cases, the
+/// account is identified by an Ethereum address.  It has a corresponding
+/// Starknet Address - The corresponding Starknet Contract for EOAs, and the
+/// KakarotCore address for ContractAccounts.
+#[derive(Copy, Drop)]
+enum Account {
+    EOA: EOA,
+    ContractAccount: ContractAccount
```
Eikix, line 32: Should we add an EmptyAccount variant for semantics?
And have account_at return Result<Account, ...> since we no longer use None but rather EmptyAccount?

crates/evm/src/model.cairo
```diff
@@ -13,3 +20,89 @@ struct ExecutionResult {
...
+    /// Returns the balance in native token for a given EVM account (EOA or CA)
+    /// This is equivalent to checking the balance in native coin, i.e. ETHER of an account in Ethereum
+    #[inline(always)]
+    fn balance(self: @Account) -> Result<u256, EVMError> {
+        match self {
+            // Case 1: EOA is deployed
+            // BALANCE is the EOA's native_token.balanceOf(eoa_starknet_address)
+            Account::EOA(eoa) => { eoa.balance() },
+            // Case 2: EOA is not deployed and CA is deployed
+            // We check if a contract account is initialized at evm_address
```
Eikix, line 102: I think this check is not performed in this function

crates/evm/src/tests/test_model.cairo
```diff
@@ -0,0 +1,100 @@
...
+#[available_gas(20000000)]
+fn test_account_at_eoa() {
+    // Given
+    let native_token = deploy_native_token();
+    let kakarot_core = deploy_kakarot_core(native_token.contract_address);
+    let eoa = kakarot_core.deploy_eoa(evm_address());
+
+    // When
+    set_contract_address(kakarot_core.contract_address);
+    set_caller_address(kakarot_core.contract_address);
```
Eikix, line 20: why do we need both

crates/evm/src/model.cairo
Eikix: that way we can implement balance, nonce, etc. on it?

Eikix (commented):
this is great

Eikix:
I wonder if it makes sense to modify the entrypoints in Kakarot then,
maybe only for 
    /// Gets the balance associated to a contract account
        fn contract_account_balance(self: @ContractState, evm_address: EthAddress) -> u256 {
            let ca = ContractAccountTrait::new(evm_address);
            ca.balance().unwrap()
        }

crates/evm/src/model.cairo
enitrat: Perfect example of misleading useless comments :)

crates/evm/src/tests/test_model.cairo
enitrat: we don't these are leftovers from debugging session 😵

enitrat:
> I wonder if it makes sense to modify the entrypoints in Kakarot then,

- Modify to what? Generic `account_balance` or `account_storage_at` rather than `contract_account` specifically?

crates/evm/src/model.cairo
enitrat: I don't see a strong use case for this

Eikix:
> > I wonder if it makes sense to modify the entrypoints in Kakarot then,
> 
> * Modify to what? Generic `account_balance` or `account_storage_at` rather than `contract_account` specifically?

I guess, storage_at and nonce cannot be made generic, or they can and will panic in case it points to an EOA

enitrat:
- Refactored `contract_account_balance` entrypoint to `account_balance`
- Removed useless comments

crates/evm/src/model.cairo
Eikix: Fine with me

Eikix (approved):
lgtm

Issue #428, opened 2023-10-16
dev: improve rlp.cairo to support recursive list


Quentash:
The actual `rlp.cairo` doesn't support recursive list(#405), it does panic if a rlp transaction would contain such data.
The reason it was implemented this way was because the compiler didn't allow recursive struct.

When/If the compiler allows it, make use of recursive struct to be able to support recursive list from a rlp transaction.

PR #457, opened 2023-10-23
dev: improve rlp.cairo to support recursive list

Quentash:
Implemented recursive struct in rlp.cairo to be able to handle recursive list. I also tried to be as close as the code given in the ethereum doc ; https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/ .

Here is a list of the biggest changes :
- Types have been reduced to only be String or List.
- `decode_type()` ( former `from_bytes()` ) now takes an array to be able to retrieve the long strings/lists size. It does now return the type which is either a string or a list, the offset at which it starts and its size.
- `decode()` is now recursively called in case of recursive lists.

Here is a screenshot of gas consommation before the modifications ;

image

And here is one after the modifications ;

image

It's worth noting that some tests show improvement.
For instance, the `test_decode_string` appears to consume more gas than before, but this test is essentially a loop that converts a single character into an array, explaining the increased gas consumption. But on the other hand, the `test_decode_long_list `test, which deals with a larger amount of bytes, consumes less gas than it did originally.

I am certain that there is still potential for further improvement, and I would appreciate feedback on the direction these changes are taking. Whether you agree or disagree with some of the modifications.

## Pull Request type

- [X] Bugfix
- [ ] Feature
- [ ] Build-related changes

`rlp.cairo.decode()` doesn't handle recusrive list.

Resolves: #428 

## What is the new behavior?

- handle recursive list
-
-

- [ ] Yes
- [X] No


crates/utils/src/rlp.cairo
```diff
@@ -88,101 +105,49 @@ impl RLPImpl of RLPTrait {
...
-                let mut list_input = input.slice(1, len);
-                let decoded_list = rlp_decode_list(ref list_input)?;
-                Result::Ok((RLPItem::List(decoded_list), 1 + len))
-            },
-            RLPType::ListLong => {
-                let len_bytes_count = prefix.into() - 0xf7;
-                if input.len() <= len_bytes_count {
-                    return Result::Err(RLPError::RlpInputTooShort(RLP_INPUT_TOO_SHORT));
-                }
+    fn decode(input: Span<u8>) -> Result<Span<RLPItem>, RLPError> {
```
Eikix, line 108: why does this return a span of RLP items and not a RLP item?

crates/utils/src/rlp.cairo
```diff
@@ -88,101 +105,49 @@ impl RLPImpl of RLPTrait {
...
+        let mut output: Array<RLPItem> = Default::default();
+        let input_len = input.len();
+        let mut i = 0;
+
+        let mut decode_error: Option<RLPError> = loop {
+            let res = RLPTrait::decode_type(input.slice(i, input_len - i));
```
Eikix, line 114: can't we use ? operator to avoid having to use a match arm and break with an error option?

crates/utils/src/rlp.cairo
```diff
@@ -88,101 +105,49 @@ impl RLPImpl of RLPTrait {
...
+            let (rlp_type, offset, len) = match res {
+                Result::Ok(res_dec) => { res_dec },
+                Result::Err(err) => { break Option::Some(err); }
+            };
+
+            if input_len < offset + len {
+                break Option::Some(RLPError::RlpInputTooShort(RLP_INPUT_TOO_SHORT));
```
Eikix, line 121: I'm not a fan of option of errors tbh, can use a Result<(), E> instead, but don't have an extremely strong opinion, wdyt @enitrat

Eikix (commented):
First review, will have more questions in the coming days as I have more time to do a second review

crates/utils/src/rlp.cairo
Quentash: No because we are inside a loop. I would like to find a better solution if there is because it is indeed "heavy" like this

crates/utils/src/rlp.cairo
Quentash: I do agree with you, I'd like to hear a better way to handle those.
But since I am calling the function in a loop, I cannot use the "?" nor I can return an error directly.

crates/utils/src/rlp.cairo
Quentash: image

From my understanding of this function, it doesn't seem to stop after it found one item.
If we take this example : `the list [ "cat", "dog" ] = [ 0xc8, 0x83, 'c', 'a', 't', 0x83, 'd', 'o', 'g' ]` but removes the "0xc8". Their decore_function will concat "cat" and "dog" together, am I right ?
In our previous version, we would only have taken "cat" and then stops, ignoring the rest.
In this version, we will have a span of the 2 RLPItem "cat" and "dog" which seemed closer to the official doc algorithm. And if we keep the "0xc8", we will have a span containing a RLPItem List containing the 2 strings "cat" and "dog".

But this is one of the changes that I am not 100% sure about. It almost feels like this is a business logic, like... "do we take the first complete RLPItem decoded from the inputs or do we decode the inputs untill its end ?"

I hope I made myself understandable !

Quentash: Ok, GPT seems to agree with me that with the algorithm found in the ethereum doc (https://ethereum.org/en/developers/docs/data-structures-and-encoding/rlp/) give the same result for [ 0xc8, 0x83, 'c', 'a', 't', 0x83, 'd', 'o', 'g' ] as for [ 0x83, 'c', 'a', 't', 0x83, 'd', 'o', 'g' ]. It ends with the results being "catdog" in both case.

But I can't find any tests that actually verifiy that case... maybe the code presented in the doc isn't precise ( which I doubt ) or their implementation doesn't behave like the doc suggest ( which is weird ). This part is the most confusing one imo.

crates/utils/src/rlp.cairo
enitrat: so painful tbh 😵‍💫

Quentash: I'm so sorry...
Is there a proper reason to prevent people from returning inside a loop ? Should we panic instead ? ( I mean... the code, not trying to stress us )
I will think about a better way to handle this recursivity, my bad !

Quentash:
Made a better recursive implementation for the rlp_decode().
No more "Option::Error", no more loop!

Quentash:
I had to downgrade starknet/cairo versions in scarb files back to 2.3.0-rc0 because the recursivity seems to fail in Cairo 2.3.0 giving a very strange error (see screenshot).
image

Which still appears on github CL/Tests.
To make the tests run on local, install scarb-rc1 or 0, they both seem to handle struct recursivity. I don't understand why the 2.3.0 break it yet, maybe an issue should be open on the compiler github ?

Eikix:
Interestingly, there was a bug in the compiler, it'll be fixed in the next release

crates/utils/src/rlp.cairo
```diff
@@ -1,47 +1,62 @@
...
+    /// * `input` - Leading byte of the RLP encoded data
+    /// Return result with (RLPType, offset, size)
+    fn decode_type(input: Span<u8>) -> Result<(RLPType, u32, u32), RLPError> {
+        let input_len = input.len();
+        if input_len == 0 {
+            return Result::Err(RLPError::RlpEmptyInput(RLP_EMPTY_INPUT));
         }
-    }
 
+        let byte = *input[0];
```
bajpai244, line 32: A better name for this identifier can be `prefix_byte`, it makes it more clear what this byte does.

crates/utils/src/rlp.cairo
```diff
@@ -1,47 +1,62 @@
...
+
+        if byte < 0x80 { // Char
+            Result::Ok((RLPType::String, 0, 1))
+        } else if byte < 0xb8 { // Short String
+            Result::Ok((RLPType::String, 1, byte.into() - 0x80))
+        } else if byte < 0xc0 { // Long String
+            let len_bytes_count: u32 = (byte - 0xb7).into();
+            if input_len <= len_bytes_count {
+                return Result::Err(RLPError::RlpInputTooShort(RLP_INPUT_TOO_SHORT));
+            }
```
bajpai244, line 42: I was thinking, that might be we can rename this enum to:

`RlpInputTooShort` -> `RlpShortInput`, following the name of variant `RlpEmptyInput`, I think that is a better name for this enum.

Lemme know what u think about it.

CC: @Eikix , @enitrat

Eikix: > I was thinking, that might be we can rename this enum to:
> 
> `RlpInputTooShort` -> `RlpShortInput`, following the name of variant `RlpEmptyInput`, I think that is a better name for this enum.
> 
> Lemme know what u think about it.
> 
> CC: @Eikix , @enitrat

I disagree, though I think you're right to suggest a slight rename.
How about
```
enum RLPError {
   InputTooshort: felt252,
   EmptyInput: felt252,
}
```

Eikix: I personally like when errors are readable in natural language

Quentash: I feel like missing the "too" may be confusing because in RLP context, we can speak about Short List or Short String therefore "ShortInput" could sounds like the inputs of on of those, but I'm open to changes !

bajpai244: Okay, not a big issues with me then, @Quentash can u expand a little bit more on the above comment, I don't get much on it.

Quentash: RLP items can either be a String or a List, and within those types, there are Short and Long ones. So, I thought 'ShortInput' could refer to the inputs of a Short String or a Short List.
I hope I've made myself clearer ! :s

crates/utils/src/rlp.cairo
bajpai244: So, from my understanding, the results should differ:

For `0xc8, 0x83, 'c', 'a', 't', 0x83, 'd', 'o', 'g'`

The return type should be RLP item of type of `List`, where it should have two values inside.

Something like:
```javascript
List: ['cat', 'dog']
```
Now, for `[ 0x83, 'c', 'a', 't', 0x83, 'd', 'o', 'g' ]`

I don't think this should count as a valid RLP, because if u have more than 1 RLP item, it should be in a List, now I can understand the way the algorithms are implemented might end up picking `cat` and continue, but I think ideally it should fail! 

Lemme know what u think.

----------------------------------

## Return type

Following @Eikix comment, I think we should return an RLP Item, We can technically also do it with a Span, but an RLP Item sounds more accurate.

This way for an Ethereum transaction, we will get back the returned data as a List:

And the list will have the following structure { check EIP [2178](https://eips.ethereum.org/EIPS/eip-2718) for more info }:

EIP-1559:
`[chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, amount, data, access_list, signature_y_parity, signature_r, signature_s]`

EIP-2930 transactions:
`rlp([chainId, nonce, gasPrice, gasLimit, to, value, data, accessList, signatureYParity, signatureR, signatureS])`

Where access list should the type where we need recursion.

This is real decoding of txn from ethers { EIP-2930 tx  }:
tx: `0x01f86d8080843b9aca00825208941f9840a85d5af5bf1d1762f925bdaddc4201f98488016345785d8a000080c001a051ab92ea68ffb9289895d45fc1f1e7c163339bbba9be7abca4a5514d46ec2499a057834c79c1672fee11b168dd0d58488af5e4d9279962391307b318338bafb250`

decoded tx:
```javascript
[
  '0x', // chain id
  '0x', // nonce
  '0x3b9aca00', //gasPrice
  '0x5208', // gasLimit
  '0x1f9840a85d5af5bf1d1762f925bdaddc4201f984', // to
  '0x016345785d8a0000', // value
  '0x', // data
  [], // accessList
  '0x01', //  signatureYParity
  '0x51ab92ea68ffb9289895d45fc1f1e7c163339bbba9be7abca4a5514d46ec2499', //  signatureR
  '0x57834c79c1672fee11b168dd0d58488af5e4d9279962391307b318338bafb250' // signatureS
]
```

Quentash: Agree that decode should fail with an error (InputTooLong) for scenari like [ 0x83, 'c', 'a', 't', 0x83, 'd', 'o', 'g' ].

Decode() will now return a RLPItem instead of a Span<RLPItem>.

Scarb.toml
```diff
@@ -4,14 +4,14 @@ members = ["crates/*"]
...
-cairo-version = "2.3.0-rc0"
+cairo-version = "2.3.1"
 version = "0.1.0"
 readme = "README.md"
 repository = "https://github.com/kkrt-labs/kakarot-ssj/"
 license-file = "LICENSE"
 
 [workspace.dependencies]
-starknet = "2.3.0-rc0"
+starknet = "2.3.0"
```
Eikix: 2.3.1 I think

Scarb.toml
```diff
@@ -4,14 +4,14 @@ members = ["crates/*"]
 [workspace.package]
 description = "Kakarot is an (zk)-Ethereum Virtual Machine implementation written in Cairo."
 documentation = "https://www.kakarot.org/"
-cairo-version = "2.3.0-rc0"
+cairo-version = "2.3.1"
```
Eikix, line 7: you need to replace all occurences of 2.3.0-rc0, 2.3.0-rc1 and 2.3.0 in general in the whole repository

Eikix (approved):
lgtm

Scarb.lock
```diff
@@ -4,7 +4,7 @@ version = 1
 [[package]]
 name = "alexandria_storage"
 version = "0.2.0"
-source = "git+https://github.com/keep-starknet-strange/alexandria.git?rev=3356bf0c5c1a089167d7d3c28d543e195325e596#3356bf0c5c1a089167d7d3c28d543e195325e596"
+source = "git+https://github.com/keep-starknet-strange/alexandria.git?rev=ae1d5149ff601a7ac5b39edc867d33ebd83d7f4f#ae1d5149ff601a7ac5b39edc867d33ebd83d7f4f"
```
Eikix, line 7: this is the correct rev 0682b1e096fe40b2bbab69863e63f2c88ca06c77

Eikix: https://github.com/keep-starknet-strange/alexandria/commits/main

Eikix (approved):
Congratulations! Well fought

PR #459, opened 2023-10-23
feat: final pc should trigger stop

Eikix:
## Pull Request type

- added logic to STOP when PC reaches the end bytecode,
- open to challenges about: 
    - maybe pc reaching the end index should not trigger exec_stop but rather execute the logic of STOP.
    - should we wipe return_data field? I think so, especially considering #458 

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

when a PC reaches the end of a bytecode sequence, without any STOP, RETURN or REVERT, then it errors out. We need to change the behaviour so that it gracefully exits.

Resolves: #446 

## What is the new behavior?

- when reaching the final pc (pc == bytecode.len()), the execution context should halt successfully, without returning anything

- [ ] Yes
- [ ] No


Eikix:
Waiting on #458 to get this PR from draft to reviewable

Issue #422, opened 2023-10-12
feat: deploy contract account


enitrat:
## Feature Request
Create the deploy contract account flow.

 The bytecode of the contract deployed

## Minimal required parts in ssj

```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/kakarot-ssj/issues/424
- [ ] Call the `run` function with a machine initialized with the deploy transaction bytecode
- [ ] Use the return data of the run as the bytecode of the contract
- [ ] https://github.com/kkrt-labs/kakarot-ssj/issues/423
```

PR #460, opened 2023-10-23
feat: deploy contract account

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #422

## What is the new behavior?

- Moved `ContractAccountTrait` to be with `EOATrait` inside `evm` crate
- Refactored `ContractAccountTrait/EOATrait::at` to return `Option<Self>` depending on whether there is / the account at address `evm_address` is of the corresponding type
- `ContractAccountTrait::deploy` as the implementation logic
- `deploy_contract_account` internal KakarotCore method

- [ ] Yes
- [x] No


crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -78,6 +78,12 @@ mod KakarotCore {
         starknet_address: ContractAddress,
     }
 
+    #[derive(Drop, starknet::Event)]
+    struct ContractAccountDeployed {
+        #[key]
+        evm_address: EthAddress,
```
Eikix, line 84: just thinking out loud, should we include the deployer (for ease of indexing)

crates/evm/src/model/contract_account.cairo
```diff
@@ -43,6 +82,20 @@ impl ContractAccountImpl of ContractAccountTrait {
         Store::<u64>::read(0, storage_address).map_err(EVMError::SyscallFailed(READ_SYSCALL_FAILED))
     }
 
+    /// Sets the nonce of a contract account.
+    /// The new nonce is written in Kakarot Core's contract storage.
+    /// The storage address used is h(sn_keccak("contract_account_nonce"), evm_address), where `h` is the poseidon hash function.
```
Eikix, line 87: we should probably have a markdown document that records / centralizes all the storage addresses we decide to use

crates/evm/src/model/contract_account.cairo
```diff
@@ -24,12 +27,48 @@ struct ContractAccount {
...
 #[generate_trait]
 impl ContractAccountImpl of ContractAccountTrait {
-    /// Creates a new ContractAccount instance from the given `evm_address`.
+    /// Deploying a ContractAccount is setting its nonce to `1`.
+    /// # Arguments
+    /// * `evm_address` - The EVM address of the contract account to deploy
+    /// # Errors
+    /// Returns an `EVMError` if the nonce of the contract account for
+    /// this EVM address was not `0`.
+    fn deploy(evm_address: EthAddress) -> Result<ContractAccount, EVMError> {
```
Eikix, line 36: I think we need to add a todo for:
- setting code
- setting balance (in case of a payable constructor)

Eikix: Nvm it's down in kakarot_core above,
Then maybe this function should be call initialize?
or "touch"?

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -295,5 +301,38 @@ mod KakarotCore {
...
+            let execution_result = execute(
+                :evm_address,
+                :bytecode,
+                calldata: array![].span(),
+                :value,
+                gas_price: 0,
+                gas_limit: 0,
+            );
+            //TODO gas params
+            if execution_result.status == Status::Reverted {
```
Eikix, line 330: In this case, should the ContractAccount be "undeployed"? i.e. the nonce be set back to 0?
What does the spec say

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -295,5 +301,38 @@ mod KakarotCore {
...
+        ) -> (ContractAddress, bool) {
+            let mut ca = match ContractAccountTrait::deploy(evm_address) {
+                Result::Ok(account) => { account },
+                Result::Err(err) => { panic_with_felt252(err.to_string()) }
+            };
+            let execution_result = execute(
+                :evm_address,
+                :bytecode,
+                calldata: array![].span(),
+                :value,
```
Eikix, line 325: as I understand it, execute does not perform value transfer yet,
i wonder if the value transfer flow will happen inside `execute` or somewhere else using our greater struct Journal or LocalStorage

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -295,5 +301,38 @@ mod KakarotCore {
...
+                gas_price: 0,
+                gas_limit: 0,
+            );
+            //TODO gas params
+            if execution_result.status == Status::Reverted {
+                return (ca.starknet_address, false);
+            }
+
+            ca.store_bytecode(execution_result.return_data);
```
Eikix, line 334: this is an operation that can return an error, we should handle it

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -295,5 +301,38 @@ mod KakarotCore {
...
+        /// Deploys a contract account by setting up the storage associated to a contract account for a particular EVM address
+        /// # Arguments
+        /// * `evm_address` - The EVM address of the contract account
+        /// * `nonce` - The value to be transferred as part of this operation
+        /// * `balance` - The value to be transferred as part of this operation
+        /// * `bytecode` - The deploy bytecode
+        /// # Returns
+        /// * The address the contract is deployed at - which is KakarotCore
+        /// * `true` if the operation is successfull, false otherwise
+        fn deploy_contract_account(
```
Eikix, line 314: it may be worth it to look at how @ClementWalter does this in[ his recent PR](https://github.com/kkrt-labs/kakarot/pull/771/files#diff-10932ea6303303a44b4724d0a66ed9e547d2ea924e32f69ad1fbdcd4e68f9028R92)

crates/contracts/src/kakarot_core/kakarot.cairo
enitrat: If the context reverts then pending storage updates will not be committed so it should be a problem

crates/evm/src/errors.cairo
```diff
@@ -34,8 +38,27 @@ enum EVMError {
     UnknownOpcode: u8,
     SyscallFailed: felt252,
     WriteInStaticContext: felt252,
+    DeployError: felt252,
 }
 
+#[generate_trait]
+impl EVMErrorImpl of EVMErrorTrait {
+    fn to_string(self: EVMError) -> felt252 {
```
Eikix, line 46: Maybe we should altogether delete the Into<u256> I created, I'm not sure it's actually useful

crates/evm/src/model/contract_account.cairo
```diff
@@ -24,12 +27,48 @@ struct ContractAccount {
...
+        let mut maybe_ca = ContractAccountTrait::at(evm_address)?;
```
Eikix, line 37: Should we actually check for account_at -> we don't want an EOA to be deployed there

ideally, we'd want to have stronger invariants, such as the inability to deploy an EOA if you don't own the private key to it

crates/contracts/src/kakarot_core/kakarot.cairo
Eikix: but the nonce will be written to 1 above

Eikix:
some conflicts

crates/contracts/src/kakarot_core/kakarot.cairo
enitrat: Yes because we don't have transient state yet, but in the future we will have

crates/evm/src/model/contract_account.cairo
Eikix: okay

crates/evm/src/model/contract_account.cairo
enitrat: agreed

crates/contracts/src/kakarot_core/kakarot.cairo
enitrat: Happen inside execute normally

Eikix: ok

enitrat:
Related: #465

enitrat:
- Added some _very_ basic necessary error management features - these will most likely need to be refactored down the line to a stronger model
- Changed `decode_and_execute` to return if `pc>= bytecode_len` and set the status to Stopped
- use deploy in contract_account instead of direct access tests
- moved `deploy_contract_account` to `ContractAccountTrait::deploy` entirely. The KakarotCore contract doesn't need an internal function for this. #469

Eikix:
Some conflicts, reviewing now

crates/evm/src/interpreter.cairo
```diff
@@ -72,7 +72,8 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
 
         // Check if PC is not out of bounds.
         if pc >= bytecode_len {
-            return Result::Err(EVMError::InvalidProgramCounter(PC_OUT_OF_BOUNDS));
+            machine.set_stopped();
```
Eikix, line 75: Already started to tackle this in #459 
But I think if pc == bytecode_len we stop,
if pc > bytecode_len, there's a problem, we should return an error,
though this should never happen, as jump forbids non-valid jumps

crates/evm/src/interpreter.cairo
```diff
@@ -72,7 +72,8 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
...
+            return Result::Ok(());
```
Eikix, line 76: also, as in #459 if pc == bytecode_len, then we need to wipe current_ctx.return_data, otherwise as part of #458, it'll get copied onto its parent return_data field

crates/evm/src/machine.cairo
```diff
@@ -16,7 +17,8 @@ struct Machine {
     ctx_count: usize,
     stack: Stack,
     memory: Memory,
-    storage_journal: Journal
+    storage_journal: Journal,
+    error: Option<EVMError>
```
Eikix, line 21: we can revisit this later, I wonder how EELS or Geth bubbles up the error message or revert reason to the tx root ctx

crates/evm/src/machine.cairo
```diff
@@ -235,6 +239,12 @@ impl MachineCurrentContextImpl of MachineCurrentContextTrait {
         current_call_ctx.calldata()
     }
 
+
+    #[inline(always)]
+    fn error(ref self: Machine) -> Option<EVMError> {
```
Eikix, line 244: must we consume machine here?
we can't go with a snapshot?

crates/evm/src/model/contract_account.cairo
```diff
@@ -24,12 +32,62 @@ struct ContractAccount {
...
+    /// # Returns
+    /// * The evm_address and starknet_address the CA is deployed at - which is KakarotCore
+    /// # Errors
+    /// * `ACCOUNT_EXISTS` - If a contract account already exists at the given `evm_address`
+    /// * Any error returned by `execute`
+    fn deploy(
+        origin: EthAddress, evm_address: EthAddress, value: u256, bytecode: Span<u8>
+    ) -> Result<ContractAccount, EVMError> {
+        let mut maybe_acc = AccountTrait::account_at(evm_address)?;
+        match maybe_acc {
```
Eikix, line 53: I've seen this pattern a few times in this PR,
why not a simple early return inside an if?

```
let maybe_acc = AccountTrait::account_at(evm_address)?;
if maybe_acc.is_some() {
return ...
}
```

crates/evm/src/machine.cairo
Eikix: Ahh interesting, EELS has an optional error field on their evm object:) lgtm

crates/evm/src/model/contract_account.cairo
```diff
@@ -24,12 +32,62 @@ struct ContractAccount {
...
+            Option::Some(acc) => { return Result::Err(EVMError::DeployError(ACCOUNT_EXISTS)); },
+            Option::None => {}
+        }
+        let execution_result = execute(
```
Eikix, line 57: Are we sure this is the responsibiliy of AccountTrait::deploy to know of the EVM?
I'd argue AccountTrait::deploy should come after the execute is successful, and execute should be called outside of it with the proper arguments

Eikix: Or at least we call it process_create / process_create_message like the executions-specs?

enitrat: > I'd argue AccountTrait::deploy should come after the execute is successful, and execute should be called outside

I can agree with that indeed - question is, where should we do this `execute` then. Revert back to how it was inside KakarotCore?

crates/evm/src/model/contract_account.cairo
enitrat: yep

crates/evm/src/interpreter.cairo
enitrat: Let's get #459 merged before this PR then, as it's required to properly test

enitrat:
Restricting the scope of this PR, will split in more issues

https://github.com/kkrt-labs/kakarot-ssj/issues/474

crates/evm/src/tests/test_model.cairo
```diff
@@ -25,16 +27,17 @@ fn test_account_at_eoa() {
...
 #[available_gas(20000000)]
-#[ignore]
-fn test_account_at_ca() {
-    //TODO: implement this once ContractAccounts are deployable
+fn test_account_at_ca_exists() {
     // Given
     let native_token = deploy_native_token();
     let kakarot_core = deploy_kakarot_core(native_token.contract_address);
-    // kakarot_core.deploy_contrat_account(...);
+    // We need to set_contract_address as the nonce is stored inside the contract that calls this function - here, it's the test contract by default.
```
enitrat, line 34: funny thing heh

Eikix (approved):
lgtm

Eikix (approved):

Issue #150, opened 2023-08-22
feat: implement 0x41 - COINBASE Opcode


Eikix:
[...]

PR #466, opened 2023-10-24
feat: coinbase

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #150 

## What is the new behavior?

- coinbase is now the execution_info.sequencer_address field
- it is up to the underlying full node to set this variable: https://github.com/keep-starknet-strange/madara/blob/6b369880ea7ebd574abf7860286fa6c93c973249/crates/pallets/starknet/src/lib.rs#L451

- [ ] Yes
- [x] No


crates/utils/src/traits.cairo
```diff
@@ -1,3 +1,5 @@
+use core::option::OptionTrait;
```
enitrat, line 1: no need

crates/utils/src/traits.cairo
```diff
@@ -85,6 +87,13 @@ trait TryIntoResult<T, U> {
     fn try_into_result(self: T) -> Result<U, EVMError>;
 }
 
+impl EthAddressTryIntoResult of TryIntoResult<ContractAddress, EthAddress> {
```
enitrat, line 90:
```suggestion
impl EthAddressTryIntoResultContractAddress of TryIntoResult<ContractAddress, EthAddress> {
```

enitrat (commented):
So coinbase is going to return a 31bytes address right?

Eikix:
> So coinbase is going to return a 31bytes address right?

More like, when we use Madara / Katana / any full node, we'll make sure to set the sequencer address to be a 20bytes address 

and record a corresponding sn address in our registry mappping

enitrat:
> we'll make sure to set the sequencer address to be a 20bytes addres

How will you do that? It's going to take ages to compute

enitrat (dismissed):

enitrat (approved):

PR #472, opened 2023-10-25
chore: make contract crates compilable

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [x] Build-related changes

Resolves: #

## What is the new behavior?

- Allows compiling the contract crate to output a compiled contract that can be deployed.
- 
-

- [ ] Yes
- [ ] No


enitrat:
@Eikix why does `compute_starknet_address` fail?

crates/eoa/Scarb.toml
```diff
@@ -7,3 +7,6 @@ version = "0.1.0"
 [dependencies]
 starknet.workspace = true
 evm = { path = "../evm" }
+
+[tool]
```
Eikix, line 11: does this make our pre-commit hook finally work?

Eikix:
> @Eikix why does `compute_starknet_address` fail?
> 
> Is it possible that the class hash has changed with update to 2.3?

Weird, I think Harsh's PR has a similar problem. I don't understand why

Eikix:
Weirdly enough I rebased the feat/coinbase PR on main, and ran scarb test, but it passes
What could be different here?

enitrat:
The only change is import ordering which shouldn't modify the class hash of EOA 🤔

crates/eoa/Scarb.toml
enitrat: no, this just makes all crates use import formatting

Eikix (approved):
lgtm

Issue #469, opened 2023-10-24
dev: move `deploy_externally_owned_account` to `EOATrait::deploy`


enitrat:
Similar to what is done for `ContractAccountTrait::deploy`

PR #476, opened 2023-10-26
refactor: eoa deployment

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #469 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/evm/src/model/eoa.cairo
+++ b/crates/evm/src/model/eoa.cairo
@@ -1,6 +1,10 @@
+use contracts::kakarot_core::kakarot::KakarotCore::{
+    eoa_class_hashContractMemberStateTrait, eoa_address_registryContractMemberStateTrait,
+    ContractStateEventEmitter, EOADeployed
+};
 use contracts::kakarot_core::{IKakarotCore, KakarotCore};
 use evm::errors::{EVMError, CONTRACT_SYSCALL_FAILED, EOA_EXISTS};
-use evm::model::{Account, AccountType};
+use evm::model::{Account, AccountTrait, AccountType};
 use integer::BoundedInt;
 use openzeppelin::token::erc20::interface::{
     IERC20CamelSafeDispatcher, IERC20CamelSafeDispatcherTrait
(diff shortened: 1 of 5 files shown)


crates/evm/src/model/eoa.cairo
```diff
@@ -13,35 +17,35 @@ struct EOA {
...
     fn deploy(evm_address: EthAddress) -> Result<EOA, EVMError> {
-        //TODO finish in another PR(I started but realised it's out of scope)
-        // let maybe_eoa = EOATrait::at(evm_address)?;
-        // if maybe_eoa.is_some() {
-        //     return Result::Err(EVMError::DeployError(EOA_EXISTS));
-        // }
-
-        // let kakarot_address = get_contract_address();
-        // let calldata: Span<felt252> = array![kakarot_address.into(), evm_address.into()].span();
+        let maybe_eoa = EOATrait::at(evm_address)?;
```
Eikix, line 24: should this be AccountTrait::account_at?

crates/evm/src/model/eoa.cairo
```diff
@@ -13,35 +17,35 @@ struct EOA {
...
-        // match maybe_address {
-        //     Result::Ok((
-        //         contract_address, _
-        //     )) => {
-        //         Result::Ok(EOA { evm_address, starknet_address: contract_address }) },
-        //     Result::Err(err) => panic(err)
-        // }
-        panic_with_felt252('unimplemented')
+        let maybe_address = deploy_syscall(
+            eoa_class_hash.try_into().unwrap(), evm_address.into(), calldata, false
```
Eikix, line 35: why do we need to try into on eoa_class_hash?

crates/evm/src/model/eoa.cairo
```diff
@@ -13,35 +17,35 @@ struct EOA {
...
+        );
+        // Panic with err as syscall failure can't be caught, so we can't manage
```
Eikix, line 37: i'm looking at this match and I can't remember what we mean by "can't be caught", we're catching it here, no?

Eikix (commented):
some questions

crates/evm/src/model/eoa.cairo
enitrat: we don't, it's a no op

crates/evm/src/model/eoa.cairo
enitrat: Syscall errors management are not supported by the Starknet OS, so we can't actually manage the error. This will make the transaction fail either way

crates/evm/src/model/eoa.cairo
enitrat: yes

crates/evm/src/model/eoa.cairo
```diff
@@ -13,35 +18,53 @@ struct EOA {
...
+    /// * `evm_address` - The EVM address of the eoa
+    /// # Returns
+    /// * The corresponding Account instance
+    fn fetch(self: @EOA) -> Result<Account, EVMError> {
+        Result::Ok(
+            Account {
+                account_type: AccountType::EOA(*self),
+                code: Default::default().span(),
+                storage: Default::default(),
+                nonce: 1,
```
Eikix, line 64: We should add a comment saying that for now, this nonce is not used and value 1 is a sentinel value

crates/evm/src/model/eoa.cairo
```diff
@@ -13,35 +18,53 @@ struct EOA {
...
+                selfdestruct: false
```
Eikix, line 65: We should add a comment saying this is also a sentinel value?

Eikix (approved):

Issue #477, opened 2023-10-26
refactor: eoa_address_registry


enitrat:
### Describe the Feature Request
`eoa_address_registry` maps eoa addresses to their sn address equivalent.

### Describe Preferred Solution
You will rename it to `evm_address_registry`, so that it can be used for ALL types of addresses.

You will modify account deployment functions so that it writes the account's starknet address corresponding to the account's evm address in kakarot storage.
### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

akhercha:
Hey, can try this one! 🫡

enitrat:
@akhercha I think it has already been done but the issue wasn't closed, i will double check

PR #478, opened 2023-10-26
feat: fetch account from SN

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

- fetch an account from SN storage and return a memory representation of it
 - CA accounts load the nonce and byecode from storage
 - EOA accounts simply return a struct with nonce=1
 - If an account can't be fetched at address `address`, it means that it's an undeployed CA - it's instantiated empty in memory with a 0 nonce
 - This means we have to refactor the way of knowing whether an account is a CA or an EOA or undeployed, #477 
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/model.cairo
+++ b/crates/evm/src/model.cairo
@@ -5,13 +5,25 @@ use evm::errors::EVMError;
 use evm::execution::Status;
 use evm::model::contract_account::{ContractAccount, ContractAccountTrait};
 use evm::model::eoa::{EOA, EOATrait};
+use evm::storage_journal::{Journal, JournalTrait};
 use starknet::{EthAddress, get_contract_address, ContractAddress};
 use utils::helpers::ByteArrayExTrait;
 
+/// The struct representing an EVM event.
 #[derive(Drop)]
 struct Event {
     keys: Array<u256>,
     data: Array<u8>,
+}
+
+
+/// A struct to save Starknet native ETH transfers to be made when finalizing a
+/// tx
+#[derive(Copy, Drop)]
+struct Transfer {
+    sender: EthAddress,
+    recipient: EthAddress,
+    amount: u256
 }
 
 struct ExecutionResult {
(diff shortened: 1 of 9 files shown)


crates/evm/src/model.cairo
```diff
@@ -5,15 +5,27 @@ use evm::errors::EVMError;
...
     keys: Array<u256>,
     data: Array<u8>,
 }
 
+
+/// A struct to save Starknet native ETH transfers to be made when finalizing a
+/// tx
+#[derive(Copy, Drop)]
+struct Transfer {
+    sender: EthAddress,
```
Eikix, line 24: thinking out loud after @ClementWalter presentetion in weekly, should we adopt the Address{ethaddress, snaddress} type everywhere?

crates/evm/src/model.cairo
```diff
@@ -29,14 +41,60 @@ struct ExecutionResult {
...
+                match account_type {
+                    AccountType::EOA(eoa) => { eoa.fetch() },
+                    AccountType::ContractAccount(ca) => { ca.fetch() },
+                }
+            },
+            Option::None => {
+                return Result::Ok(
+                    // If no account exists at `address`, then
+                    // we are trying to access an undeployed contract account
+                    Account {
```
Eikix, line 80: but here we don't create? why fetch_or_create?

Eikix (commented):
looking at this PR, it feels as though when every kakarot account will be a SN account,
idk if it makes sense to have this separation between AccountType and Account

crates/evm/src/model.cairo
enitrat: This actually creates an account in "memory". It will be committed if the tx doesn't revert

enitrat:
> looking at this PR, it feels as though when every kakarot account will be a SN account,
idk if it makes sense to have this separation between AccountType and Account

I believe that our account model should reflect distinction, because even with KKT Accounts as SN Accounts, they remain two different account types and abstracting the "settlement" layer away from the data model used is a good design choice for refactoring imo. By using the type system and correctly instantiating account objects with the correct account type, we can actually "abstract/simplify/improve" our codebase and reduce bugs

Eikix (approved):
lgtm 🧠

Issue #481, opened 2023-10-27
refactor: make each CA in Kakarot SSJ a unique Starknet Smart Contract


Eikix:
Refactor:

```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/kakarot-ssj/issues/482
- [x] Modify contract_account.cairo to call a SN contract rather than KakarotCore storage
- [x] Modify deploy
- [x] Modify nonce
- [x] Modify bytecode
- [ ] #477
- [x] Test the new behaviour
```

Issue #468, opened 2023-10-24
tests: Test `deploy_contract_account` with an executable bytecode


enitrat:
Provide `deploy_contract_account` with the bytecode of the Counter contract. Verify that the stored bytecode matches the expected one.

PR #483, opened 2023-10-27
feat: create an account transparent proxy

Eikix:
## Pull Request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Resolves: #482 

## What is the new behavior?

- deploying an EOA now deploys an Account contract first and then upgrades it to become an EOA
- EOA constructor is never called if deployed from Kakarot -> maybe can delete it?

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #486, opened 2023-10-30
Feat/ca sn contract

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #481 #468

## What is the new behavior?

- create contract account SC (first commit)
- refactor to use contract account SC (second commit)
- added Kakarot Core getters and setters
- Fixed tests
- Added an enum for account in storage of KakarotCore address registry and a TODO to replace it once both accounts are ISO (e.g. nonce of EOA can be handled at application level instead of protocol)

- [x] Yes
- [ ] No


Eikix:
Rebased the PR on #483, will rebase on main once #483 is merged

crates/contracts/src/contract_account/contract_account.cairo
```diff
@@ -0,0 +1,223 @@
...
+        self.set_nonce(1);
+    }
+
+    #[event]
+    #[derive(Drop, starknet::Event)]
+    enum Event {
+        UpgradeableEvent: upgradeable_component::Event,
+    }
+
+    #[external(v0)]
```
enitrat, line 57: use `#[abi(per_item)]`

crates/evm/src/model/contract_account.cairo
```diff
@@ -50,34 +61,55 @@ impl ContractAccountImpl of ContractAccountTrait {
...
+        let calldata: Span<felt252> = array![kakarot_address.into(), evm_address.into()].span();
+
+        let maybe_address = deploy_syscall(account_class_hash, evm_address.into(), calldata, false);
+        // Panic with err as syscall failure can't be caught, so we can't manage
+        // the error
+        match maybe_address {
+            Result::Ok((
+                starknet_address, _
+            )) => {
+                let account = IUninitializedAccountDispatcher {
```
enitrat, line 79: change variable name (or better - just call directly on the dispatcher struct without storing it into a var since we won't reuse it)

crates/contracts/src/contract_account/contract_account.cairo
```diff
@@ -0,0 +1,223 @@
...
+        #[substorage(v0)]
+        upgradeable: upgradeable_component::Storage,
+    }
+
+    /// Constructor for Contract Account 
+    /// ⚠️ This constructor SHOULD not be called, as accounts in Kakarot are deployed as
+    /// Account.cairo instances, and then upgraded to EOA or CA implementations
+    /// In order to have predictable and unique Starknet Address upon deployment (<=> address mapping consistency)
+    #[constructor]
+    fn constructor(
```
enitrat, line 43: constructor is never ran since CAs are not meant to be directly deployed. I would rather see an `initialize` function called by the `upgrade` function in `UninitializedAccount` contract right after the replace class is done. 

This way, we could call `IAccount.initialize()` (which would set nonce to 1 for CAs and do nothing for EOAs)

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -2,6 +2,16 @@ use starknet::{ContractAddress, EthAddress, ClassHash};
 
 const INVOKE_ETH_CALL_FORBIDDEN: felt252 = 'KKT: Cannot invoke eth_call';
 
+// Local enum to differentiate EOA and CA in storage
+// TODO: remove distinction between EOA and CA as EVM accounts
+// As soon as EOA::nonce can be handled at the application level
+#[derive(Drop, starknet::Store, Serde, PartialEq, Default)]
```
enitrat, line 8: I don't like this being named `AccountType`. It's already defined in `model.cairo` but with different semantics?

\+ Storing enums is not super efficient, I would rather pack the address with a single bit indicating if it's an EOA or CA

crates/contracts/src/contract_account/contract_account.cairo
Eikix: I'm adding a constructor to be able to deploy a CA directly in unit tests of the CA

Eikix: Added in doc that in a normal flow, it is never called

crates/contracts/src/kakarot_core/kakarot.cairo
Eikix: Ok how about:
Renaming it to `StoredAccountType` and implementing StorePack on it?

crates/contracts/src/contract_account/contract_account.cairo
enitrat: not a fan of this. why not just a helper test function that deploys + initializes a CA

Eikix: ok!

crates/contracts/src/kakarot_core/storage_types.cairo
```diff
@@ -0,0 +1,51 @@
...
+            .expect('Address should fit in 251 bytes');
+        if discriminant == 1 {
+            return StoredAccountType::EOA(address);
+        }
+        if discriminant == 2 {
+            return StoredAccountType::ContractAccount(address);
+        }
+
+        // SHOULD be unreachable code
+        return StoredAccountType::UninitializedAccount;
```
enitrat, line 49: I'd rather panic for something that should never happen than return something

crates/evm/src/model/contract_account.cairo
```diff
@@ -298,12 +260,11 @@ impl ContractAccountImpl of ContractAccountTrait {
...
-        let data_address = compute_storage_base_address(
-            selector!("contract_account_valid_jumps"),
-            array![self.evm_address.into(), offset.into()].span()
-        );
-        Store::<bool>::write(0, data_address, true)
-            .map_err(EVMError::SyscallFailed(WRITE_SYSCALL_FAILED))
+        let mut contract_account = IContractAccountDispatcher {
+            contract_address: self.starknet_address
+        };
+        contract_account.set_false_jumpdest(offset);
```
enitrat, line 266: mismatch with function name

crates/evm/src/model/contract_account.cairo
```diff
@@ -186,12 +204,11 @@ impl ContractAccountImpl of ContractAccountTrait {
...
-        let storage_address = compute_storage_base_address(
-            selector!("contract_account_storage_keys"),
-            array![self.evm_address.into(), key.low.into(), key.high.into()].span()
-        );
-        Store::<u256>::write(0, storage_address, value)
-            .map_err(EVMError::SyscallFailed(WRITE_SYSCALL_FAILED))
+        let mut contract_account = IContractAccountDispatcher {
+            contract_address: self.starknet_address
+        };
+        contract_account.set_storage_at(key, value);
```
enitrat, line 210: should we handle SyscallErrors like in `balance()`

crates/contracts/src/contract_account/contract_account.cairo
```diff
@@ -0,0 +1,220 @@
...
+
+        /// Checks if for a specific offset, i.e. if  bytecode at index `offset`, bytecode[offset] == 0x5B && is part of a PUSH opcode input.
+        /// Prevents false positive checks in JUMP opcode of the type: jump destination opcode == JUMPDEST in appearance, but is a PUSH opcode bytecode slice.
+        #[external(v0)]
+        fn is_false_jumpdest(self: @ContractState, offset: usize) -> bool {
+            panic_with_felt252('unimplemented')
+        }
+
+        #[external(v0)]
+        fn set_false_jumpdest(ref self: ContractState, offset: usize) {
```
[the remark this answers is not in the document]
Eikix: the reasoning is that we want to minimize the storage that we use. For a bytecode that has 50 jumps, 50 jumpdest, and say, 3 PUSH opcodes that contain 0x5B (jumpdest), i want to store only 3 values

a very simple initial check -> offset in pc == 0x5B can help us determine if a jump destination is invalid
then we must simple check against storage that it's not a false positive

i named it false as in, false positive, i can rename it

crates/contracts/src/contract_account/contract_account.cairo
```diff
@@ -0,0 +1,220 @@
...
+        #[external(v0)]
+        fn kakarot_core_address(self: @ContractState) -> ContractAddress {
+            self.kakarot_core_address.read()
+        }
+        #[external(v0)]
+        fn evm_address(self: @ContractState) -> EthAddress {
+            self.evm_address.read()
+        }
+
+        // Contract Account Specific Methods
```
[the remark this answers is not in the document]
Eikix: okay!

crates/evm/src/model/contract_account.cairo
Eikix: yeah I asked myself the same question, there are a lot of .expect in my PR.
I am under the impression that matching on SyscalllErrors does nothing rn, but will be enabled "soon"?

crates/contracts/src/kakarot_core/storage_types.cairo
```diff
@@ -0,0 +1,51 @@
...
+use starknet::{ContractAddress, StorePacking};
+use utils::constants::POW_2_8;
+use utils::math::Bitshift;
+use utils::traits::U256TryIntoContractAddress;
+
+// Local enum to differentiate EOA and CA in storage
+// TODO: remove distinction between EOA and CA as EVM accounts
+// As soon as EOA::nonce can be handled at the application level
+#[derive(Drop, Serde, PartialEq, Default)]
+enum StoredAccountType {
```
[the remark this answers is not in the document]
Eikix: What do you mean Store 1 + address?
do you mean store 0x01 at the first byte, and then the address itself, with a shift left of 8 bits?

crates/evm/src/model/contract_account.cairo
Eikix: thx

crates/contracts/src/kakarot_core/storage_types.cairo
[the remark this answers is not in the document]
Eikix: I don't think this works 100% of the time as addresses are at most 251 bits, and max felt is strictly less than 252 bits

crates/contracts/src/contract_account/contract_account.cairo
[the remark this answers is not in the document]
enitrat: ok, i would prefer a rename because I did not understand what it was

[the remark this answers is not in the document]
Eikix: renamed to set_false_positive_jumpdest

crates/evm/src/model/contract_account.cairo
enitrat: well here we return Result<(), EVMERror> but nothing can throw an EVMError so either we change the signature or we map_err the set_storage_at to an EVMError type

Eikix: https://github.com/kkrt-labs/kakarot-ssj/issues/494

enitrat (approved):
image

Issue #485, opened 2023-10-30
dev: move EOA modules from EOA crate inside `contracts` crate as a module


Eikix:
```[tasklist]
### Tasks
- [ ] Create an eoa folder in the crate `contracts`, inside `src`
- [ ] Move all code from eoa crate to this newly created folder
- [ ] Fix imports
- [ ] Delete the EOA crate
- [ ] Delete dependencies to the EOA crate in all other crates of the workspace
- [ ] Fix tests
- [ ] Fix imports again^^
- [ ] Don't forget to scarb format
```

Quentash:
Want me to take care of this ?

Eikix:
> Want me to take care of this ?

Great, yes please

PR #497, opened 2023-11-01
Dev : eoa crate migration into contracts crates

Quentash:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #485 

## What is the new behavior?

- Migrated eoa crate into contratcs crate src directory and corrected imports.
-
-

- [ ] Yes
- [X] No


crates/contracts/src/eoa.cairo
enitrat, line 1: do we need eoa.cairo -> externally_owned_account.cairo ?
I think we can just add the content of `externally_owned_account` inside `eoa `directly

Eikix: i vote that we have interface.cairo and eoa.cairo inside the eoa folder, to match contract_account

Eikix: wdyt @Quentash @enitrat

enitrat: and re-export them from there like OZ does?

enitrat: > to match contract_account

well, the question would be, do we also need this in CA ?

Eikix: yeah we could do that

> and re-export them from there like OZ does?

Eikix: tbh i don't really mind either way

Quentash: Hmm, I'm not really sure what I gotta do here ^^'
I've moved `externally_owned_account` inside `eoa` directly, but are you suggesting to create a eoa folder inside contracts/src instead ?

Eikix:
Closing in favour of #515 because of immense git conflicts and change of scope of the PR

Issue #499, opened 2023-11-02
bug: default values for integer not implemented in the RLP


bajpai244:
## Bug Report

**Kakarot version: 8b695ec5203cb645ab5a75d30cad9608bb227dec**

**Current behavior:**

In the current codebase, if we encounter the following byte [0x80], this should result in the int `0`, it is the default value for integer `0`.

<img width="313" alt="image" src="https://github.com/kkrt-labs/kakarot-ssj/assets/41180869/3d8f5897-b76d-4922-80ee-bc3d105a8f4d">

[source](url).

In our case, [we return](https://github.com/kkrt-labs/kakarot-ssj/blob/8b695ec5203cb645ab5a75d30cad9608bb227dec/crates/utils/src/rlp.cairo#L39) the `offset` to be `1`, and the `length` to be `0`, hence when we create a slice, we [do](https://github.com/kkrt-labs/kakarot-ssj/blob/8b695ec5203cb645ab5a75d30cad9608bb227dec/crates/utils/src/rlp.cairo#L120C5-L120C5) the following:
- `input.slice(1, 0)`

This will return in an empty slice [], but according to the spec, what should be returned should be `0` or in our case `[0]`.

**Expected behavior:**

The returned value should be `[0]` and not `[]` when the byte `[0x80]` is RLP decoded.

PR #500, opened 2023-11-02
fix: fix parsing of default value for int `0` from byte `[0x80]`

bajpai244:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

When we encounter byte `[0x80]` we return `[]` i.e an empty slice, but we should return the default value `[0]` as per the spec.

Resolves: #499 

## What is the new behavior?

We return the default value `[0]` when we encounter the byte `[0x80]`.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/utils/src/rlp.cairo
+++ b/crates/utils/src/rlp.cairo
@@ -121,8 +121,9 @@ impl RLPImpl of RLPTrait {
                 // checking for default value `0`
                 if (*input[0] == 0x80) {
                     output.append(RLPItem::String(array![0].span()));
+                } else {
+                    output.append(RLPItem::String(input.slice(offset, len)));
                 }
-                output.append(RLPItem::String(input.slice(offset, len)));
             },
             RLPType::List => {
                 if len > 0 {
--- a/crates/utils/src/rlp.cairo
+++ b/crates/utils/src/rlp.cairo
@@ -119,7 +119,7 @@ impl RLPImpl of RLPTrait {
         match rlp_type {
             RLPType::String => {
                 // checking for default value `0`
-                if (*input[0] == 0x80) {
+                if (len == 0) {
                     output.append(RLPItem::String(array![0].span()));
                 } else {
                     output.append(RLPItem::String(input.slice(offset, len)));
--- a/crates/utils/src/rlp.cairo
+++ b/crates/utils/src/rlp.cairo
@@ -26,7 +26,7 @@ impl RLPImpl of RLPTrait {
     /// the offset and the size of the RLPItem to decode
     /// # Errors
     /// * RLPError::EmptyInput - if the input is empty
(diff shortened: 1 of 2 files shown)


bajpai244:
Looks like 1 RLP test is failing, lemme look why this is happening!

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -17,6 +17,17 @@ fn test_rlp_decode_type_byte() {
...
 
+#[test]
+#[available_gas(99999999)]
+fn test_rlp_decode_string_default_value() {
+    let mut arr = array![0x80];
+
+    let rlp_item = RLPTrait::decode(arr.span()).unwrap();
+    let expected = RLPItem::String(array![0].span());
+
+    assert(*rlp_item[0] == expected, 'default value not 0');
```
Eikix, line 28: can assert the rlp_item is of length 1?

Eikix (commented):
nice catch, we handled this case iirc in the first implementation of rlp in ssj, seems it was lost

crates/utils/src/rlp.cairo
```diff
@@ -117,7 +117,13 @@ impl RLPImpl of RLPTrait {
         }
 
         match rlp_type {
-            RLPType::String => { output.append(RLPItem::String(input.slice(offset, len))); },
+            RLPType::String => {
+                // checking for default value `0`
+                if (*input[0] == 0x80) {
+                    output.append(RLPItem::String(array![0].span()));
+                }
+                output.append(RLPItem::String(input.slice(offset, len)));
```
Eikix, line 125: this runs also

crates/utils/src/tests/test_rlp.cairo
bajpai244: great suggestion!

Added via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/500/commits/1e7e11af0b4cc8bfcaf13b10e2259c9c2c72acee) commit.

crates/utils/src/rlp.cairo
```diff
@@ -117,7 +117,14 @@ impl RLPImpl of RLPTrait {
         }
 
         match rlp_type {
-            RLPType::String => { output.append(RLPItem::String(input.slice(offset, len))); },
+            RLPType::String => {
+                // checking for default value `0`
+                if (*input[0] == 0x80) {
```
Eikix, line 122:
```suggestion
                if (len == 0) {
```

since len is computed by decode_type, 
since further increases consistency with the codebase

bajpai244: good change! Implemented via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/500/commits/184de9f588e7226c2e29718ef92cec39b36bedf7 ) commit.

Issue #504, opened 2023-11-02
feat: infinite approval to native token in accounts deployment


enitrat:
### Describe the Feature Request
Add an infinite approval to the native token of Kakarot during account deployment
### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

PR #505, opened 2023-11-02
feat: account infinite token approval

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Lost sanity

Resolves: #504 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/contracts/src/tests/test_contract_account.cairo
+++ b/crates/contracts/src/tests/test_contract_account.cairo
@@ -4,11 +4,14 @@ use contracts::contract_account::interface::{
 use contracts::tests::test_data::counter_evm_bytecode;
 use contracts::tests::test_utils as contract_utils;
 use evm::tests::test_utils::{ca_address, native_token};
+use starknet::testing::set_contract_address;
 
 #[test]
 #[available_gas(3000000000)]
 fn test_ca_deploy() {
-    let kakarot_core = contract_utils::deploy_kakarot_core(native_token());
+    let native_token = contract_utils::deploy_native_token();
+    let kakarot_core = contract_utils::deploy_kakarot_core(native_token.contract_address);
+    set_contract_address(kakarot_core.contract_address);
     let contract_account = contract_utils::deploy_contract_account(
         kakarot_core.contract_address, Default::default().span()
     );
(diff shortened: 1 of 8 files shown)


crates/contracts/src/tests/test_contract_account.cairo
```diff
@@ -4,11 +4,14 @@ use contracts::contract_account::interface::{
 use contracts::tests::test_data::counter_evm_bytecode;
 use contracts::tests::test_utils as contract_utils;
 use evm::tests::test_utils::{ca_address, native_token};
+use starknet::testing::set_contract_address;
 
 #[test]
 #[available_gas(3000000000)]
 fn test_ca_deploy() {
-    let kakarot_core = contract_utils::deploy_kakarot_core(native_token());
+    let native_token = contract_utils::deploy_native_token();
```
Eikix, line 12: we should put this in a util fn called kakarot_fixture or smth, to say that it includes all the necesary steps to make a SC test work

it'd include all these 3 lines, would save a bit of time and redundancy

crates/contracts/src/uninitialized_account/uninitialized_account.cairo
```diff
@@ -40,6 +41,12 @@ mod UninitializedAccount {
                 get_caller_address() == self.kakarot_core_address.read(),
                 'Caller not Kakarot Core address'
             );
+            let kkt_address = self.kakarot_core_address.read();
```
Eikix, line 44: let's never use kkt in the repo, only kakarot

crates/contracts/src/uninitialized_account/uninitialized_account.cairo
```diff
@@ -40,6 +41,12 @@ mod UninitializedAccount {
...
+            let native_token = IKakarotCoreDispatcher { contract_address: kkt_address }
+                .native_token();
+            IERC20CamelDispatcher { contract_address: native_token }
+                .approve(kkt_address, integer::BoundedInt::<u256>::max());
```
Eikix, line 48: Maybe we can add a comment to explain why we do this

crates/contracts/src/tests/test_contract_account.cairo
enitrat: I agree but I think this will be handled in a future where we eventually transit to sn-foundry

Eikix: i completely agree, i mean even for now as a quick fix, it'd make us "win" a lot of lines to replace these lines by a standalone util like setup_sc_fixture(), it could even help for the migration as itll centralize all the SC fixtures we need

Eikix: this can be a good first issue if you'd prefer

enitrat: yes, let's do this.

enitrat: https://github.com/kkrt-labs/kakarot-ssj/issues/509

PR #506, opened 2023-11-02
Feat/value transfer

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No


crates/evm/src/model/contract_account.cairo
```diff
@@ -96,6 +97,19 @@ impl ContractAccountImpl of ContractAccountTrait {
         }
     }
 
+    #[inline(always)]
+    fn addresses(self: @ContractAccount) -> Address {
```
Eikix, line 101: since the model is called Address, i think we can call this getter Address?

crates/evm/src/state.cairo
```diff
@@ -282,7 +282,10 @@ impl StateImpl of StateTrait {
         match self.balances.read(address.evm.into()) {
             Option::Some(value) => { Result::Ok(value) },
             //TODO(state) - implement this post CA-rebase
```
Eikix, line 284: can delete the comment

crates/evm/src/model/account.cairo
```diff
@@ -51,6 +51,46 @@ impl AccountImpl of AccountTrait {
...
+        match maybe_acc {
+            Option::Some(account_type) => {
+                match account_type {
+                    AccountType::EOA(eoa) => {
+                        // no - op
+                        Result::Ok(())
+                    },
+                    AccountType::ContractAccount(mut ca) => {
+                        if *self.selfdestruct {
+                            return ca.selfdestruct();
```
Eikix, line 67: what do you mean by return ca.selfdestruct();?

crates/evm/src/execution.cairo
```diff
@@ -58,16 +63,36 @@ fn execute(
     // Initiate the Machine with the root context
     let mut machine: Machine = MachineCurrentContextTrait::new(ctx);
 
+    // Handle value transfers
+    let sender: Address = AccountTrait::fetch_or_create(origin)
+        .expect('couldnt fetch sender acc')
```
Eikix, line 68: since this is part of the execution of the EVM, we cannot revert the cairovm imo, 
should we find an elegant way to catch errors and simply revert the root tx?

crates/evm/src/interpreter.cairo
```diff
@@ -668,9 +668,7 @@ impl EVMInterpreterImpl of EVMInterpreterTrait {
     fn finalize_context(ref self: EVMInterpreter, ref machine: Machine) {
         machine.state.commit_context();
         match machine.ctx_type() {
-            ExecutionContextType::Root => {
-                // TODO: error handling
-                machine.state.commit_state();
+            ExecutionContextType::Root => {// TODO: error handling
```
Eikix, line 671: why did we delete commit state?

crates/evm/src/model/account.cairo
```diff
@@ -51,6 +51,46 @@ impl AccountImpl of AccountTrait {
...
+            },
+            Option::None => {
+                //Case new account
+                // If SELFDESTRUCT, just do nothing
+                if (*self.selfdestruct == true) {
+                    return Result::Ok(());
+                }
+                ContractAccountTrait::deploy(self.address().evm, *self.code);
+                Result::Ok(())
+            //Storage is handled outside of the account and must be commited after all accounts are commited.
```
Eikix, line 83: Wdym

crates/evm/src/tests/test_execution.cairo
```diff
@@ -0,0 +1,54 @@
...
+#[available_gas(20000000)]
+fn test_execute_value_transfer() {
+    let native_token = contract_utils::deploy_native_token();
+    // Transfer native tokens to sender
+    let kakarot_core = contract_utils::deploy_kakarot_core(native_token.contract_address);
+    set_contract_address(kakarot_core.contract_address);
+    let sender = EOATrait::deploy(evm_address()).expect('sender deploy failed');
+    let recipient = EOATrait::deploy(other_evm_address()).expect('recipient deploy failed');
+    // Transfer native tokens to sender - we need to set the contract address for this
+    set_contract_address(contract_utils::constants::ETH_BANK());
```
Eikix, line 20: there's a fund_address util iirc (im modifying it in an open PR to take as argument to amount to fund)

crates/evm/src/tests/test_execution.cairo
```diff
@@ -0,0 +1,54 @@
...
+        gas_limit: 0,
+        read_only: false,
+    );
+    // `commit_state` is applied in `eth_send_tx` only - to test that `execute` worked correctly, we manually apply it here.
+    let sender_balance = IERC20CamelDispatcher { contract_address: native_token.contract_address }
+        .balanceOf(sender.starknet_address);
+    let recipient_balance = IERC20CamelDispatcher {
+        contract_address: native_token.contract_address
+    }
+        .balanceOf(recipient.starknet_address);
```
Eikix, line 42: why do we shadow these values?
i mean why bother in the first place?

crates/evm/src/execution.cairo
enitrat: yes, but since these don't run from the machine, we will need to merge errors from the machine with errors from these in the `ExecutionResult` struct returned.

Wait, actually, if this fails (even before the machine is started) then we should just return early from the function with this error. I'll change this

crates/evm/src/model/account.cairo
enitrat: First you commit accounts on chain ( so you write nonce and bytecode on chain).
Then, once you have done this commitment which handles the deployment of a CA in case it's not deployed yet, you commit the storage of the account. But since this Storage is not contained in the `Account` data structure, you need to do this from the `commit` function in `StateTrait` directly, and not in the `AccountTrait::commit`

crates/evm/src/tests/test_execution.cairo
enitrat: I don't see it yet (or under another name?)

crates/evm/src/tests/test_execution.cairo
enitrat: legacy of a difficult old debugging session - forgot to delete it

crates/evm/src/interpreter.cairo
enitrat: state commitment is handled at the top-level in `eth_send_transaction`. 

`eth_send_transaction` = `eth_call` + state commitment

crates/evm/src/tests/test_execution.cairo
enitrat: reminescence of an old hard debugging session - forgot to remove this

crates/evm/src/execution.cairo
```diff
@@ -58,16 +63,60 @@ fn execute(
...
+                events: Default::default().span(),
+                state: machine.state,
+                error: Option::Some(err)
+            };
+        }
+    };
+    let transfer = Transfer { sender, recipient, amount: value };
+    match machine.state.add_transfer(transfer) {
+        Result::Ok(x) => x,
+        Result::Err(revert_reason) => machine.set_error(revert_reason),
```
Eikix, line 100: Should this early return the ExecutionResult?

crates/evm/src/tests.cairo
```diff
@@ -1,23 +1,25 @@
 #[cfg(test)]
-mod test_execution_context;
+mod test_execution;
+// #[cfg(test)]
+// mod test_execution_context;
 
-#[cfg(test)]
-mod test_instructions;
+// #[cfg(test)]
```
Eikix, line 6: we don't we test anymore?

crates/evm/src/tests/test_execution.cairo
```diff
@@ -0,0 +1,49 @@
...
+#[available_gas(20000000)]
+fn test_execute_value_transfer() {
+    let native_token = contract_utils::deploy_native_token();
+    // Transfer native tokens to sender
+    let kakarot_core = contract_utils::deploy_kakarot_core(native_token.contract_address);
+    set_contract_address(kakarot_core.contract_address);
+    let sender = EOATrait::deploy(evm_address()).expect('sender deploy failed');
+    let recipient = EOATrait::deploy(other_evm_address()).expect('recipient deploy failed');
+    // Transfer native tokens to sender
+    let native_token_dispatcher = IERC20CamelDispatcher {
```
Eikix, line 20: i think you already have the dispatcher on line 13

crates/evm/src/tests/test_execution.cairo
```diff
@@ -0,0 +1,49 @@
...
+        calldata: Default::default().span(),
+        value: 2000,
+        gas_price: 0,
+        gas_limit: 0,
+        read_only: false,
+    );
+    // `commit_state` is applied in `eth_send_tx` only - to test that `execute` worked correctly, we manually apply it here.
+    exec_result.state.commit_state();
+
+    let sender_balance = IERC20CamelDispatcher { contract_address: native_token.contract_address }
```
Eikix, line 40: you already have the dispatcher in line 13

crates/evm/src/execution.cairo
enitrat: modified

Eikix (commented):
a few comments and lgtm

Eikix (approved):
gg 👑👑👑

Issue #393, opened 2023-10-04
feat: implement STATICCALL opcode


Eikix:
## Feature

Use the generic_call util function from #391 and implement the staticcall opcode as per https://github.com/ethereum/execution-specs/blob/8b8c7c1a7437510f614272392275855de26209d3/src/ethereum/berlin/vm/instructions/system.py#L244

## Tests
Draw tests from geth, execution-specs, and kakarot cairo0

PR #508, opened 2023-11-02
Feat/delegate call

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #394 #393

## What is the new behavior?

- add delegate call 
- add staticcall
- ⚠️ rebased on staticcall PR (has its commits as well)
- make addresses in machine and execution context an Address struct rather than an evm address

- [ ] Yes
- [x] No

### DIFF
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -231,6 +231,9 @@ mod KakarotCore {
             if !self.is_view() {
                 panic_with_felt252('fn must be called, not invoked');
             };
+
+            let from = Address { evm: from, starknet: self.compute_starknet_address(from) };
+
             let result = self.handle_call(:from, :to, :gas_limit, :gas_price, :value, :data);
             match result {
                 Result::Ok(result) => result.return_data,
@@ -303,7 +306,7 @@ mod KakarotCore {
 
         fn handle_call(
             self: @ContractState,
-            from: EthAddress,
+            from: Address,
             to: Option<EthAddress>,
             gas_limit: u128,
             gas_price: u128,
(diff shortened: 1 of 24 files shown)


crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -316,6 +316,27 @@ mod KakarotCore {
                         Option::Some(account) => account.bytecode()?,
                         Option::None => Default::default().span(),
                     };
+
+                    // `from` address should be an EOA, i.e. a deployed Starknet account
+                    // checking invariants here:
```
Eikix, line 321: don't have a strong opinion about invariants here

crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -317,6 +317,27 @@ mod KakarotCore {
...
+                    // - `from` MUST not be an uninitialized account
+                    // - `from` MUST not be a contract account
+                    let from_starknet_address = match self.address_registry(from) {
```
enitrat, line 325: We should have these checks outside of `handle_call`

 - In `eth_call`, verify that `from` is a an address that is registered and has a corresponding starknet address
 - In `eth_send_transaction`, verify that `from` is the transaction origin, and that it's registered and has a corresponding SN address

enitrat: Handle call could therefore receive one extra parameter which would be `starknet_address`

crates/evm/src/execution.cairo
```diff
@@ -31,8 +31,8 @@ use utils::helpers::compute_starknet_address;
 /// *   The created contracts
 /// *   The events emitted
 fn execute(
-    origin: EthAddress,
-    target: EthAddress,
+    origin: Address,
+    address: Address,
```
enitrat, line 35: semantically I think `target` has more meaning, especially given that we have a lot of variables named `address` and I'm often a bit lost when trying to remember which one corresponds to what

crates/evm/src/execution.cairo
```diff
@@ -63,38 +63,7 @@ fn execute(
     // Initiate the Machine with the root context
     let mut machine: Machine = MachineCurrentContextTrait::new(ctx);
 
-    // Handle value transfers
-    let result_sender = machine.state.get_account(origin);
```
enitrat, line 67: Do you see any cases where this might cause a problem to not fetch the account from state first?

Eikix: afai see, the get_account is solely used to get the sender's Address struct, here we provide it before hand

enitrat: ok

enitrat (approved):
image

Issue #455, opened 2023-10-23
feat: implement CREATE2 opcode


Eikix:
Implement the CREATE opcode.

Check:
- KKT0 
- Current PR @ClementWalter from Clement to refactor the state in KKT0
- EELS: execution-specs https://github.com/ethereum/execution-specs/tree/master/src/ethereum/shanghai/vm
- edge cases

```[tasklist]
### Tasks
- [ ] Implement CREATE2 opcode
- [ ] Don't forget to increase the CA nonce for each CREATE2 invocation. Note that a PR from @enitrat will soon abstract the nonce storage changes in a Journal, or LocalStorage struct. CREATE can wait until then.
- [ ] Test that the correct address was computed using an example.
- [ ] test that the constructor is executed.
- [ ] check that the nonce of the creator is incremented.
- [ ] check that the nonce of the created account is set to 1
- [ ] check that in case of failure, 0 is returned to the stack and no state is changed
- [ ] check that in case of success, the address of the newly created account is returned to the stack
```

Eikix:
Help https://github.com/ethereum/execution-specs/blob/master/src/ethereum/shanghai/utils/address.py

Issue #454, opened 2023-10-23
feat: implement CREATE opcode


Eikix:

Eikix:
Blocked by RLP::encode I think @Quentash is on it

Eikix:
CREATE is implemented but still needs to be tested

Issue #453, opened 2023-10-23
epic: implement the create family of opcodes: CREATE and CREATE2


Eikix:
## Feature Request

Implement CREATE and CREATE2 opcodes.

```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/kakarot-ssj/issues/454
- [ ] https://github.com/kkrt-labs/kakarot-ssj/issues/455
``` 

### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

PR #510, opened 2023-11-02
feat: compute create 2 address helpers

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Starts: #454 #455 #453 

## What is the new behavior?

- creating create_helpers.cairo to set up a similar flow as call_helpers
- create2 helpers 
- create helpers are stuck until we have encoding

- [ ] Yes
- [x] No


crates/evm/src/create_helpers.cairo
```diff
@@ -0,0 +1,99 @@
...
+    fn prepare_create(ref self: Machine, create_type: CreateType) -> Result<CreateArgs, EVMError> {
+        let value = self.stack.pop()?;
+        let offset = self.stack.pop_usize()?;
+        let size = self.stack.pop_usize()?;
+
+        let mut bytecode = Default::default();
+        self.memory.load_n(size, ref bytecode, offset);
+
+        // TODO(state): when the tx starts, 
+        // store get_tx_info().unbox().nonce inside the sender account nonce
```
enitrat, line 47: What's the end goal here?
Do EOAs neeed to have an increased nonce SN nonce when deploying with create ?

crates/evm/src/create_helpers.cairo
```diff
@@ -0,0 +1,99 @@
...
+        ref self: Machine, sender_address: EthAddress, sender_nounce: felt252
+    ) -> Result<EthAddress, EVMError> {
+        panic_with_felt252('get_create_address todo')
+    }
+
+
+    fn get_create2_address(
+        self: @Machine, sender_address: EthAddress, salt: u256, bytecode: Span<u8>
+    ) -> Result<EthAddress, EVMError> {
+        let mut bytecode = ByteArrayExTrait::from_bytes(bytecode);
```
enitrat, line 72: once again: Why are we converting to ByteArray to create `Span<u64>` ? I've seen this in multiple places now, we need to open an issue to refactor this. In the meantime please add a TODO(optimize)

crates/evm/src/create_helpers.cairo
```diff
@@ -0,0 +1,99 @@
...
+        let salt = salt.to_bytes();
+
+        let mut preimage: Array<u8> = array![];
+
+        preimage.concat(array![0xff].span());
+        preimage.concat(sender_address);
+        preimage.concat(salt);
+        preimage.concat(hash);
+
+        let mut preimage = ByteArrayExTrait::from_bytes(preimage.span());
```
enitrat, line 89: same

enitrat (commented):
Generally speaking we should not have any interactions with ByteArray(Trait) except when loading/writing to storage. All the internals are using Span<u8> to express bytes, we don't want useless conversions

crates/evm/src/create_helpers.cairo
Eikix: it was just an idea to have the correct nonce at least for the caller of the tx inside its Account struct instead of nonce 1,
this is important for CREATE salt for instance

crates/evm/src/create_helpers.cairo
Eikix: https://github.com/kkrt-labs/kakarot-ssj/issues/516

crates/evm/src/create_helpers.cairo
enitrat: https://github.com/kkrt-labs/kakarot-ssj/issues/518

enitrat (approved):

Issue #496, opened 2023-11-01
feat: Implementation of rlp::encode()


Quentash:
### Describe the Feature Request

Kakarot needs to be able to RLP encode transactions. Implement rlp::encode() by referring to the Kakarot v0 implementation (https://github.com/kkrt-labs/kakarot/blob/main/src/utils/rlp.cairo) and thoroughly test it.

### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [X] Yes
- [ ] No

PR #513, opened 2023-11-03
Feat : Implementation of RLP::encode()

Quentash:
Implementation of the RLP encoding function.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #496 

## What is the new behavior?

- 
-
-

- [ ] Yes
- [X] No


Quentash:
I have a few questions/remarks about this PR :

1) Since rlp::encode() will simply call rlp::encode_string() when dealing with strings, can I modify already existing tests calling rlp::encode_string() to call rlp::encode() instead ? That will avoid too much duplicated tests.

2) I feel like the rlp::decode() function could make use of the ByteArray type for the sake of consistency. As well as the RLPType::String could be a ByteArray instead of a Span<u8>. What do you think about it ? Could it be done in this PR ?

3) I will add more tests but I first wanted to know about 1) so that's why this push is a bit early, sorry. I am also looking at a better way to implement the recursivity but since I saw that I was blocking other tickets, I wanted to, at least, propose a working solution.

crates/utils/src/rlp.cairo
```diff
@@ -62,6 +64,35 @@ impl RLPImpl of RLPTrait {
...
     }
 
+    /// RLP encodes multiple RLPItem
+    /// # Arguments
+    /// * `input` - Span of RLPItem to encode
+    /// # Returns
+    /// * `ByteArray - RLP encoded ByteArray
+    /// # Errors
+    /// * RLPError::RlpEmptyInput - if the input is empty
+    fn encode(mut input: Span<RLPItem>) -> Result<ByteArray, RLPError> {
```
bajpai244, line 74: I think the return type should be `Span<u8>`.

This will make it consistent with that [decode](https://github.com/kkrt-labs/kakarot-ssj/blob/862be44424dc665d26f532ebcb13706a9c94d90f/crates/utils/src/rlp.cairo#L109C5-L109C5) takes as input as of now.

bajpai244:
@Quentash I think the input type for `encode_string` can also be changed to `Span<u8>`, will add more consistency.

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -227,6 +238,30 @@ fn test_rlp_encode_large_bytearray_inputs() {
     }
 }
```
bajpai244, line 240: can we also add a test for the encoding `default` value?

crates/utils/src/tests/test_rlp.cairo
```diff
@@ -91,6 +91,17 @@ fn test_rlp_empty() {
     assert(res.unwrap_err() == RLPError::EmptyInput(RLP_EMPTY_INPUT), 'err != EmptyInput');
 }
 
+#[test]
+#[available_gas(20000000)]
+fn test_rlp_encode_empty_input() {
```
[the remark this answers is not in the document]
Quentash: I'm fine with your idea, I'm very bad at namind variables/functions ;)

crates/utils/src/rlp.cairo
Quentash: I was actually thinking about changing decode() so that it uses ByteArray as well, what is you preferred option ?

Quentash:
Pushed requested modifications.
One question persists tho : Should Decode() use ByteArray or should both encode() use Span<u8> ?

crates/utils/src/rlp.cairo
bajpai244: I think `Span<u8>` is better, as we also have a lot of utils related to it which we have made, which consider tx_data to be a `Span<u8>`.

bajpai244:
> Pushed requested modifications. One question persists tho : Should Decode() use ByteArray or should both encode() use Span ?

@Quentash I have replied for this, lemme know if u have some doubts :)

Eikix (approved):
lgtm

Issue #480, opened 2023-10-27
feat: make all account interactions local state-based


enitrat:
## Feature Request
The only interactions with storage must be done through the state trait. When retrieving account information for opcodes execution or any other operation, we must call the `StateTrait::get_account` function.

Blocked by #448
### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

PR #517, opened 2023-11-06
feat: integrate state

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

#480

## What is the new behavior?

- Integrates the state mechanism in existing opcodes and other places in code
- Fixes some access control issues in KakarotCore
-

- [ ] Yes
- [x] No


crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -304,6 +305,14 @@ mod KakarotCore {
             true
         }
 
+        /// Maps an EVM address to a Starknet address
+        /// Triggerred when deployment of an EOA or CA is successful
+        fn set_address_registry(
+            ref self: ContractState, evm_address: EthAddress, account: StoredAccountType
+        ) {
+            self.address_registry.write(evm_address, account);
```
Eikix, line 313: should we add only_owner

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -273,28 +238,23 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
...
-                let mut bytecode = ca.load_bytecode()?;
-                if bytecode.is_empty() {
-                    return self.stack.push(EMPTY_KECCAK);
-                }
-
-                let mut bytecode: ByteArray = ByteArrayExTrait::from_bytes(bytecode);
+        let account = self.state.get_account(evm_address)?;
+        if (account.is_precompile()
+            || (account.account_type.is_ca() && account.nonce == 0)
+            || account.is_selfdestruct()) {
```
Eikix, line 244: i think selfdestruct registers an account for destruction,
meaning if it calls selfdestruct and then extcodehash, it still has code (non-empty), only later will it be deleted

crates/evm/src/model/account.cairo
```diff
@@ -33,13 +34,19 @@ impl AccountImpl of AccountTrait {
                 }
             },
             Option::None => {
+                let kakarot_state = KakarotCore::unsafe_new_contract_state();
+                let starknet_address = compute_starknet_address(
```
Eikix, line 38: you don't want to call kakarot_state.compute_starknet_address?

crates/evm/src/model/account.cairo
```diff
@@ -51,6 +58,17 @@ impl AccountImpl of AccountTrait {
...
+    ///
+    /// # Arguments
+    /// * `self` - The account to commit
+    ///
+    /// # Returns
+    ///
+    /// `Ok(())` if the commit was successful, otherwise an `EVMError`.
     fn commit(self: @Account) -> Result<(), EVMError> {
         // Case account exists
         let maybe_acc = AccountImpl::account_type_at(self.address().evm)?;
```
Eikix, line 74: you already have access to self.account_type here no?

Eikix: we need to check that nonce != 0 to deploy the contract #519

crates/evm/src/model/eoa.cairo
```diff
@@ -24,7 +24,12 @@ struct EOA {
...
     /// Deploys a new EOA contract.
+    ///
+    /// # Arguments
+    ///
+    /// * `evm_address` - The EVM address of the EOA to deploy.
     fn deploy(evm_address: EthAddress) -> Result<EOA, EVMError> {
+        // Unlike CAs, there is not check for the existence of an EOA prealably to calling `EOATrait::deploy` - therefore, we need to check that there is no collision.
         let mut maybe_acc = AccountTrait::account_type_at(evm_address)?;
         if maybe_acc.is_some() {
             return Result::Err(EVMError::DeployError(EOA_EXISTS));
```
Eikix, line 35: Just noticed a redundant line (line 44) twice maybe_address (not your pr but can solve)

crates/evm/src/state.cairo
```diff
@@ -269,21 +268,22 @@ impl StateImpl of StateTrait {
...
-        match self.balances.read(address.evm.into()) {
+    fn read_balance(ref self: State, evm_address: EthAddress) -> Result<u256, EVMError> {
+        match self.balances.read(evm_address.into()) {
             Option::Some(value) => { Result::Ok(value) },
             Option::None => {
-                let account = AccountTrait::fetch_or_create(address.evm)?;
-                account.account_type.balance()
+                let account = AccountTrait::fetch_or_create(evm_address)?;
+                let balance = account.account_type.balance()?;
+                self.write_balance(evm_address.into(), balance);
```
Eikix, line 285: I don't think you need evm_address.into(), it's already the right type

crates/evm/src/model/account.cairo
enitrat: At first I wanted to avoid calling functions from KKT directly, but I guess it makes sense...

crates/evm/src/model/account.cairo
enitrat: the goal is to know whether there is an account deployed at this address. There is a fundamental problem in the current account model which is that `fetch_or_create` cannot know the type of the account we interact with if it's not deployed yet. I am thinking about remodelling everything so that we introduce an agnostic `address` field inside `

Eikix: Mmhm

Eikix (approved):
lgtm

Issue #509, opened 2023-11-02
dev: refactor test contracts deployment


enitrat:
- When we deploy any kind of account, we need deploy a `native_token` and the `kakarot_core` contract because the generic Account's `initialize` function reads from KakarotCore storage to approve spending of the native token.
- When we interact in any way with the Kakarot State, we need to `set_contract_address` the test runner to the kakarot contract.

This pattern is often used when testing some features that require access to the Starknet state:
```rust
    let native_token = deploy_native_token();
    let kakarot_core = deploy_kakarot_core(native_token.contract_address);
    testing::set_contract_address(kakarot_core.contract_address);
```

You will implement a test utils that performs these 3 operations in a single function. This util will then be called as

```rust
let (native_token, kakarot_core) = setup_contracts_for_testing()
```

akhercha:
Hey, happy to do this one!

PR #521, opened 2023-11-06
dev: Refactor test contracts deployment

akhercha:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #509 

## What is the new behavior?

- Created `setup_contracts_for_testing` inside `crates::contracts::tests::test_utils`,
- Replaced the contracts deployment by a call to `setup_contracts_for_testing` when applicable,
- Removed unused imports in some tests

- [ ] Yes
- [X] No

### DIFF
--- a/crates/contracts/src/tests/test_contract_account.cairo
+++ b/crates/contracts/src/tests/test_contract_account.cairo
@@ -1,9 +1,7 @@
 use contracts::contract_account::{IContractAccountDispatcherTrait, IContractAccountDispatcher};
-use contracts::kakarot_core::{interface::IExtendedKakarotCoreDispatcher, KakarotCore};
 use contracts::tests::test_data::counter_evm_bytecode;
 use contracts::tests::test_utils::{setup_contracts_for_testing, deploy_contract_account};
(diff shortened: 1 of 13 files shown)


akhercha:
Looks like the CI fails when running the `compare_snapshot.py` script:
```
Total gas change: 7930099612 -> 7930155072 (100.00%)
ERROR:__main__:Gas usage increased
```
([complete output here](https://gist.github.com/akhercha/30c556f3077f23c3a761a5de3e4461bc))
I think this is because now `testing::set_contract_address` is called whenever `setup_contracts_for_testing` is used; before some tests just deployed the contracts without using `testing::set_contract_address` right after.

Shall I add a condition in `setup_contracts_for_testing` so `set_contract_address` is just called when we want?
I'd say it's not a big deal and if we can ignore it & make the CI pass ; but can update if prefered!

enitrat:
@akhercha thanks for investigating, the CI failures are just informative so that we can check manually why gas usage increased / decreased, no big deal!

enitrat (approved):
💯

Issue #519, opened 2023-11-06
feat: stop deploying all undeployed accounts in state commit_accounts


Eikix:
### Describe the Feature Request

- check if the account exists before deploying (already in place)
- check if the account SHOULD be deployed, i.e. has been scheduled for deployment, or is simply "touched" (e.g. only transfer value). this can be done with the hasNonceOrCode check -> nonce != 0 OR code is not empty

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

enitrat:
Closed by #526

PR #526, opened 2023-11-07
refactor: account model

### LINKED ISSUE #518: feat: load SN Account nonce into origin EOA's cached account
### Describe the Feature Request
When instanciating the machine and its state, call `state.get_account` to cache the EOA of `origin` inside the state, and set its nonce to `get_tx_info().unbox().nonce`. This allows us to call `create` with the correct nonce as the salt.
### Additional Context

- [ ] Yes
- [ ] No

enitrat:
Refactors the account model and the way we interact with accounts to

- Make it safer by avoiding direct accesses to SN state
- Make it more consistent with the account model updates

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

## What is the new behavior?

- Prevents deployment of all undeployed accounts during commitment phase - First step for #519 is to check if the account must be deployed
- `StoredAccountType` is an internal type to the KakarotCore contract and is only used to check whether there is an account at a certain address, and what its actual type is. 
- `AccountType` methods disappeared in favour of `Account` methods. Eventually `AccountType` will disappear from the `Account` model - as we want to keep it as it's done in the EVM.
- Loads the Starknet Nonce of the origin into its evm account representation to correctly  call `create` with the correct nonce as the salt. Resolves #518

- [ ] Yes
- [x] No

### DIFF
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -9,7 +9,7 @@ const INVOKE_ETH_CALL_FORBIDDEN: felt252 = 'KKT: Cannot invoke eth_call';
 #[derive(Drop, starknet::Store, Serde, PartialEq, Default)]
 enum StoredAccountType {
     #[default]
-    UninitializedAccount,
+    UnexistingAccount,
     EOA: ContractAddress,
     ContractAccount: ContractAddress,
 }
@@ -25,7 +25,7 @@ mod KakarotCore {
     use evm::context::Status;
     use evm::errors::{EVMError, EVMErrorTrait, CALLING_FROM_CA, CALLING_FROM_UNDEPLOYED_ACCOUNT};
     use evm::execution::execute;
-    use evm::model::account::{Account, AccountTrait};
+    use evm::model::account::{Account, AccountType, AccountTrait};
     use evm::model::contract_account::{ContractAccount, ContractAccountTrait};
     use evm::model::eoa::{EOA, EOATrait};
     use evm::model::{ExecutionResult, Address};
@@ -167,8 +167,16 @@ mod KakarotCore {
             compute_starknet_address(deployer, evm_address, self.account_class_hash.read())
         }
 
-        fn address_registry(self: @ContractState, evm_address: EthAddress) -> StoredAccountType {
-            self.address_registry.read(evm_address)
+        fn address_registry(self: @ContractState, evm_address: EthAddress) -> Option<AccountType> {
+            match self.address_registry.read(evm_address) {
+                StoredAccountType::UnexistingAccount => Option::None,
+                StoredAccountType::EOA(starknet_address) => Option::Some(
+                    AccountType::EOA(EOA { evm_address, starknet_address })
+                ),
+                StoredAccountType::ContractAccount(starknet_address) => Option::Some(
+                    AccountType::ContractAccount(ContractAccount { evm_address, starknet_address })
+                ),
+            }
         }
 
 
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -28,7 +28,7 @@ mod KakarotCore {
     use evm::model::account::{Account, AccountType, AccountTrait};
     use evm::model::contract_account::{ContractAccount, ContractAccountTrait};
     use evm::model::eoa::{EOA, EOATrait};
-    use evm::model::{ExecutionResult, Address};
+    use evm::model::{ExecutionResult, Address, AddressTrait};
     use starknet::{
         EthAddress, ContractAddress, ClassHash, get_tx_info, get_contract_address, deploy_syscall
     };
@@ -188,11 +188,10 @@ mod KakarotCore {
         }
 
         fn account_balance(self: @ContractState, evm_address: EthAddress) -> u256 {
-            let maybe_account = AccountTrait::fetch(evm_address).expect('Fetching account failed');
-            match maybe_account {
-                Option::Some(account) => account.balance().unwrap(),
-                Option::None => 0
-            }
+            let address = Address {
+                evm: evm_address, starknet: self.compute_starknet_address(evm_address)
+            };
+            address.balance().unwrap()
         }
 
         fn contract_account_storage_at(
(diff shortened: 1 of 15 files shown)


crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -9,7 +9,7 @@ const INVOKE_ETH_CALL_FORBIDDEN: felt252 = 'KKT: Cannot invoke eth_call';
 #[derive(Drop, starknet::Store, Serde, PartialEq, Default)]
 enum StoredAccountType {
     #[default]
-    UninitializedAccount,
+    UnexistingAccount,
```
Eikix, line 12: nitpick: I believe unexisting is not a standard word,
could we go for:
- undeployed
- nonexistent
- void

enitrat: im so french

crates/evm/src/execution.cairo
```diff
@@ -78,6 +78,26 @@ fn execute(
         }
     }
 
+    // cache the EOA of origin inside the state, and set its nonce to get_tx_info().unbox().nonce.
+    // This allows us to call create with the correct nonce as the salt.
+    let mut origin_account = match machine.state.get_account(origin.evm) {
+        Result::Ok(account) => { account },
+        //TODO create helper method to avoid this boilerplace
+        Result::Err(err) => {
```
Eikix, line 86: Something to keep in mind:
as per the spec someone can call `eth_call` entrypoint, and simply input a random origin address, or even 0.

that being said, i believe get_account will simply return a default acct, if get_account points to a non existent acct, so all is fine

crates/evm/src/execution.cairo
```diff
@@ -78,6 +78,26 @@ fn execute(
...
+                return_data: Default::default().span(),
+                destroyed_contracts: Default::default().span(),
+                create_addresses: Default::default().span(),
+                events: Default::default().span(),
+                state: machine.state,
+                error: Option::Some(err)
+            };
+        }
+    };
+    origin_account.nonce = starknet::get_tx_info().unbox().nonce.try_into().unwrap();
```
Eikix, line 98: is this unwrap safe?

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -240,7 +240,7 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
         let evm_address = self.stack.pop_eth_address()?;
 
         let account = self.state.get_account(evm_address)?;
-        if (account.is_precompile() || (account.account_type.is_ca() && account.nonce == 0)) {
+        if (account.is_precompile() || (account.is_ca() && account.nonce == 0)) {
```
Eikix, line 243: maybe there is a better condition here for "void account?"

Eikix: before, every non existent acct was "casted" to CA in get_account, and if they had nonce == 0, they had to be non existent

crates/evm/src/model/account.cairo
```diff
@@ -68,40 +118,34 @@ impl AccountImpl of AccountTrait {
...
+                AccountType::EOA(eoa) => {
+                    // no - op
+                    Result::Ok(())
+                },
+                AccountType::ContractAccount(ca) => {
+                    let mut ca = *ca;
+                    if *self.selfdestruct {
+                        return ca.selfdestruct();
                     }
+                    ca.set_nonce(*self.nonce)
```
Eikix, line 134: should we set bytecode as well?

crates/evm/src/model/account.cairo
```diff
@@ -158,38 +211,20 @@ impl AccountImpl of AccountTrait {
...
-            AccountType::EOA => false,
-            AccountType::ContractAccount => true
-        }
-    }
-
     /// Returns the bytecode of the EVM account (EOA or CA)
-    fn bytecode(self: @AccountType) -> Result<Span<u8>, EVMError> {
-        match self {
+    #[inline(always)]
+    fn bytecode(self: @Account) -> Result<Span<u8>, EVMError> {
```
Eikix, line 226: can just access *self.code no?

crates/evm/src/model/eoa.cairo
```diff
@@ -87,13 +82,15 @@ impl EOAImpl of EOATrait {
     /// Returns an EOA instance from the given `evm_address`.
     fn at(evm_address: EthAddress) -> Result<Option<EOA>, EVMError> {
```
Eikix, line 83: Do we really need those two `at` methods?

crates/evm/src/execution.cairo
enitrat: Indeed, the error here can only be of type SyscallResult

crates/evm/src/instructions/environmental_information.cairo
enitrat: > before, every non existent acct was "casted" to CA in get_account, and if they had nonce == 0, they had to be non existent

1) What

I didn't get it

crates/evm/src/execution.cairo
enitrat: will a user ever send more than 18_446_744_073_709_551_615 transactions?

crates/evm/src/model/eoa.cairo
enitrat: They'll be removed next PR

crates/evm/src/instructions/environmental_information.cairo
Eikix: tl;dr -> i dont think we should use account.is_deployed, since its too expensive, but it still feels weird that we're checking the type of an acct in an opcode, we should just check if it "exists", regardless of its type

enitrat: > we should just check if it "exists", regardless of its type

The problem is that we can't check actual nonces of EOA.

Next PR I will work on this account type thing to determine whether we actually need to attach it to the account model or if we can just query it from the associated contracts if required. In KKT0 we do this, but by attaching it to the account model we save some gas by avoiding to perform useless calls on EOAs

Eikix: ok! fine with me

crates/evm/src/execution.cairo
Eikix: ☠️

Eikix (approved):
lgtm

enitrat:
@Eikix fixed check

Issue #406, opened 2023-10-06
feat: implement decode_tx() helper method for EOA


Eikix:
Implement decode_tx() helper method from Kakarot's EOA

Note that in 
https://github.com/kkrt-labs/kakarot-ssj/issues/276
We chose a new input format for the RPC, whereby we slightly transform eth_sendRawTransaction payload -> 
- RLP decode, extract R,S,V, put them in the signature field of the Starknet transaciton struct
- RLP encode the payload of tx without sig, put it in data input of SN tx

-> Changes from KKT0,
- We now use tx.signature for the signature

The steps the EOA has to do in validate are:
- Hash the RLP encoded payload
- Check the ECDSA signature of the user
- Accept the validate

execute has to:
- decode the transaction
- call kakarotcore.eth_send_transaction

Eikix:
@Quentash already started afaik

PR #529, opened 2023-11-09
feat: add eip-2718 tx decode

bajpai244:
## Pull Request type

PR adds support for decoding EIP-2718 txns + improves testing for legacy decoding of transaction.

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Decoding for EIP-2718 transactions is not supported.

Resolves: #406 

## What is the new behavior?

Decoding for EIP-2718 transactions is not supported.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -7,9 +7,9 @@ use utils::errors::RLPErrorTrait;
 
 use utils::errors::{EthTransactionError, RLPErrorImpl, RLPHelpersErrorImpl, RLPHelpersErrorTrait};
 use utils::helpers::ByteArrayExTrait;
+use utils::helpers::U256Trait;
 
 use utils::helpers::{U256Impl, ByteArrayExt};
-use utils::helpers::{U256Trait, compute_msg_hash};
 
 use utils::rlp::RLPItem;
 use utils::rlp::{RLPTrait, RLPHelpersTrait};
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -1,3 +1,4 @@
+use core::array::SpanTrait;
 use core::option::OptionTrait;
 use core::traits::TryInto;
 
@@ -128,9 +129,12 @@ impl EthTransactionImpl of EthTransaction {
             RLPItem::String => { Result::Err(EthTransactionError::ExpectedRLPItemToBeList) },
             RLPItem::List(val) => {
                 // total items in EIP 2930 (unsigned): 8
+                if (tx_type == 1 && val.len() != 8) {
+                    return Result::Err(EthTransactionError::Other('Length is not 8'));
+                }
                 // total items in EIP 1559 (unsigned): 9
-                if (val.len() != 8 && val.len() != 9) {
-                    return Result::Err(EthTransactionError::Other('Length is not 8 or 9'));
+                if (tx_type == 2 && val.len() != 9) {
+                    return Result::Err(EthTransactionError::Other('Length is not 9'));
                 }
 
                 let chain_id = (*val.at(chain_idx)).parse_u128_from_string().map_err()?;
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -144,7 +144,7 @@ impl EthTransactionImpl of EthTransaction {
         let result: Result<EthereumTransaction, EthTransactionError> = match decoded_data {
             RLPItem::String => { Result::Err(EthTransactionError::ExpectedRLPItemToBeList) },
             RLPItem::List(val) => {
-                // total items in EIP 2930 (unsgined): 8
+                // total items in EIP 2930 (unsigned): 8
                 // total items in EIP 1559 (unsigned): 9
                 if (val.len() != 8 && val.len() != 9) {
                     return Result::Err(EthTransactionError::Other('Length is not 8 or 9'));
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -93,7 +93,7 @@ impl EthTransactionImpl of EthTransaction {
     /// transaction data, which includes the chain ID as part of the transaction data itself.
     /// # Arguments
     /// tx_data The raw transaction data
-    fn decode_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
+    fn decode_typed_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
         let tx_type: u32 = (*tx_data.at(0)).into();
         let rlp_encoded_data = tx_data.slice(1, tx_data.len() - 1);
 
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -23,7 +23,6 @@ struct EthereumTransaction {
     amount: u256,
     calldata: Span<u8>,
     chain_id: u128,
-    msg_hash: u256,
 }
 
 #[generate_trait]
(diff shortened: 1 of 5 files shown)


crates/utils/src/eth_transaction.cairo
```diff
@@ -111,8 +111,78 @@ impl EthTransactionImpl of EthTransaction {
...
+            return Result::Err(EthTransactionError::Other('Not EIP-2718 transaction'));
+        }
+        // Only EIP-1559 and EIP 2930 are supported
+        if (tx_type != 1 && tx_type != 2) {
+            return Result::Err(EthTransactionError::Other('transaction type not supported'));
+        }
+
+        let chain_idx = 0;
+        let nonce_idx = 1;
+        let gas_price_idx = tx_type + nonce_idx;
```
Eikix, line 129: can you add as doc comments the format of the tx?
like `// EIP1559 RLP[chain_id, nonce, ...]` and 2930

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,8 +111,78 @@ impl EthTransactionImpl of EthTransaction {
...
+        if (decoded_data.len() != 1) {
+            return Result::Err(EthTransactionError::Other('Length is not 1'));
+        }
+
+        let decoded_data = *decoded_data.at(0);
+
+        let result: Result<EthereumTransaction, EthTransactionError> = match decoded_data {
+            RLPItem::String => { Result::Err(EthTransactionError::ExpectedRLPItemToBeList) },
+            RLPItem::List(val) => {
+                // total items in EIP 2930 (unsgined): 8
```
Eikix, line 145: typo

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,8 +111,78 @@ impl EthTransactionImpl of EthTransaction {
...
+                // total items in EIP 1559 (unsigned): 9
+                if (val.len() != 8 && val.len() != 9) {
```
Eikix, line 147: are you handling the edge case where the tx type is EIP 1559 and the val.len() is 8?

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,8 +111,78 @@ impl EthTransactionImpl of EthTransaction {
...
+
+                let chain_id = (*val.at(chain_idx)).parse_u128_from_string().map_err()?;
+                let nonce = (*val.at(nonce_idx)).parse_u128_from_string().map_err()?;
+                let gas_price = (*val.at(gas_price_idx)).parse_u128_from_string().map_err()?;
+                let gas_limit = (*val.at(gas_limit_idx)).parse_u128_from_string().map_err()?;
+                let to = (*val.at(to_idx)).parse_u256_from_string().map_err()?;
+                let amount = (*val.at(value_idx)).parse_u256_from_string().map_err()?;
+                let calldata = (*val.at(calldata_idx)).parse_bytes_from_string().map_err()?;
+
+                let mut transaction_data_byte_array = ByteArrayExt::from_bytes(tx_data);
```
Eikix, line 159: this is a rough context switch from what the function is doing,
I think the decode_tx should do one thing to avoid such surprises,
it's worth encapsulating this logic into a compute_msg_hash function

Eikix: also, are you sure it's worth it doing all those things at once?
we might as well first compute the msg hash, check it,
then decode the tx, check chain id 
then end validate?

do you think it's saving steps to do both at the same time?

Eikix: this solution would give us an early return route, and separate concerns further

crates/utils/src/eth_transaction.cairo
bajpai244: makes sense!

crates/utils/src/eth_transaction.cairo
bajpai244: By definition, it is incorrect, and no one should sign a transaction with tx_type 2 but with less fields, I am not sure how can one handle this, but lemme know if u have some idea.

crates/utils/src/eth_transaction.cairo
bajpai244: fixed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/529/commits/df5d5e58d7e74fbfc83ecac5f9112deff2034ed4) commit.

crates/utils/src/eth_transaction.cairo
Eikix: I guess one resulting consequence of my suggestion (yes you're correct in your guess), is that the msg_hash would no longer be a field of the EthereumTransaction, but rather handled alone "first", and then the eth tx would be decoded and handled.

bajpai244: Gotcha, yeah, that sounds great to me!

We can put this directly in validate!

bajpai244: Lemme refactor.

bajpai244: removed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/529/commits/136314cfdae800f56296173b4135319fb9cec4db) commit.

Also, we can keep the util to compute the message hash in this PR, I will use it with the validate PR.

crates/utils/src/helpers.cairo
```diff
@@ -870,6 +870,16 @@ fn compute_starknet_address(
     normalized_address
 }
 
+// keccack256 on a bytes message
+fn compute_msg_hash(msg: Span<u8>) -> u256 {
```
Eikix, line 874: nice! can you:
- make it part of the SpanExtTrait (if it exists, or create a Span extension trait otherwise),
- make msg a `self` argument

and create an issue for using this util everywhere we hash an array of bytes?
there are multiple parts of the codebase where we do exactly these lines:) would help reduce redundant loc and can be a good first issue for new comers

crates/utils/src/eth_transaction.cairo
bajpai244: @Eikix I don't think u can make some malform EIP-1559 to EIP-2930, as the user signs over the RLP of these transactions, which will be different values.

So, if someone signs 1559, then no one can replay it as an EIP-2930 because the digital signature won't allow this, lemme know if this makes sense.

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,8 +94,70 @@ impl EthTransactionImpl of EthTransaction {
     /// # Arguments
     /// tx_data The raw transaction data
     fn decode_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
```
Eikix, line 96: let's rename it decode_typed_tx

crates/utils/src/helpers.cairo
bajpai244: SpanExtension already exists, but it for generic spans which sounds correct to me.

I have created a BytesTrait for this and refactored via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/529/commits/fd1db8d4c85ce35923115b2876eccfe101c533ed) commit.

crates/utils/src/eth_transaction.cairo
bajpai244: done via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/529/commits/f3dffe655678b3442eb8a4a4adb4201b1c6998c2) commit.

crates/utils/src/eth_transaction.cairo
bajpai244: I have a simple solution for this! Lemme push it.

bajpai244: [This](https://github.com/kkrt-labs/kakarot-ssj/pull/529/commits/3d97db04e9eec07cfbbbc74a2c6721a4eb058d8c) commit adds specific length check depending on the type of transaction!

This way what u described above should be restricted! Lemme know if this works :)

Eikix (approved):
lgtm

Issue #530, opened 2023-11-09
feat: implement `decode` and `is_legacy_tx`


bajpai244:
### Describe the Feature Request

Implement `decode` and `is_legacy_tx` for [EthTransaction](https://github.com/kkrt-labs/kakarot-ssj/blob/832f2132ca4a4f041ecaed2191796775fb00de97/crates/utils/src/eth_transaction.cairo#L30) utils.

- [x] Yes
- [ ] No

PR #531, opened 2023-11-09
Feat: implement `decode` and is `legacy tx`

bajpai244:
## Pull Request type

Implement `decode` and is `legacy tx` for [EthTransaction](implement `decode` and is `legacy tx`) utlils.

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

These methods are not implemented.

Resolves: #530 

## What is the new behavior?

-` decode` and is `legacy tx` are implemented

- [x] Yes
- [ ] No

### DIFF
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -42,7 +42,9 @@ impl EthTransactionImpl of EthTransaction {
     /// # Arguments
     /// tx_data The raw transaction data
     /// tx_data is of the format: rlp![nonce, gasPrice, gasLimit, to , value, data, chainId, 0, 0]
-    fn decode_legacy_tx(tx_data: EncodedLegacyTransaction) -> Result<EthereumTransaction, EthTransactionError> {
+    fn decode_legacy_tx(
+        tx_data: EncodedLegacyTransaction
+    ) -> Result<EthereumTransaction, EthTransactionError> {
         let decoded_data = RLPTrait::decode(tx_data);
         let decoded_data = decoded_data.map_err()?;
 
@@ -100,7 +102,9 @@ impl EthTransactionImpl of EthTransaction {
     /// transaction data, which includes the chain ID as part of the transaction data itself.
     /// # Arguments
     /// tx_data The raw transaction data
-    fn decode_typed_tx(tx_data: EncodedTypedTransaction) -> Result<EthereumTransaction, EthTransactionError> {
+    fn decode_typed_tx(
+        tx_data: EncodedTypedTransaction
+    ) -> Result<EthereumTransaction, EthTransactionError> {
         let tx_type: u32 = (*tx_data.at(0)).into();
         let rlp_encoded_data = tx_data.slice(1, tx_data.len() - 1);
 
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -37,6 +37,10 @@ impl EthTransactionImpl of EthTransaction {
     /// tx_data The raw transaction data
     /// tx_data is of the format: rlp![nonce, gasPrice, gasLimit, to , value, data, chainId, 0, 0]
     fn decode_legacy_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
+        if (!EthTransaction::is_legacy_tx(tx_data)) {
+            return Result::Err(EthTransactionError::Other('Not legacy transaction'));
+        }
+
         let decoded_data = RLPTrait::decode(tx_data);
         let decoded_data = decoded_data.map_err()?;
 
@@ -98,11 +102,6 @@ impl EthTransactionImpl of EthTransaction {
         let tx_type: u32 = (*tx_data.at(0)).into();
         let rlp_encoded_data = tx_data.slice(1, tx_data.len() - 1);
 
-        // EIP 2718:
-        // TransactionType is a positive unsigned 8-bit number between 0 and 0x7f
-        if (tx_type == 0 || tx_type >= 0x7f) {
-            return Result::Err(EthTransactionError::Other('Not EIP-2718 transaction'));
-        }
         // Only EIP-1559 and EIP 2930 are supported
         if (tx_type != 1 && tx_type != 2) {
             return Result::Err(EthTransactionError::Other('transaction type not supported'));
@@ -169,9 +168,15 @@ impl EthTransactionImpl of EthTransaction {
     /// according to EIP-2718. If the transaction type is less than or equal to 0xc0, it's a legacy transaction.
     /// # Arguments
     /// - `tx_data` The raw transaction data
-    fn is_legacy_tx(tx_data: Span<u8>) -> Result<bool, EthTransactionError> {
-        // todo
-        panic_with_felt252('is_legacy_tx unimplemented')
+    fn is_legacy_tx(tx_data: Span<u8>) -> bool {
+        let tx_type = *tx_data[0];
+
+        // TransactionType for EIP-2718 txns is a positive unsigned 8-bit number between 0 and 0x7f
+        if (tx_type >= 0x7f) {
+            return true;
+        }
+
+        return false;
     }
 
     /// Decode a raw Ethereum transaction
@@ -181,8 +186,11 @@ impl EthTransactionImpl of EthTransaction {
     /// # Arguments
     /// - `tx_data` The raw transaction data
     fn decode(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
-        // todo
-        panic_with_felt252('decode unimplemented')
+        if (EthTransaction::is_legacy_tx(tx_data)) {
+            EthTransaction::decode_legacy_tx(tx_data)
+        } else {
+            EthTransaction::decode_typed_tx(tx_data)
+        }
     }
 
     /// Validate an Ethereum transaction
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -49,7 +49,7 @@ impl EthTransactionImpl of EthTransaction {
         let decoded_data = decoded_data.map_err()?;
 
         if (decoded_data.len() != 1) {
-            return Result::Err(EthTransactionError::Other('Length is not 1'));
+            return Result::Err(EthTransactionError::TopLevelRlpListWrongLength(decoded_data.len()));
         }
 
         let decoded_data = *decoded_data.at(0);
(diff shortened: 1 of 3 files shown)


bajpai244:
## Note

I have opened this RP just for review purpose and hence in draft, it builds on top of work done in the open PR  #529.

We firstly have to merge #529 to merge this, and hence some changes that we see here apart from the scope of the work comes from the work on that PR.

Hence flow to merge:
- merge #529
- rebase the branch of this PR

crates/utils/src/eth_transaction.cairo
```diff
@@ -37,6 +37,10 @@ impl EthTransactionImpl of EthTransaction {
     /// tx_data The raw transaction data
     /// tx_data is of the format: rlp![nonce, gasPrice, gasLimit, to , value, data, chainId, 0, 0]
     fn decode_legacy_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
+        if (!EthTransaction::is_legacy_tx(tx_data)) {
```
Eikix, line 40: I think this check can be done by the type system at compile time
My advice: create an enum `EncodedTransaction` : 

```
enum EncodedTransaction {
    Legacy: Span<u8>,
    Eip2718: Span<u8>,
}
```

Eikix: Lmk if this works, haven't tested this out

Eikix: UPDATE/ERRATUM: I think this is bad design actually, you don't want to match the variants inside the body of this function, it's nonsensical.

The option to use a type alias seems better:
```
type EncodedLegacyTx = Span<u8>;
```

And then use this type in the entry of your function,
And make sure you only convert to this type when is_legacy_tx has been checked

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,18 +115,93 @@ impl EthTransactionImpl of EthTransaction {
     /// # Arguments
     /// tx_data The raw transaction data
     fn decode_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
```
Eikix, line 117: should this be named decode_2718_tx? or decode_non_legacy_tx?

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,18 +115,93 @@ impl EthTransactionImpl of EthTransaction {
...
-        // todo
-        panic_with_felt252('decode_tx unimplemented')
+        if (EthTransaction::is_legacy_tx(tx_data)) {
+            return Result::Err(EthTransactionError::Other('Not EIP-2718 transaction'));
```
Eikix, line 119: Again, here I think it'd be better to type the entry arg, to not run these runtime checks

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,18 +115,93 @@ impl EthTransactionImpl of EthTransaction {
...
+        }
+
+        let tx_type: u32 = (*tx_data.at(0)).into();
```
Eikix, line 122: although very unlikely, we should check that the payload is the right size, i.e. at least of len 2 for instance?

this check could be put into the Span<u8> to_eip_2718_tx trait function, which, would simply check if the tx is malformed (len < 2 or some other length), if the tx is not legacy, etc.

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,18 +115,93 @@ impl EthTransactionImpl of EthTransaction {
...
+        let rlp_encoded_data = tx_data.slice(1, tx_data.len() - 1);
+
+        // Only EIP-1559 and EIP 2930 are supported
+        if (tx_type != 1 && tx_type != 2) {
```
Eikix, line 126: this can also be put inside to_eip_2718_tx fn

crates/utils/src/eth_transaction.cairo
```diff
@@ -111,18 +115,93 @@ impl EthTransactionImpl of EthTransaction {
...
+        let nonce_idx = 1;
+        let gas_price_idx = tx_type + nonce_idx;
+        let gas_limit_idx = gas_price_idx + 1;
+        let to_idx = gas_limit_idx + 1;
+        let value_idx = to_idx + 1;
+        let calldata_idx = value_idx + 1;
+
+        let decoded_data = RLPTrait::decode(rlp_encoded_data).map_err()?;
+        if (decoded_data.len() != 1) {
+            return Result::Err(EthTransactionError::Other('Length is not 1'));
```
Eikix, line 140: I don't think this should be an Other error

crates/utils/src/eth_transaction.cairo
bajpai244: we have rename this to `decode_typed_tx`.

crates/utils/src/eth_transaction.cairo
bajpai244: This sounds like an interesting approach. Lemme refactor.

crates/utils/src/eth_transaction.cairo
bajpai244: So, from my understanding, we can not avoid runtime check at all, we can just make sure they are not redundant, so I think decode should do the runtime check { otherwise the function using it would have to do it! }

I am using type alias for these tx types for:
- `decode_legacy_tx`
- `decode_typed_tx`

and, that should help make sure that they don't do check for transaction types inside them.

Hence, decode does the runtime checks and make sure that the `tx_data` is only being passed to the other functions once it has been guaranteed that they are of a certain transaction type.

Other functions hence can rely on this guarantee and avoid and redundant check.

crates/utils/src/eth_transaction.cairo
bajpai244: So, this check is different from checking wether a transaction is EIP-2718 type or not, this checks for the fact that whether we support this tx_type or not, and I think this check should be something that this function should handle.

Lemme know what u think!

crates/utils/src/eth_transaction.cairo
bajpai244: Implemented via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/531/commits/468716b8efbc48c599fd75687c38f76275150319) commit.

crates/utils/src/eth_transaction.cairo
bajpai244: In the `decode_typed_tx` we are already checking for correct length of the payload.

image

crates/utils/src/eth_transaction.cairo
bajpai244: created enum variants by [this](https://github.com/kkrt-labs/kakarot-ssj/pull/531/commits/932f1ff1919a87744c94d979beeb2af022c8fc0d) commit.

crates/utils/src/eth_transaction.cairo
```diff
@@ -169,9 +174,15 @@ impl EthTransactionImpl of EthTransaction {
...
     /// # Arguments
     /// - `tx_data` The raw transaction data
-    fn is_legacy_tx(tx_data: Span<u8>) -> Result<bool, EthTransactionError> {
-        // todo
-        panic_with_felt252('is_legacy_tx unimplemented')
+    fn is_legacy_tx(tx_data: Span<u8>) -> bool {
+        let tx_type = *tx_data[0];
+
+        // TransactionType for EIP-2718 txns is a positive unsigned 8-bit number between 0 and 0x7f
+        if (tx_type >= 0x7f) {
```
Eikix, line 181: From: https://eips.ethereum.org/EIPS/eip-2718

> For the forseable future, 0x7f is plenty and it leaves open a number of options for extending the range such as using the high bit as a continuation bit. This also prevents us from colliding with legacy transaction types, which always start with a byte >= 0xc0.

I think this should be strict inequality

crates/utils/src/eth_transaction.cairo
```diff
@@ -26,6 +26,12 @@ struct EthereumTransaction {
     chain_id: u128,
 }
 
+
+/// represents an encoded legacy transaction, a Span<u8> should only best casted to this type if check for it being a legacy transaction has been done.
+type EncodedLegacyTransaction = Span<u8>;
```
Eikix, line 31: Mhm this seems as though it doesn't work as expected. This doesn't add any protection from the type system

crates/utils/src/eth_transaction.cairo
```diff
@@ -173,12 +199,13 @@ impl EthTransactionImpl of EthTransaction {
...
     /// # Arguments
-    /// - `tx_data` The raw transaction data
+    /// * `tx_data` - The raw transaction data
+    #[inline(always)]
     fn is_legacy_tx(tx_data: Span<u8>) -> bool {
         let tx_type = *tx_data[0];
 
         // TransactionType for EIP-2718 txns is a positive unsigned 8-bit number between 0 and 0x7f
-        if (tx_type >= 0x7f) {
+        if (tx_type > 0x7f) {
```
bajpai244, line 208: nice catch.

bajpai244: wait, I guess this was correct.

tx_type is (0,0x7f) if I am correct, hence the byte can be 0x7f.

crates/utils/src/eth_transaction.cairo
```diff
@@ -26,22 +27,64 @@ struct EthereumTransaction {
...
     /// This function decodes a legacy Ethereum transaction in accordance with EIP-155.
     /// It returns transaction details including nonce, gas price, gas limit, destination address, amount, payload,
     /// message hash, chain id. The transaction hash is computed by keccak hashing the signed
     /// transaction data, which includes the chain ID in accordance with EIP-155.
     /// # Arguments
-    /// tx_data The raw transaction data
-    /// tx_data is of the format: rlp![nonce, gasPrice, gasLimit, to , value, data, chainId, 0, 0]
+    /// * tx_data - The raw transaction data
+    /// * tx_data - is of the format: rlp![nonce, gasPrice, gasLimit, to , value, data, chainId, 0, 0]
     fn decode_legacy_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
```
bajpai244, line 82: We should add a note in the comment, that function should be called only after verification of transaction type to be legacy.

crates/utils/src/eth_transaction.cairo
```diff
@@ -93,21 +136,11 @@ impl EthTransactionImpl of EthTransaction {
     /// message hash, and chain id. The transaction hash is computed by keccak hashing the signed
     /// transaction data, which includes the chain ID as part of the transaction data itself.
     /// # Arguments
-    /// tx_data The raw transaction data
+    /// * `tx_data` - The raw transaction data
     fn decode_typed_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
```
bajpai244, line 140: same here

crates/utils/src/eth_transaction.cairo
Eikix: good catch

Eikix (approved):
gggg

PR #533, opened 2023-11-09
Refactor/account model

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #480

## What is the new behavior?

- Removed `EOA` and `CA` types - methods kept now operate on the `Address` type.
- AccoutType variants don't take values. Introduced `Unknown` variant, for accounts that are touched but whose actual type is not known yet. They can be undeployed EOAs or undeployed CAs. If an `Unknown` account is deployed using create opcodes or in a deploy tx, its type is manually set to `ContractAccount`.
- Simplifies interactions with accounts overall. Less confusing methods.
- Account model is closer to the actual EVM model - the account type is used for optimisation.
- Fix selfdestruct by removing the entry from the kakarot address_registry upon commitment.
- Fix extcodehash selfdestruct test by expecting an empty hash instead of 0 (account still exists until TX ends!)

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -101,7 +101,7 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
         }
 
         target_account.set_nonce(1);
-        target_account.account_type = AccountType::ContractAccount;
+        target_account.set_type(AccountType::ContractAccount);
         target_account
             .address = Address { evm: target_address.evm, starknet: target_address.starknet };
         self.state.set_account(target_account);
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -165,13 +165,6 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
             },
             // Failure
             Status::Reverted => {
-                //TODO(accounts) (ENSURE THIS IS CORRECT)
-                // Reverts changes brought to cache in `init_create_sub_ctx`
-                // by reverting the nonce and the type
-                let mut account = self.state.get_account(account_address)?;
-                account.set_nonce(0);
-                account.set_type(AccountType::Unknown);
-                self.state.set_account(account);
                 self.return_to_parent_ctx();
                 self.stack.push(0)
             },
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -161,7 +161,6 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
                     account.account_type == AccountType::ContractAccount,
                     'type should be CA in finalize'
                 );
-                account.set_type(AccountType::ContractAccount);
                 self.state.set_account(account);
                 self.return_to_parent_ctx();
                 self.stack.push(account_address.into())
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -104,8 +104,7 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
 
         target_account.set_nonce(1);
         target_account.set_type(AccountType::ContractAccount);
-        target_account
-            .address = Address { evm: target_address.evm, starknet: target_address.starknet };
+        target_account.address = target_address;
         self.state.set_account(target_account);
 
         let call_ctx = CallContextTrait::new(
@@ -158,6 +157,10 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
 
                 let mut account = self.state.get_account(account_address)?;
                 account.set_code(return_data);
+                assert(
+                    account.account_type == AccountType::ContractAccount,
+                    'type should be CA in finalize'
+                );
                 account.set_type(AccountType::ContractAccount);
                 self.state.set_account(account);
                 self.return_to_parent_ctx();
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -8,8 +8,8 @@ use evm::errors::{EVMError, CALL_GAS_GT_GAS_LIMIT, ACTIVE_MACHINE_STATE_IN_CALL_
 use evm::machine::{Machine, MachineCurrentContextTrait};
 use evm::memory::MemoryTrait;
 use evm::model::account::{AccountTrait};
-use evm::model::contract_account::{ContractAccount, ContractAccountTrait};
-use evm::model::{AccountType, Transfer};
+use evm::model::contract_account::{ContractAccountTrait};
+use evm::model::{Address, AccountType, Transfer};
 use evm::stack::StackTrait;
 use evm::state::StateTrait;
 use keccak::cairo_keccak;
@@ -46,7 +46,7 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
         let mut bytecode = Default::default();
         self.memory.load_n(size, ref bytecode, offset);
 
-        // TODO(state): when the tx starts, 
+        // TODO(state): when the tx starts,
         // store get_tx_info().unbox().nonce inside the sender account nonce
         let sender_nonce = get_tx_info().unbox().nonce;
 
@@ -101,13 +101,9 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
         }
 
         target_account.set_nonce(1);
+        target_account.account_type = AccountType::ContractAccount;
         target_account
-            .account_type =
-                AccountType::ContractAccount(
-                    ContractAccount {
-                        evm_address: target_address.evm, starknet_address: target_address.starknet
-                    }
-                );
+            .address = Address { evm: target_address.evm, starknet: target_address.starknet };
         self.state.set_account(target_account);
 
         let call_ctx = CallContextTrait::new(
@@ -160,6 +156,7 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
 
                 let mut account = self.state.get_account(account_address)?;
                 account.set_code(return_data);
+                account.set_type(AccountType::ContractAccount);
                 self.state.set_account(account);
                 self.return_to_parent_ctx();
                 self.stack.push(account_address.into())
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -94,11 +94,13 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
         caller_account.set_nonce(caller_current_nonce + 1);
         self.state.set_account(caller_account);
 
-        // Check collision
-        let is_collision = target_account.nonce() != 0 || !target_account.code.is_empty();
-        if is_collision {
+        // Collision happens if a
+        // - contract is already deployed at this location (type fetched from storage)
+        // - Contract has been scheduled for deployment (type set in cache)
+        // If the AccountType is unknown, then there's no collision.
+        if target_account.is_deployed() {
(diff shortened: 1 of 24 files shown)


crates/evm/src/create_helpers.cairo
```diff
@@ -94,20 +94,18 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
...
         target_account.set_nonce(1);
+        target_account.set_type(AccountType::ContractAccount);
         target_account
-            .account_type =
-                AccountType::ContractAccount(
-                    ContractAccount {
-                        evm_address: target_address.evm, starknet_address: target_address.starknet
-                    }
-                );
+            .address = Address { evm: target_address.evm, starknet: target_address.starknet };
```
Eikix, line 108: target_account.address = target_address works?

crates/evm/src/create_helpers.cairo
```diff
@@ -160,6 +158,7 @@ impl MachineCreateHelpersImpl of MachineCreateHelpers {
 
                 let mut account = self.state.get_account(account_address)?;
                 account.set_code(return_data);
+                account.set_type(AccountType::ContractAccount);
```
Eikix, line 161: wasn't this set previously?

Eikix: we may want to add an invariant, such as 
if account.type != AccountType::ContractAccount { return Err}

crates/evm/src/model/account.cairo
```diff
@@ -65,41 +68,46 @@ impl AccountImpl of AccountTrait {
...
+                    AccountType::ContractAccount => {
+                        let address = Address { evm: evm_address, starknet: starknet_address };
                         Option::Some(
                             Account {
-                                account_type: AccountType::ContractAccount(ca),
-                                code: ca.load_bytecode()?,
-                                nonce: ca.nonce()?,
+                                account_type: AccountType::ContractAccount,
+                                address,
+                                code: ContractAccountTrait::fetch_bytecode(@address)?,
```
Eikix, line 90: why do we need to snap address

crates/evm/src/model/account.cairo
```diff
@@ -65,41 +68,46 @@ impl AccountImpl of AccountTrait {
...
+                                nonce: ContractAccountTrait::fetch_nonce(@address)?,
                                 selfdestruct: false,
                             }
                         )
-                    }
+                    },
+                    AccountType::Unknown(_) => Option::None,
```
Eikix, line 96:
```suggestion
                    AccountType::Unknown => Option::None,
```

crates/evm/src/model/account.cairo
```diff
@@ -65,41 +68,46 @@ impl AccountImpl of AccountTrait {
...
-    // If the nonce is 0, the account is just "touched" (e.g.
-    // balance transfer) and is not set for deployment.
+    /// Returns whether an account should be deployed or not.
+    /// To be deployed, the type must have been resolved to be a CA - must not
+    /// be registered already, and the nonce must not be 0 or the code must not
+    /// be empty
     #[inline(always)]
-    fn should_deploy(self: @Account) -> bool {
-        if self.is_ca() && (*self.nonce != 0 || !(*self.code).is_empty()) {
+    fn should_deploy(self: @Account, is_registered: bool) -> bool {
```
Eikix, line 109: feels bad to add this here,
imo self.AccountType == Unknown <=> is_registered

crates/evm/src/model/account.cairo
```diff
@@ -117,32 +125,38 @@ impl AccountImpl of AccountTrait {
...
+        } else if self.should_deploy(is_registered) {
             //Case new account
             // If SELFDESTRUCT, just do nothing
             if (*self.selfdestruct == true) {
                 return Result::Ok(());
             };
-            let mut ca = ContractAccountTrait::deploy(self.address().evm, *self.code)?;
-            ca.set_nonce(*self.nonce)
+            let mut ca_address = ContractAccountTrait::deploy(self.address().evm, *self.code)?;
+            ca_address.store_nonce(*self.nonce)
```
Eikix, line 159: Isn't this already done in ContractAccountTrait::deploy ?

Eikix, line 159: to save storage syscalls, should we check *self.nonce > 1?

crates/evm/src/model/contract_account.cairo
```diff
@@ -126,94 +115,44 @@ impl ContractAccountImpl of ContractAccountTrait {
...
-    /// * `self` - The contract account instance
+    /// * `self` - The address of the Contract Account
     /// * `key` - The key to set
     /// * `value` - The value to set
     #[inline(always)]
-    fn set_storage_at(ref self: ContractAccount, key: u256, value: u256) -> Result<(), EVMError> {
-        let mut contract_account = IContractAccountDispatcher {
-            contract_address: self.starknet_address
-        };
+    fn store_storage(self: @Address, key: u256, value: u256) -> Result<(), EVMError> {
```
Eikix, line 154: Is this not type-unsafe to relax the type constraints on self?

Eikix: it's reasonable to think that if we've come to be forced into this, then we've missed something in the design of our types

crates/evm/src/model/account.cairo
Eikix: ah, this amounts to .is_deployed() no?

crates/evm/src/model/account.cairo
```diff
@@ -176,49 +189,45 @@ impl AccountImpl of AccountTrait {
...
-    fn is_deployed(address: EthAddress) -> bool {
-        let mut kakarot_state = KakarotCore::unsafe_new_contract_state();
-        let maybe_account = kakarot_state.address_registry(address);
-        match maybe_account {
-            Option::Some(_) => true,
-            Option::None => false
-        }
-    }
+    fn is_deployed(self: @Account) -> bool {
+        let is_known = *self.account_type != AccountType::Unknown;
```
Eikix, line 193:
```suggestion
        let is_deployed = *self.account_type != AccountType::Unknown;
```

crates/evm/src/model/account.cairo
```diff
@@ -176,49 +189,45 @@ impl AccountImpl of AccountTrait {
...
-    /// Returns `true` if the account is an Externally Owned Account (EOA).
-    #[inline(always)]
-    fn is_eoa(self: @Account) -> bool {
-        match self.account_type {
-            AccountType::EOA => true,
-            AccountType::ContractAccount => false
+        //TODO(account) verify whether is_known is a sufficient condition
+        // as if an account's nonce != 0 or its code is not empty,
+        // its type is necessarily known
+        if (is_known || *self.nonce != 0 || !(*self.code).is_empty()) {
```
Eikix, line 198:
```suggestion
        if (is_deployed || *self.nonce != 0 || !(*self.code).is_empty()) {
```

crates/evm/src/model/contract_account.cairo
```diff
@@ -89,35 +80,33 @@ impl ContractAccountImpl of ContractAccountTrait {
...
+    fn selfdestruct(self: @Address) -> Result<(), EVMError> {
+        let contract_account = IContractAccountSafeDispatcher { contract_address: *self.starknet };
         contract_account.selfdestruct().map_err(EVMError::SyscallFailed(CONTRACT_SYSCALL_FAILED))
     }
 
-    /// Returns a ContractAccount instance from the given `evm_address`.
+    /// Returns the addresses of a CA at the given `evm_address`.
     #[inline(always)]
-    fn at(evm_address: EthAddress) -> Result<Option<ContractAccount>, EVMError> {
+    fn at(evm_address: EthAddress) -> Result<Option<Address>, EVMError> {
```
Eikix, line 97: do we ever need ContractAccountTrait::at ? and EOATrait::at ?
They feel redundant to fetch_or_create

crates/evm/src/model/contract_account.cairo
Eikix: But abstracting the account model would mean implementing those methods on @Account, and, making the variants AccountType::Unknown and EOA no-ops, rather than panic, no?

Eikix: Or maybe we want the panic, to make sure it's never called in the wrong place

crates/evm/src/create_helpers.cairo
enitrat: added invariant but using a panic

crates/evm/src/model/contract_account.cairo
enitrat: > would mean implementing those methods on @Account

I thought about it too but sometimes we want to invoke these methods on "accounts" that are not in the cache

enitrat: > we want the panic, to make sure it's never called in the wrong place

Following discussion on invariants, I think it's good if we have these panics for now

crates/evm/src/model/account.cairo
Eikix: what is the added value of a snap on a Struct with copy types?

crates/evm/src/model/account.cairo
Eikix: I think deploy should always mean on-chain, both for kakarot and starknet (they're the same thing, if we consider being the local state to not mean deployed, but rather "maybe_scheduled_for_deployment")

scheduled for deploy means it's in cache, ready to be deployed

crates/evm/src/model/account.cairo
Eikix: ContractAccountTrait::deploy already calls set_nonce(1),
Meaning here, the only reason we have for setting a new nonce, is if the newly created account has performed many CREATE and CREATE2 calls while still being in cache.

crates/evm/src/model/account.cairo
Eikix: Ok!

Eikix: Then perhaps, is_registered

Eikix: In order to limit as much as possible new words and concepts

crates/evm/src/model/account.cairo
enitrat: it's idiomatic in cairo, functions arguments are
-> snap if. you're not modifying them
-> ref if you're modifying them
-> pass-by-value if you want to transfer ownership (and the type is not copyable) (very rare)

crates/evm/src/model/account.cairo
enitrat: Well not really because the account type can be known, and still not be registered. :thinking:

crates/evm/src/model/contract_account.cairo
enitrat: refactored

crates/evm/src/model/account.cairo
Eikix: true, resolving

crates/evm/src/model/account.cairo
enitrat: but it's actually deployed in the sense that you can interact with it. In the kakarot-sense, it's deployed. It has all the properties you expect from a deployed account.

Eikix:
Clement just uncovered a problem in Kakarot 0: the case where a self destructed account attempts to do transfers in the same tx it is deployed

consider the following actions in the same tx
- deploy an acct with value positive
- self destruct the acct
- before the end of the tx, perform a bunch of transfers
- at end of tx, acct is not deployed, so how can it transfer native token? (no approval)

Eikix (approved):
lgtm

Issue #516, opened 2023-11-06
refactor: implement `to_u64_words()` in Span<u8> extension trait similar to ByteArrayExtTrait


Eikix:
## Feature Request

When hashing parts of the bytecode, we convert from Span<u8> to ByteArray and then to u64 words.

### Describe the Feature Request

We want to go directly from Span<u8> to u64 words (ready for keccak hash)

### Describe Preferred Solution

Implement and ISO-feature to_u64_words for Span<u8>, same as for ByteArray.
Test it.
Then, remove the useless conversions from Span<u8> to Byterray to u64 words in the codebase (extcodehash, create2, any place where you can find a cairo_keccak).

### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

PR #534, opened 2023-11-09
dev: optimize to_u64_words to use span<u8> as input

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Other (please describe): opti

Resolves: #516

## What is the new behavior?

- Span<u8> are not converted to ByteArray before packing them in u64 words
-
-

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/instructions/environmental_information.cairo
+++ b/crates/evm/src/instructions/environmental_information.cairo
@@ -16,7 +16,7 @@ use openzeppelin::token::erc20::interface::{IERC20CamelDispatcher, IERC20CamelDi
 use pedersen::{PedersenTrait, HashState};
 use starknet::{Store, storage_base_address_from_felt252, ContractAddress, get_contract_address};
 use utils::constants::EMPTY_KECCAK;
-use utils::helpers::{load_word, U256Trait, ByteArrayExTrait};
+use utils::helpers::{load_word, U256Trait, U8SpanExTrait};
 use utils::math::BitshiftImpl;
 use utils::traits::{EthAddressIntoU256};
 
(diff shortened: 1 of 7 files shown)


crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -240,16 +240,14 @@ impl EnvironmentInformationImpl of EnvironmentInformationTrait {
         let evm_address = self.stack.pop_eth_address()?;
 
         let account = self.state.get_account(evm_address)?;
-        if (account.is_precompile() || (account.is_ca() && account.nonce == 0)) {
+        if (account.is_precompile() || account.nonce == 0) {
```
Eikix, line 243: why is this here?

crates/utils/src/helpers.cairo
```diff
@@ -520,13 +520,72 @@ impl SpanExtension<T, +Copy<T>, +Drop<T>> of SpanExtTrait<T> {
...
+
+        (u64_words, last_input_word, last_input_num_bytes)
+    }
+}
+
+
 #[generate_trait]
 impl BytesImpl of BytesTrait {
     // keccack256 on a bytes message
     fn compute_keccak256_hash(self: Span<u8>) -> u256 {
```
Eikix, line 587: it seems this trait isn't used anywhere?

enitrat: I think it comes from a rebase?

crates/evm/src/instructions/environmental_information.cairo
enitrat: oops, idk, must have sneaked in but it was meant for another pr

enitrat:
@Eikix rebased and ready to go 🚀

crates/utils/src/helpers.cairo
Eikix: because in your pr you modify eth_transaction hashing step
maybe it's not worth it anymore to have it and could be part of this pr to delete it

or to leave it and use it anywhere we have a span of u8? 
should a span have a self method that turns these arbitrary bytes to a hash of itself?

Eikix (approved):
lgtm 🍀🍀

Eikix (approved):
lgtm

Issue #523, opened 2023-11-07
feat: implement `validate_tx` method for ethereum transactions.


bajpai244:
## Feature Request

Implement the [`validate_tx`](https://github.com/kkrt-labs/kakarot-ssj/blob/d8cbf9c9467a1ef7512d1c157fa0cb2cec8a3059/crates/utils/src/eth_transaction.cairo#L152) method for Ethereum transactions.

Eikix:
Note that the new validate flow is := RPC will send to the EOA for validation (for tx type legacy, 1 and 2):
- The RLP encoded transaction payload (without r, s, v),
- The signature inside the get_tx_info.unbox.signature field

This is slightly different from Kakarot 0 where the full payload+signature was sent, and the EOA had to decode, and reencode. 
This is documented 

The job of the `validate` function will be to:
- hash the RLP encoded transaction payload
- verify the eth signature
- decode the RLP encoded transaction, and check chain_id and relevant fields
- return VALIDATE if successful

PR #536, opened 2023-11-11
feat: implement eth transaction validate

bajpai244:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Transaction validation is not implemented.

Resolves: #523 

## What is the new behavior?

- The logic was transaction validation has been implemented.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/utils/src/eth_transaction.cairo
+++ b/crates/utils/src/eth_transaction.cairo
@@ -63,14 +63,14 @@ impl EncodedTransactionImpl of EncodedTransactionTrait {
     #[inline(always)]
     fn decode(self: EncodedTransaction) -> Result<EthereumTransaction, EthTransactionError> {
         match self {
-            EncodedTransaction::Legacy(tx_data) => {
-                EncodedTransactionTrait::decode_legacy_tx(tx_data)
+            EncodedTransaction::Legacy(encoded_tx_data) => {
+                EncodedTransactionTrait::decode_legacy_tx(encoded_tx_data)
             },
-            EncodedTransaction::EIP1559(tx_data) => {
-                EncodedTransactionTrait::decode_typed_tx(tx_data)
+            EncodedTransaction::EIP1559(encoded_tx_data) => {
+                EncodedTransactionTrait::decode_typed_tx(encoded_tx_data)
             },
-            EncodedTransaction::EIP2930(tx_data) => {
-                EncodedTransactionTrait::decode_typed_tx(tx_data)
+            EncodedTransaction::EIP2930(encoded_tx_data) => {
+                EncodedTransactionTrait::decode_typed_tx(encoded_tx_data)
             },
         }
     }
@@ -81,11 +81,13 @@ impl EncodedTransactionImpl of EncodedTransactionTrait {
     /// message hash, chain id. The transaction hash is computed by keccak hashing the signed
     /// transaction data, which includes the chain ID in accordance with EIP-155.
     /// # Arguments
-    /// * tx_data - The raw transaction data
-    /// * tx_data - is of the format: rlp![nonce, gasPrice, gasLimit, to , value, data, chainId, 0, 0]
+    /// * encoded_tx_data - The raw rlp encoded transaction data
+    /// * encoded_tx_data - is of the format: rlp![nonce, gasPrice, gasLimit, to , value, data, chainId, 0, 0]
     /// Note: this function assumes that tx_type has been checked to make sure it is a legacy transaction
-    fn decode_legacy_tx(tx_data: Span<u8>) -> Result<EthereumTransaction, EthTransactionError> {
-        let decoded_data = RLPTrait::decode(tx_data);
+    fn decode_legacy_tx(
+        encoded_tx_data: Span<u8>
+    ) -> Result<EthereumTransaction, EthTransactionError> {
+        let decoded_data = RLPTrait::decode(encoded_tx_data);
         let decoded_data = decoded_data.map_err()?;
 
         if (decoded_data.len() != 1) {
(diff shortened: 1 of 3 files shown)


crates/utils/src/eth_transaction.cairo
```diff
@@ -3,18 +3,26 @@ use core::option::OptionTrait;
...
 use utils::helpers::U256Trait;
+use utils::helpers::{ByteArrayExTrait, BytesTrait};
 
 use utils::helpers::{U256Impl, ByteArrayExt};
 
 use utils::rlp::RLPItem;
 use utils::rlp::{RLPTrait, RLPHelpersTrait};
 
+#[derive(Drop)]
+struct ValidateTxParam {
```
Eikix, line 19: I'm not sure this name conveys semantic meaning,
It either feels like the name of this struct is just  "valide_eth_tx input param" which is weird,
or that there's a class of transaction called a Validate Tx, and this is its params,
like `DeclareTxParam`

crates/utils/src/eth_transaction.cairo
```diff
@@ -3,18 +3,26 @@ use core::option::OptionTrait;
...
+    address: EthAddress,
+    account_nonce: u128,
+    chain_id: u128,
+    tx_data: Span<u8>,
```
Eikix, line 23:
```suggestion
    encoded_tx: Span<u8>,
```

This would help differentiate with calldata

crates/utils/src/eth_transaction.cairo
Eikix: How about 
`TransactionMetadata` that would contain sender_address, account_nonce, chain_id, signature

And encoded_tx_data would be separated, as it's not considered metadata, and validate_eth_tx would take two arguments and not one

crates/utils/src/eth_transaction.cairo
```diff
@@ -242,10 +250,23 @@ impl EthTransactionImpl of EthTransaction {
...
-    ) -> Result<bool, EthTransactionError> {
-        // todo
-        panic_with_felt252('validate_eth_tx unimplemented')
+    fn validate_eth_tx(param: ValidateTxParam) -> Result<bool, EthTransactionError> {
+        let ValidateTxParam{address, account_nonce, chain_id, tx_data, signature } = param;
+
+        let decoded_tx = EthTransaction::decode(tx_data)?;
+
+        if (decoded_tx.nonce != account_nonce) {
+            return Result::Err(EthTransactionError::AccountNonceIsIncorrect);
```
Eikix, line 259: I suggest SomethingIsSomeway -> SomewaySomething
IncorrectAccountNonce

IncorrectChainId

crates/utils/src/eth_transaction.cairo
```diff
@@ -242,10 +250,23 @@ impl EthTransactionImpl of EthTransaction {
...
+        }
+        if (decoded_tx.chain_id != chain_id) {
+            return Result::Err(EthTransactionError::ChainIdIsIncoorect);
+        }
+
```
Eikix, line 264: can we add a TODO, we need to also check maxfee = gas_price * gas_limit 

because : user signs on gas_price and gas_limit, but we inject maxfee in the starknet tx for them in the rpc, and there is no way for them to check we dont put an infinite maxfee

Eikix: this can be a standalone issue for when we check in RPC that we respect this invariant

Eikix (changes requested):
great encapsulation, it's way easier for us to test this feature;D 
nice thinking and good design, a few proposed changes for semantics and but overall lgtm

crates/utils/src/eth_transaction.cairo
bajpai244: I agree with this refactor, implemented via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/536/commits/2dc1c00afe12b224beaa2ec1dad2f476afa11f20) commit.

crates/utils/src/eth_transaction.cairo
bajpai244: makes sense, done via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/536/commits/41fbc2644a31798522450e49ae19a6789771ba2d) commit.

crates/utils/src/eth_transaction.cairo
bajpai244: I am adding a TODO for this now, but if we know the gas_price and gas_limit, why does the RPC need to be provide this then?

We can simply calculate this on the smart contracts itself. Lemme know what u think.

crates/utils/src/eth_transaction.cairo
bajpai244: These names sound better, completed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/536/commits/f889e5119b2880a97ad744d5807e89c84dd90308) commit.

crates/utils/src/eth_transaction.cairo
bajpai244: I have added a TODO as of now via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/536/commits/e0352e3908a474c15625824587368c040701bb9c) commit.

Eikix: > I am adding a TODO for this now, but if we know the gas_price and gas_limit, why does the RPC need to be provide this then?
> 
> We can simply calculate this on the smart contracts itself. Lemme know what u think.

You need to provide maxFee parameter of the Starknet transaction at the RPC level, this is a protocol parameter

Eikix (approved):
lgtm

crates/utils/src/eth_transaction.cairo
```diff
@@ -238,11 +249,28 @@ impl EthTransactionImpl of EthTransaction {
...
-        panic_with_felt252('validate_eth_tx unimplemented')
+        let TransactionMetadata{address, account_nonce, chain_id, signature } = param;
+
+        let decoded_tx = EthTransaction::decode(encoded_tx_data)?;
+
+        if (decoded_tx.nonce != account_nonce) {
+            return Result::Err(EthTransactionError::IncorrectAccountNonce);
+        }
+        if (decoded_tx.chain_id != chain_id) {
+            return Result::Err(EthTransactionError::IncoorectChainId);
```
Eikix, line 264: typo

crates/utils/src/eth_transaction.cairo
```diff
@@ -238,11 +249,28 @@ impl EthTransactionImpl of EthTransaction {
     /// # Arguments
     /// * `address` -The ethereum address that is supposed to have signed the transaction
     /// * `account_nonce` - The nonce of the account
-    /// * `param tx_data` - The raw transaction data
+    /// * `param encoded_tx_data` - The raw rlp encoded transaction data
     fn validate_eth_tx(
-        address: EthAddress, account_nonce: u128, tx_data: Span<u8>
+        param: TransactionMetadata, encoded_tx_data: Span<u8>
```
Eikix, line 254:
```suggestion
        tx_metadata: TransactionMetadata, encoded_tx_data: Span<u8>
```

Eikix: will have to change throughout the function as well^^

Eikix (changes requested):
oops, one typo,
can you activate cspell if it's not already activated?
chances are it can't catch these enum names typo, but in case

crates/utils/src/eth_transaction.cairo
bajpai244: Oh! Makes sense, I forgot that we use starknet gas directly!

crates/utils/src/eth_transaction.cairo
bajpai244: fixed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/536/commits/2ab54973cd51818841cbaf4cf8efaf0400855ac3) commit.

crates/utils/src/eth_transaction.cairo
bajpai244: fixed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/536/commits/c3c6bffc210cb384edd58631cd4633f4bd8f09c3) commit.

bajpai244:
> oops, one typo, can you activate cspell if it's not already activated? chances are it can't catch these enum names typo, but in case

Gotcha, will looks into the extension as well :)

Eikix (approved):
lgtm

Issue #540, opened 2023-11-14
bug: delegatecall caller 


enitrat:
in https://github.com/kkrt-labs/kakarot-ssj/blob/506a900fdcc75431654f14c348950167fafdbd9b/crates/evm/src/call_helpers.cairo#L79

```rust
// The caller in the subcontext is the current context's current address
        let caller = self.address();

        let call_ctx = CallContextTrait::new(
            caller,
            //...
```            
This is not true for delegate call where the caller in the subctx is the caller of the current ctx.

The `init_call_sub_ctx ` functions needs  a `self_call` boolean to indicate whether the account to message-call into is self (address of the current executing account) or the call argument's address (address of the call's target account)

PR #543, opened 2023-11-15
fix: delegatecall caller

enitrat:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #540 

## What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/evm/src/call_helpers.cairo
+++ b/crates/evm/src/call_helpers.cairo
@@ -21,6 +21,7 @@ use utils::traits::{BoolIntoNumeric, U256TryIntoResult};
 /// Created in order to simplify setting up the call opcodes
 #[derive(Drop)]
 struct CallArgs {
+    caller: Address,
     to: Address,
     gas: u128,
     value: u256,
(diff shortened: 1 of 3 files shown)


crates/evm/src/call_helpers.cairo
```diff
@@ -101,7 +101,11 @@ impl MachineCallHelpersImpl of MachineCallHelpers {
         let bytecode = self.state.get_account(call_args.to.evm)?.code;
 
         // The caller in the subcontext is the current context's current address
-        let caller = self.address();
+        let caller = if self_call {
```
Eikix, line 104: I think we should add caller to the CallArgs struct, and make use of CallType enum to properly "fill" it. 
That way we don't create a new argument to init_sub_ctx, wdyt

enitrat: sounds good

Eikix (approved):
lgtm

Issue #353, opened 2023-09-15
dev(memory_operations): JUMP/JUMPI Operation Allows Invalid Jump Destinations 


lambda-0x:
See https://github.com/kkrt-labs/kakarot/issues/693 for context

This can be solved by keeping map of all valid jump locations which can be calculated during contract initialization.

The exact detail of how this map should be passed to interpreter is still not decided

PR #544, opened 2023-11-15
Feat/valid jumpdest

Eikix:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #353 

## What is the new behavior?

- add a `false_positive_jumpdest` field in the Account model, for accounts that are scheduled for deployment but aren't deployed yet, this is a Span<usize>, as Account can't hold a Felt252Dict (since StateLogChange is a dict). checks for false positive jumpdest are done in O(n) with SpanExTrait `contains`
- commit this `false_positive_jumpdest` as storage at the end of the deploy tx, henceafter, checks are done in O(1) with storage LegacyMap<usize, bool>

- [ ] Yes
- [ ] No

I do think this is quite a cumbersome update, and I wonder if we should somehow simply check the false_positive condition inside the JUMP opcode in a stateless manner.
In the end, in a function call, there aren't that many jumps, ~50 jumps in a contract, and say, 10 per method call. We simply might need to do a "search and find" in a slice of up to 32 length, in search of a PUSH_X opcode

And store in cache "hot" found false jumpdest for instance.

I'm open to challenges on this design and can refactor "quickly" this PR to change the design


enitrat (commented):
Reviewed with Elias: we drop this PR in favor of lazy checking if 0x5B opcodes are JUMP or PUSHN arguments during execution

PR #549, opened 2023-11-19
feat: implement some remaining todos

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

- Guard functions in Contract Account
- Moved `ignored` tests to `should_panic` instead - after thinking about it it makes more sense to expect a panic for a feature not working/not implemented than just ignoring the test, as we will directly notice that the test result changed if we changed a function
- Tests for KakarotCore contract-account related functions
- `reverted_with_err` function in `execute` to avoid code dup

- [ ] Yes
- [ ] No


Eikix (dismissed):
lgtm

Eikix (approved):
lgtm

Issue #537, opened 2023-11-12
feat: selfdestruct


enitrat:
[...]

Issue #535, opened 2023-11-10
bug: selfdestruct token transfers


enitrat:
## Bug Report

**Kakarot version:**

**Current behavior:**
We can't transfer tokens from a self-destructed contract crated in the same tx that is not deployed to SN state.

**Expected behavior:**
seldestruct transfers token to the recipient

**Steps to reproduce:**

**Related code:**
see https://github.com/kkrt-labs/kakarot/pull/805

```bash
insert short code snippets here
```

**Other information:**

enitrat:
Closed by #554

PR #567, opened 2023-11-22
feat: selfdestruct

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #537 #535 

## What is the new behavior?

- implementes selfdestruct
-
-

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/contracts/src/tests/test_contract_account.cairo
+++ b/crates/contracts/src/tests/test_contract_account.cairo
@@ -7,9 +7,8 @@ use evm::tests::test_utils::{ca_address, native_token};
 #[available_gas(3000000000)]
 fn test_ca_deploy() {
     let (native_token, kakarot_core) = setup_contracts_for_testing();
-    let contract_account = deploy_contract_account(
-        kakarot_core.contract_address, Default::default().span()
-    );
+    let ca_address = deploy_contract_account(ca_address(), Default::default().span());
+    let contract_account = IContractAccountDispatcher { contract_address: ca_address.starknet };
 
     let initial_bytecode = contract_account.bytecode();
     assert(initial_bytecode.is_empty(), 'bytecode should be empty');
@@ -24,7 +23,8 @@ fn test_ca_deploy() {
 fn test_ca_bytecode() {
     let (native_token, kakarot_core) = setup_contracts_for_testing();
     let bytecode = counter_evm_bytecode();
-    let contract_account = deploy_contract_account(kakarot_core.contract_address, bytecode);
+    let ca_address = deploy_contract_account(ca_address(), bytecode);
+    let contract_account = IContractAccountDispatcher { contract_address: ca_address.starknet };
 
     let contract_bytecode = contract_account.bytecode();
     assert(contract_bytecode == bytecode, 'wrong contract bytecode');
@@ -35,9 +35,8 @@ fn test_ca_bytecode() {
 #[available_gas(3000000000)]
 fn test_ca_nonce() {
     let (native_token, kakarot_core) = setup_contracts_for_testing();
-    let contract_account = deploy_contract_account(
-        kakarot_core.contract_address, Default::default().span()
-    );
+    let ca_address = deploy_contract_account(ca_address(), Default::default().span());
+    let contract_account = IContractAccountDispatcher { contract_address: ca_address.starknet };
 
     let initial_nonce = contract_account.nonce();
     assert(initial_nonce == 1, 'nonce should be 1');
@@ -55,9 +54,8 @@ fn test_ca_nonce() {
 #[available_gas(3000000000)]
 fn test_ca_storage() {
     let (native_token, kakarot_core) = setup_contracts_for_testing();
-    let contract_account = deploy_contract_account(
-        kakarot_core.contract_address, Default::default().span()
-    );
+    let ca_address = deploy_contract_account(ca_address(), Default::default().span());
+    let contract_account = IContractAccountDispatcher { contract_address: ca_address.starknet };
 
     let storage_slot = 0x555;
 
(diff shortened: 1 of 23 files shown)


crates/evm/src/model/account.cairo
```diff
@@ -196,21 +194,22 @@ impl AccountImpl of AccountTrait {
...
-                return Result::Ok(());
+            // If SELFDESTRUCT, deploy empty SN account
+            let (initial_nonce, initial_code) = if (*self.selfdestruct == true) {
+                (0, Default::default().span())
+            } else {
+                (*self.nonce, *self.code)
             };
-            let mut ca_address = ContractAccountTrait::deploy(self.address().evm, *self.code)?;
-            self.store_nonce(*self.nonce)
+            ContractAccountTrait::deploy(self.address().evm, initial_nonce, initial_code)?;
```
Eikix, line 211: I wonder
since here, we deploy an empty SN acct,
but still set back its AccountType to unknown, 

Will we collide next time we try to deploy it, but this time with Starknet OS itself?

enitrat: $\text{contract-account} \land \left( (\neg \text{is-deployed} \land \text{has-code-or-nonce}) \lor (\text{is-deployed} \land \text{has-code-or-nonce} \land \text{deployed-nonce} \neq \text{account-nonce}) \right) \Rightarrow \text{to-deploy}$

$\neg \text{is-deployed} \land \text{to-deploy} \land \text{selfdestruct} \Rightarrow \text{deploy-empty}$

$\text{is-deployed} \land \text{to-deploy} \land \text{selfdestruct} \Rightarrow \text{nothing}$

$\text{is-deployed} \land \text{contract-account} \Rightarrow \text{set-nonce}$

$\text{is-deployed} \land \neg \text{contract-account} \Rightarrow \text{nothing}$

I think this covers all cases

crates/contracts/src/tests/test_contract_account.cairo
```diff
@@ -7,9 +7,8 @@ use evm::tests::test_utils::{ca_address, native_token};
 #[available_gas(3000000000)]
 fn test_ca_deploy() {
     let (native_token, kakarot_core) = setup_contracts_for_testing();
-    let contract_account = deploy_contract_account(
-        kakarot_core.contract_address, Default::default().span()
-    );
+    let ca_address = deploy_contract_account(ca_address(), Default::default().span());
```
Eikix, line 10: should this util return an Address struct and a dispatcher?
It feels bad to declare a dispatcher every time no?
Or it could simply return only a dispatcher

enitrat: it should only return addresses as most of the time you dont need the dispatcher (except in a single file)

crates/evm/src/model/account.cairo
```diff
@@ -180,40 +175,66 @@ impl AccountImpl of AccountTrait {
...
-                        return self.selfdestruct();
-                    }
-                    self.store_nonce(*self.nonce)
-                //Storage is handled outside of the account and must be commited after all accounts are commited.
-                },
-                AccountType::Unknown => { Result::Ok(()) }
+        // If a Starknet account is already deployed for this evm address, we
+        // should "EVM-Deploy" only if the bytecode length is different (!=0) or the nonce is different.
+        let should_deploy = if is_deployed && is_ca {
+            let deployed_nonce = ContractAccountTrait::fetch_nonce(self)?;
```
Eikix, line 184: Maybe I misunderstand, but my mental model is:
-> inside this if branch, we know the CA is deployed,
- if deployed_nonce != 0, we have a problem, we've let a collision slip past! -> panic
- if deployed_nonce == 0, then yes, we can "overwrite" deploy

crates/evm/src/model/account.cairo
```diff
@@ -180,40 +175,66 @@ impl AccountImpl of AccountTrait {
...
+                deploy_starknet_contract: !is_deployed
+            )?;
         //Storage is handled outside of the account and must be commited after all accounts are commited.
-        } else {
-            Result::Ok(())
+        //TODO(bug) uncommenting this bugs, needs to be removed when fixed in the compiler
+        // return Result::Ok(());
+        };
+
+        if should_deploy {
```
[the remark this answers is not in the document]
enitrat: see comment above :)

crates/evm/src/model/account.cairo
enitrat: fixed

crates/evm/src/model/account.cairo
```diff
@@ -180,40 +175,66 @@ impl AccountImpl of AccountTrait {
...
+            return Result::Ok(());
+        };
+
+        // If the account was not scheduled for deployment - then update it if it's deployed.
+        // Only CAs have components commited on starknet.
+        if is_deployed && is_ca {
+            if *self.selfdestruct {
+                return ContractAccountTrait::selfdestruct(self);
+            }
+            self.store_nonce(*self.nonce)?;
```
Eikix, line 225: are we sure this nonce is set to 0 when someone schedules for selfdestruction?
otherwise we're destroying -> setting nonce to 0
and then here, resetting it to the former nonce value

enitrat: the selfdestruct branch returns so this line is not executed

crates/evm/src/model/contract_account.cairo
```diff
@@ -115,6 +125,15 @@ impl ContractAccountImpl of ContractAccountTrait {
         }
     }
 
+    /// Fetches the nonce of a contract account.
+    fn fetch_nonce(self: @Account) -> Result<u64, EVMError> {
```
Eikix, line 129: Should we check the account type here?

enitrat: no - it's inside contract_account

enitrat: theoritically yes but it's never called in a context where we didn't already check this.

Issue #565, opened 2023-11-21
feat: move state balances to the account object


enitrat:
## Description

The State maintains a special dict for accounts balances. This is a misconception due the the false belief that "we don't want
to fetch an account just to send ETH". But there is no way to send ETH without making a CALL (apart in SELFDESTRUCT), which means that the account code will be loaded anyway.

## Implementation details
The balance belongs to the Account instead (as in the ef tests json description for example).

PR #569, opened 2023-11-23
refactor: account balances

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #565 

## What is the new behavior?

- Moved the balance accounting to the Account object
-
-

- [ ] Yes
- [x] No


crates/evm/src/model/account.cairo
```diff
@@ -19,45 +17,69 @@ struct Account {
...
+    fn fetch_balance(mut self: AccountBuilder) -> AccountBuilder {
+        let kakarot_state = KakarotCore::unsafe_new_contract_state();
+        let native_token_address = kakarot_state.native_token();
+        let native_token = IERC20CamelDispatcher { contract_address: native_token_address };
+        self.balance = self.address.fetch_balance();
+        self
+    }
+
+    #[inline(always)]
+    fn fetch_nonce(mut self: AccountBuilder) -> AccountBuilder {
```
Eikix, line 63: I think we should either disallow fetching nonce on a non-CA account (or maybe just disallow for EOAs)
Either early return,
Panic?
Return 1?

enitrat: added ` assert!(
            self.account.account_type == AccountType::ContractAccount,
            "Cannot fetch nonce of an EOA"
        );`

Issue #572, opened 2023-11-24
feat: early return in add_transfer if recipient == sender or amount == 0 


Eikix:
### Describe the Feature Request

inside `add_transfer`, early return if (it is NOT an error):
- recipient == sender
- amount == 0

Mohiiit:
I would like to work on this one.

Mohiiit:
Hey, I was trying setting the project up in the local, I am getting these errors:

`error: Method `array` not found on type "alexandria_storage::list::List::<core::bytes_31::bytes31>". Did you import the correct trait and impl?
 --> /kakarot-ssj/crates/contracts/src/contract_account.cairo:126:35
                data: stored_list.array().expect('failed to fetch bytecode'),
                                  ^***^

error: Ambiguous method call. More than one applicable trait function with a suitable self type was found: ResultTrait::expect and OptionTrait::expect. Consider adding type annotations or explicitly refer to the impl function.
 --> /cairo-projects/kakarot-ssj/crates/contracts/src/contract_account.cairo:126:43
                data: stored_list.array().expect('failed to fetch bytecode'),`

enitrat:
Did you follow the installation steps in the readme? These one specifically.

## Installation
- Install [Scarb](https://docs.swmansion.com/scarb). To make sure your version always matches the one used by Kakarot, you can install Scarb [via asdf](https://docs.swmansion.com/scarb/download#install-via-asdf).

- [Fork](https://docs.github.com/en/get-started/quickstart/fork-a-repo) the repository and clone your fork (git clone https://github.com/<YOUR_USERNAME>/kakarot-ssj)

- Run make install to install the git hooks.

I think you're just facing a version mismatch or are not using the last `main` branch

Mohiiit:
Yes.

1. Here is my scarb version:

scarb 2.3.1 (0c8def3aa 2023-10-31)
cairo: 2.3.1 (https://crates.io/crates/cairo-lang-compiler/2.3.1)
sierra: 1.3.0


2. Yes, I forked the branch and I am on main branch.
3. Here is the output of the make install:

bash scripts/install_hook.sh
pre-commit hook has been installed successfully!
pre-push hook has been installed successfully!

enitrat:
> To make sure your version always matches the one used by Kakarot, you can install Scarb [via asdf](https://docs.swmansion.com/scarb/download#install-via-asdf).

This is why, you're not using the correct scarb version. You need this one

```shell
[...]

PR #575, opened 2023-11-24
add_transfer update

Mohiiit:
add_transfer function returning earlier when sender and recipient is same and when amount is zero

## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [*] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes

`add_transfer` would return early result when:
1. sender and recipient is ame
2. amount is  zero

Resolves: #572 

## What is the new behavior?

- behavior is same

- [ ] Yes
- [*] No

### DIFF
--- a/crates/evm/src/state.cairo
+++ b/crates/evm/src/state.cairo
@@ -254,6 +254,9 @@ impl StateImpl of StateTrait {
 
     #[inline(always)]
     fn add_transfer(ref self: State, transfer: Transfer) -> Result<(), EVMError> {
+        if (transfer.amount == 0 || transfer.sender == transfer.recipient) {
+            return Result::Ok(());
+        }
         let sender_balance = self.read_balance(transfer.sender.evm)?;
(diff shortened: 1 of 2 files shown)


crates/evm/src/state.cairo
```diff
@@ -254,6 +254,9 @@ impl StateImpl of StateTrait {
 
     #[inline(always)]
     fn add_transfer(ref self: State, transfer: Transfer) -> Result<(), EVMError> {
+        if (transfer.amount == 0 || transfer.sender == transfer.recipient) {
```
enitrat, line 257:
```suggestion
        if (transfer.amount == 0 || transfer.sender.evm == transfer.recipient.evm) {
```

let's compare evm addresses for efficiency

enitrat (commented):
great PR !

enitrat (approved):

enitrat:
Welcome to the contributors 💪

Issue #577, opened 2023-11-26
feat: fix coinbase implementation


Eikix:
## Feature Request

fix COINBASE implementation

### Describe the Feature Request

COINBASE is an opcode that should return the receiver of the block reward
currently as Kakarot is based on a Starknet full node, we can get the sequencer address with `get_execution_info_syscall`

but we cast it into an EthAddress (since we control the sequencer we can make it so that the sequencer has a 20 bytes address). but the problem is that this address will not be "linked" by our registry to its corresponding starknet address, creating a weird behaviour should someone try to send funds to this EthAddress, here is the "faulty flow":
❌: exec_coinbase returns an eth address -> add transfer to this coinbase address -> compute starknet address of eth address  -> no known address -> lost funds

### Describe Preferred Solution

- exec_coinbase should: use `get_execution_info_syscall` to get the sequencer address, this time a starknet address, call get_evm_address() on it as if it was an EOA or a CA 
- return the received eth address 
then, 
- in the constructor of kakarot core, add an argument: initial_eoas that is a Span of Eth addresses, that may or may not be empty, in its constructor, kakarot will deploy this list of initial_eoas using its own entrypoint deploy_eoa (this part is challengeable, and may be included rather in a deployment script for kakarot, rather than forced in its constructor)
- leave this list empty for now in the test utils of deployment of kakarot, the idea is that when we have a sequencer, we will treat its account as an EOA, with a couple EthAddress and StarknetAddress, and add it in the registry at deployment of kakarot, so that at all times, coinbase is a "valid" opcode that doesn't create loopholes

bajpai244:
@Eikix I can take this.

PR #581, opened 2023-11-27
feat: fix coinbase

Eikix:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

The coinbase opcode doesn't give a Kakarot registered EVM address.

Resolves: #577 

## What is the new behavior?

- The coinbase opcode gives a Kakarot registered EVM address.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/tests/test_instructions/test_block_information.cairo
+++ b/crates/evm/src/tests/test_instructions/test_block_information.cairo
@@ -224,7 +224,6 @@ fn test_exec_coinbase() {
     // Given
     let (native_token, kakarot_core) = setup_contracts_for_testing();
     let sequencer_eoa = kakarot_core.deploy_eoa(evm_address());
-    fund_account_with_native_token(sequencer_eoa, native_token, 0x1);
 
     // And
     let mut machine = MachineBuilderTestTrait::new_with_presets().build();
--- a/crates/evm/src/tests/test_instructions/test_block_information.cairo
+++ b/crates/evm/src/tests/test_instructions/test_block_information.cairo
@@ -3,14 +3,20 @@ use contracts::kakarot_core::interface::{
 };
 
 use contracts::tests::test_utils::{
-    setup_contracts_for_testing, fund_account_with_native_token, deploy_contract_account
+    setup_contracts_for_testing, fund_account_with_native_token, deploy_contract_account,
 };
+use core::clone::Clone;
+use core::result::ResultTrait;
+use core::traits::TryInto;
 use evm::instructions::BlockInformationTrait;
 use evm::model::contract_account::ContractAccountTrait;
 use evm::stack::StackTrait;
 use evm::tests::test_utils::{evm_address, MachineBuilderTestTrait};
 use openzeppelin::token::erc20::interface::IERC20CamelDispatcherTrait;
-use starknet::testing::{set_block_timestamp, set_block_number, set_contract_address};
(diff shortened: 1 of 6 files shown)


bajpai244:
Created an [issue](https://github.com/kkrt-labs/kakarot-ssj/issues/588) to keep track of the TODO.

enitrat (commented):
missing test for `exec_coinbase`

enitrat:
unless you can't test it yet?

bajpai244:
> unless you can't test it yet?

@enitrat , The current ability to deploy these sequencer accounts  on deploy of Kakarot is in TODO, I can setup a test without this functionality as well, I don't think should be an issue, but I don't think that will be a long term solution.

Lemme know, if an interim test works, then I can set one up.

enitrat:
>  I can setup a test without this functionality as well, I don't think should be an issue, but I don't think that will be a long term solution.

I think the test of `exec_coinbase` should be decorrelated from the fact that deploying KKT also deploys predefined EOAs. You can deploy a simple EOA, set the sequencer address in a mock to this EOA's SN Address, and then test that coinbase has the right behavior

Eikix:
> > I can setup a test without this functionality as well, I don't think should be an issue, but I don't think that will be a long term solution.
> 
> I think the _unit_ test of `exec_coinbase` should be decorrelated from the fact that deploying KKT also deploys predefined EOAs. You can deploy a simple EOA, set the sequencer address in a mock to this EOA's SN Address, and then test that coinbase has the right behavior

Yes the starknet::testing function exists
```rust
// Set the sequencer address to the provided value.
pub fn set_sequencer_address(address: ContractAddress) {
    cheatcode::<'set_sequencer_address'>(array![address.into()].span());
}
```

1. Deploy an EOA
2. Set the sequencer address to the EOA's stakrnet address
3. exec_coinbase
4. Check that the returned address is the EOA's

crates/evm/src/tests/test_instructions/test_block_information.cairo
```diff
@@ -208,3 +214,27 @@ fn test_randao_should_push_zero_to_stack() {
...
+// *************************************************************************
+// 0x41: COINBASE
+// *************************************************************************
+#[test]
+#[available_gas(5000000)]
+fn test_exec_coinbase() {
+    // Given
+    let (native_token, kakarot_core) = setup_contracts_for_testing();
+    let sequencer_eoa = kakarot_core.deploy_eoa(evm_address());
+    fund_account_with_native_token(sequencer_eoa, native_token, 0x1);
```
Eikix, line 227: why do you need to fund the account?

bajpai244:
@Eikix, @enitrat added the unit test + added some unwraps to this file where I could see we were missing, done via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/581/commits/0155d09638405ec0bbf4e0e07bfdbab3a9a5c03d) commit.

crates/evm/src/tests/test_instructions/test_block_information.cairo
bajpai244: Ah, I copied the boilerplate for the test, and it got in as well, removed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/581/commits/3c7f2c3a365f130329be4be3c6f814af5f2318d0) commit.

Issue #542, opened 2023-11-15
dev: unit test `init_call_sub_ctx` and `prepare_call`


enitrat:
Write unit tests for these two functions

https://github.com/kkrt-labs/kakarot-ssj/blob/main/crates/evm/src/call_helpers.cairo#L45-L47

https://github.com/kkrt-labs/kakarot-ssj/blob/main/crates/evm/src/call_helpers.cairo#L79-81

bajpai244:
@enitrat I can take this.

PR #593, opened 2023-11-28
feat: add testing for call_helpers

bajpai244:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

The [call_helpers](https://github.com/kkrt-labs/kakarot-ssj/blob/main/crates/evm/src/call_helpers.cairo) are not tested.

Resolves: #542 

## What is the new behavior?

- This adds testing for the call helpers.

- [ ] Yes
- [x] No


crates/evm/src/model/account.cairo
```diff
@@ -75,6 +76,12 @@ impl AccountBuilderImpl of AccountBuilderTrait {
         self
     }
 
+    #[inline(always)]
+    fn set_balance(mut self: AccountBuilder, balance: u256) -> AccountBuilder {
```
Eikix, line 80: we should not create a method only for the test,
might as well just mutate the account

Or better yet,
actually fund the account with the fund_with_native_token util (i forgot the exact name)

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+use evm::model::{Address, account::{AccountBuilderTrait}};
+use evm::stack::StackTrait;
+use evm::state::StateTrait;
+use evm::tests::test_utils::{MachineBuilderTestTrait, test_address, other_evm_address};
+use starknet::testing::set_contract_address;
+use starknet::{contract_address_const, EthAddress};
+
+#[test]
+#[available_gas(20000000)]
+fn test_prepare_call_type_call() {
```
Eikix, line 21: I think you can refactor those tests into a function that takes as input a @CallType and runs the tests you've written to limit redundancy in code.
And then call this function 4 times, with the 4 different types of calls

Lmk if that's not clear

bajpai244: @Eikix , so the preparation of the stack is slightly different per test, I can put that logic into the function u are mentioning, but I think this seems like a small added complexity to read the code compared to the current one, where although it is long, once can very easily read how the stack is being setup.

But, if u think going with a specific function, which encapsulate all these enums and their corresponding stack preparation is a better way to go, then let me know, and I can refactor.

crates/evm/src/model/account.cairo
Eikix: I tend to think this creates an ambiguity because you'd only probably want to call `fetch_balance`, rather than `set_balance` on the builder type
I would rather not do this

crates/evm/src/tests/test_call_helpers.cairo
Eikix: I think unifying is slightly better yes

crates/evm/src/model/account.cairo
bajpai244: can u explain why would someone want to only call `fetch_balance` and not `set_balance` on a builder?

Eikix: fetch_balance is consistent with the local storage of the current EVM context, and will yield the current balance of the account, it shouldn't be "wrong" or inacurate, so the person calling it is reassured
I would imagine `Account.set_balance` is then only called in internals, such as add_transfer

Eikix: Update: I confirm Account.set_balance is only called in State.add_transfer

When building an account, we must not be able to "set" an incorrect balance

enitrat: agreed

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+    machine.stack.push(ret_size.into());
+    machine.stack.push(ret_offset.into());
+    machine.stack.push(args_size);
+    machine.stack.push(args_offset);
+    machine.stack.push(value);
+    machine.stack.push(address.address.into());
+    machine.stack.push(gas.into());
+
+    let expected_call_args = CallArgs {
+        caller: test_address(),
```
enitrat, line 52: why is it not caller_address?

enitrat: ok, there's a problem in the test utils: the CallContext caller is `test_address()` and the ExecutionContext `address` is `test_address()`. This needs to be changed - e.g. introduce `preset_caller()` and `preset_context()` or something similar.

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+        to: Address { evm: address, starknet: kakarot_core.compute_starknet_address(address) },
+        gas,
+        value,
+        calldata: array![0x0].span(),
+        ret_offset,
+        ret_size,
+        should_transfer: true
+    };
+
+    let account = AccountBuilderTrait::new(machine.address()).set_balance(1000).build();
```
enitrat, line 267: I would prefer if the account was truely funded before - that would prevent unexpected test failures down the line - you can use the `fund_account_with_native_token` util

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+    machine.state.set_account(account);
+
+    let sender_address = machine.address().evm;
+
+    let current_ctx_prev = machine.current_ctx.as_snapshot().unbox();
```
enitrat, line 272: unused I think

enitrat (commented):
great! can you rebase on main?

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+use evm::machine::MachineTrait;
+use evm::model::{Address, account::{AccountBuilderTrait}};
+use evm::stack::StackTrait;
+use evm::state::StateTrait;
+use evm::tests::test_utils::{MachineBuilderTestTrait, test_address, other_evm_address};
+use starknet::testing::set_contract_address;
+use starknet::{contract_address_const, EthAddress};
+
+#[test]
+#[available_gas(20000000)]
```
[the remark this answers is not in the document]
bajpai244: removed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/593/commits/41a8816a11786745d7e9704cc1df5abbc9c21fb5) commit.

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+        should_transfer: false
+    };
+
+    let call_args = machine.prepare_call(@CallType::DelegateCall).unwrap();
+
+    assert(call_args == expected_call_args, 'wrong calls_args prepared');
+}
+
+#[test]
+#[available_gas(20000000)]
```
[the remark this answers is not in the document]
bajpai244: removed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/593/commits/41a8816a11786745d7e9704cc1df5abbc9c21fb5) commit.

crates/evm/src/tests/test_call_helpers.cairo
bajpai244: removed via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/593/commits/14ef9e663df1577e16e42079f193ab3ea2adfe46) commit.

crates/evm/src/tests/test_call_helpers.cairo
bajpai244: [This](https://github.com/kkrt-labs/kakarot-ssj/pull/593/commits/e705d2e3ce2d383aeadd9f835e96045e7e5fc6c8) commit makes uses a of a new util, which allows of preparing the machine for various kinds of testing.

This way, the `prepare_call` tests just need to created the expected value, and the whole redundancy of doing the machine setup is removed.

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+        value,
+        calldata: array![0x0].span(),
+        ret_offset,
+        ret_size,
+        should_transfer: true
+    };
+
+    let call_args = machine.prepare_call(@CallType::Call).unwrap();
+
+    assert(call_args == expected_call_args, 'wrong calls_args prepared');
```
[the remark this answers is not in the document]
bajpai244: resolved via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/593/commits/0384964af665de8f69bdb39fd2227461a00aa5dd) commit.

crates/evm/src/tests/test_call_helpers.cairo
```diff
@@ -0,0 +1,297 @@
...
+use starknet::{contract_address_const, EthAddress};
+
+#[test]
+#[available_gas(20000000)]
+fn test_prepare_call_type_call() {
+    let (_, kakarot_core) = setup_contracts_for_testing();
+
+    set_contract_address(kakarot_core.contract_address);
+
+    let caller_address = EthAddress { address: 0xabde2 };
```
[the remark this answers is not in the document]
bajpai244: This is the address of the [default](https://github.com/kkrt-labs/kakarot-ssj/blob/123271d5eda094cc934a1993ac9efa8a4a4373f2/crates/evm/src/tests/test_utils.cairo#L209) execution context.

Hence, to avoid any errors in testing for caller_address, made a unique caller_address.

crates/evm/src/tests/test_call_helpers.cairo
bajpai244: So,
the test is correct afaik.

This is a normal call, where the address of the current execution context should be the caller.

The `caller_address` and `test_address()` are different as well, u can check the machine builder setup, where we make sure that the `caller_address` and `test_address()` { which is the address of the current execution context } are different.

crates/evm/src/model/account.cairo
bajpai244: done via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/593/commits/47dfa142ee151fd760ee67f40674f34d8bce5ed2) commit.

crates/evm/src/tests/test_call_helpers.cairo
bajpai244: done via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/593/commits/47dfa142ee151fd760ee67f40674f34d8bce5ed2) commit.

Eikix (approved):
lgtm

PR #601, opened 2023-11-30
feat: support only valid jumps

bajpai244:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

This PR adds a new `valid_jump_destinations` to the execution context, which when initialised should contain indexes which are valid jump destinations.

Resolves: #544 

## What is the new behavior?

- Jumps to only valid jump destinations will be permitted, i.e if 0x5b is an argument to PUSH, then it cannot be confused for a valid jump.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/context.cairo
+++ b/crates/evm/src/context.cairo
@@ -141,6 +141,8 @@ struct ExecutionContext {
     return_data: Span<u8>,
     parent_ctx: Nullable<ExecutionContext>,
     gas_used: u128,
+    is_valid_jump_destinations_set: bool,
+    valid_jump_destinations: Array<u32>
 }
 
 
@@ -187,6 +189,8 @@ impl ExecutionContextImpl of ExecutionContextTrait {
             return_data,
             parent_ctx,
             gas_used: Default::default(),
+            is_valid_jump_destinations_set: false,
+            valid_jump_destinations: Default::default()
         }
     }
 
--- a/crates/evm/src/machine.cairo
+++ b/crates/evm/src/machine.cairo
@@ -1,4 +1,6 @@
+use core::array::ArrayTrait;
 use evm::errors::{EVMError, MISSING_PARENT_CONTEXT};
+use evm::instructions::comparison_operations::ComparisonAndBitwiseOperationsTrait;
 use evm::model::Address;
 use evm::state::State;
 use evm::{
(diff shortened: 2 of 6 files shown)


crates/evm/src/context.cairo
```diff
@@ -141,6 +141,8 @@ struct ExecutionContext {
     return_data: Span<u8>,
     parent_ctx: Nullable<ExecutionContext>,
     gas_used: u128,
+    is_valid_jump_destinations_set: bool,
```
enitrat, line 144: move this to the VM struct

crates/evm/src/machine.cairo
```diff
@@ -311,4 +314,69 @@ impl MachineImpl of MachineTrait {
...
+
+        let mut current_ctx = self.current_ctx.unbox();
+        let result = current_ctx.valid_jump_destinations.span().contains(dest);
+        self.current_ctx = BoxTrait::new(current_ctx);
+
+        Result::Ok(result)
+    }
+
+    fn init_valid_jump_destinations(ref self: Machine) -> Result<(), EVMError> {
+        if (self.is_valid_jump_destinations_set()) {
```
enitrat, line 332: `init_valid_jump_destinations` should do what the name suggests: init the valid jump destinations. I think checks should be done _prior_ to calling this function.

crates/evm/src/machine.cairo
```diff
@@ -311,4 +314,69 @@ impl MachineImpl of MachineTrait {
...
+        };
+
+        current_ctx.is_valid_jump_destinations_set = true;
+        self.current_ctx = BoxTrait::new(current_ctx);
+
+        Result::Ok(())
+    }
+
+    #[inline(always)]
+    fn is_valid_jump_destinations_set(ref self: Machine) -> bool {
```
enitrat, line 375: why is this field required? We could just set the valid jumpdests the first time we load the bytecode instead of doing checks at every jumps. 

I think post-refactor you would in `execute_code` run `get_valid_jumpdests` that will return a `Span<usize>` with the locations of the valid jumpdests. You will then instanciate the VM with this list of valid jumpdests.

enitrat (commented):
Did an initial review - the logic will same roughly the same but after the refactoring we can move some stuff in other places

bajpai244:
@enitrat I have opened another PR for the same [here](https://github.com/kkrt-labs/kakarot-ssj/pull/607), closing this one, u can review that!

PR #607, opened 2023-12-04
feat: valid jumps

bajpai244:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

This PR adds a new `valid_jump_destinations` to the execution context, which when initialised should contain indexes which are valid jump destinations.

Resolves: #544 

## What is the new behavior?

- Jumps to only valid jump destinations will be permitted, i.e if 0x5b is an argument to PUSH, then it cannot be confused for a valid jump.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/model/vm.cairo
+++ b/crates/evm/src/model/vm.cairo
@@ -63,12 +63,46 @@ impl VMImpl of VMTrait {
         self.pc = pc;
     }
 
+    fn init_valid_jump_destinations(ref self: VM) {
+        let bytecode = self.message.code;
+        let mut valid_jumpdests: Array<usize> = array![];
+
+        let mut i = 0;
+        loop {
+            if (i >= bytecode.len()) {
+                break;
+            }
+
+            let opcode = *bytecode.at(i);
+            // checking for PUSH opcode family
+            if (opcode >= 0x5f && opcode <= 0x7f) {
+                let number_of_args = opcode - 0x5f;
+                i += (number_of_args + 1).into();
+                continue;
+            }
+
+            // JUMPDEST
+            if (opcode == 0x5B) {
+                valid_jumpdests.append(i.into());
+            }
+
+            i += 1;
+        };
+
+        self.set_valid_jumpdests(valid_jumpdests.span());
+    }
+
     fn valid_jumpdests(self: @VM) -> Span<usize> {
         *self.valid_jumpdests
     }
 
     fn set_valid_jumpdests(ref self: VM, valid_jumpdests: Span<usize>) {
         self.valid_jumpdests = valid_jumpdests;
+    }
+
+    #[inline(always)]
+    fn is_valid_jump(self: @VM, dest: u32) -> bool {
+        self.valid_jumpdests().contains(dest)
     }
 
     fn return_data(self: @VM) -> Span<u8> {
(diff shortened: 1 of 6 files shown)


crates/evm/src/model/vm.cairo
```diff
@@ -57,6 +58,35 @@ impl VMImpl of VMTrait {
         self.pc = pc;
     }
 
+    fn init_valid_jump_destinations(ref self: VM) {
+        let bytecode = self.message.code;
+        let mut valid_jumpdests: Array<usize> = array![];
+
+        let mut i = 0;
+        loop {
```
enitrat, line 66: prefer looping using `match bytecode.pop_front()`
for optimization purposes:

see:

```rust
#[test]
fn test_loop_match() {
    let mut data = data();
    loop {
        match data.pop_front() {
            Option::Some(byte) => { println!("{}", *byte); },
            Option::None => { break; }
        }
    }
}

#[test]
fn test_loop_at() {
    let mut data = data();
    let mut i = 0;
    loop {
        if i >= data.len() {
            break;
        }
        println!("{}", *data.at(i));
        i += 1;
    }
}
```
```shell
test benchmark_loop::test_loop_match ... ok (gas usage est.: 7915600)
test benchmark_loop::test_loop_at ... ok (gas usage est.: 8202230)
```

crates/evm/src/tests/test_model/test_vm.cairo
```diff
@@ -0,0 +1,41 @@
...
+        0x60, 0x3, 0x56, 0x5b, 0x60, 0x9, 0x56, 0x60, 0x2, 0x5b, 0x60, 0x2
+    ];
+    let mut message: Message = Default::default();
+    message.code = bytecode.span();
+
+    let mut vm = VMTrait::new(message, Default::default());
+
+    vm.init_valid_jump_destinations();
+
+    let expected_valid_jump_destinations = array![0x3, 0x9].span();
```
enitrat, line 17: a bit disturbing to use hex representation for array indexes here - I personally don't count in hexadecimal in my head 😆

crates/evm/src/tests/test_model/test_vm.cairo
```diff
@@ -0,0 +1,41 @@
...
+
+    assert!(vm.is_valid_jump(0x3) == true, "expected jump to be valid");
+    assert!(vm.is_valid_jump(0x9) == true, "expected jump to be valid");
+
+    assert!(vm.is_valid_jump(0x4) == false, "expected jump to be invalid");
+    assert!(vm.is_valid_jump(0x5) == false, "expected jump to be invalid");
+}
+
+#[test]
+fn test_valid_jump_destination_failing() {
```
enitrat, line 31: I would rather have a test that mixes false positives with valid jump dests

The name is not correct, here the valid_jump_destination is not failing, actually it performs as expected (the PUSHN argument is not counted as valid jumpdest)

enitrat (commented):
nice (and very simple, nice!)

crates/evm/src/model/vm.cairo
enitrat: I guess in this context since you still need to increment`i` and are not incrementing linearly we can keep it your way but it's good to remember this nevertheless

once we have iterators we can prbably make something even cleaner

crates/evm/src/tests/test_model/test_vm.cairo
bajpai244: Makes sense, so I think I should rename this to `test_valid_jump_destination_inside_jumpn`, because this is what I wanted to test.

bajpai244: resolved via [this](https://github.com/kkrt-labs/kakarot-ssj/pull/607/commits/230b5befeb5d07b294ed69ae76607b58137fac58) commit.

Issue #603, opened 2023-11-30
feat: `ensure` for invariants


enitrat:
name to be discussed, but basically a function like
`fn ensure(cond: bool) -> Result<(), EVMError>`

that we can call like

```rust
fn foo(...) -> Result<(), EVMError> {
  ensure(index < len, EMVError::StackError(STACK_UNDERFLOW))?:
  // do stuff
```

instead of

```rust
fn foo(...) -> Result<(), EVMError> {
  if (index >= len){
     return EMVError::StackError(STACK_UNDERFLOW))?:
   }
  // do stuff
```

Eikix:
Agreed! Looks like the ensure from the execution specs

PR #625, opened 2023-12-06
feat: ensure checks

### LINKED ISSUE #545: bug: check Depth level when entering a new context
## Bug Report

**Current behavior:**

Currently, we entering a new sub context, we don't check depth < 1024

**Expected behavior:**

In create and call, check for depth < 1024
Machine.context_count will not suffice, as it never diminishes

Open to proposal for this:
- a new machine argument for "current depth"
- an execution context field

enitrat:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #603 Closes #545

## What is the new behavior?

- `ensure` function that returns an error if a condition is not met
-  replaced all `if cond then return Err` by `ensure(cond, err)`
- added missing invariants checks

- [ ] Yes
- [x] No

### DIFF
--- a/crates/evm/src/call_helpers.cairo
+++ b/crates/evm/src/call_helpers.cairo
@@ -129,7 +129,7 @@ impl CallHelpersImpl of CallHelpers {
     fn generic_call(ref self: VM, call_args: CallArgs) -> Result<(), EVMError> {
         self.return_data = Default::default().span();
         if self.message().depth >= constants::STACK_MAX_DEPTH {
-            self.gas_used -= call_args.gas;
+            self.gas_left += call_args.gas;
             return self.stack.push(0);
         }
 
--- a/crates/evm/src/interpreter.cairo
+++ b/crates/evm/src/interpreter.cairo
@@ -106,7 +106,7 @@ impl EVMImpl of EVMTrait {
                 success: false,
                 return_data: Into::<felt252, u256>::into(EVMError::DepthLimit.to_string())
                     .to_bytes(),
-                gas_used: 0,
+                gas_left: 0,
                 accessed_addresses: Default::default(),
                 accessed_storage_keys: Default::default()
             };


crates/evm/src/call_helpers.cairo
```diff
@@ -126,12 +129,15 @@ impl CallHelpersImpl of CallHelpers {
     /// newly created sub-context.
     /// Then, the EVM execution loop will start on this new execution context.
     fn generic_call(ref self: VM, call_args: CallArgs) -> Result<(), EVMError> {
-        // check if depth is too high
+        self.return_data = Default::default().span();
+        if self.message().depth >= constants::STACK_MAX_DEPTH {
+            self.gas_used += self.message().gas_limit;
```
Eikix, line 134: why += ?
this means at some point we'll have to re-assign gas used to be min(gas used, gas limit)

crates/evm/src/create_helpers.cairo
```diff
@@ -83,39 +85,45 @@ impl CreateHelpersImpl of CreateHelpers {
...
             return self.stack.push(0);
         }
 
+        let mut target_account = self.env.state.get_account(create_args.to);
+        let target_address = target_account.address();
         // Collision happens if the target account loaded in state has code or nonce set, meaning
         // - it's deployed on SN and is an active EVM contract
         // - it's not deployed on SN and is an active EVM contract in the Kakarot cache
         if target_account.has_code_or_nonce() {
+            target_account.set_nonce(target_account.nonce() + 1);
```
Eikix, line 113: i think it's the sender account here in the spec

crates/evm/src/call_helpers.cairo
enitrat: we reimburse the gas passed to the child message - but indeed it should not be `self.message().gas_limit` but `call_args.gas` if I'm not mistaken - which is the parameter that defined the amount of gas this call can consume

crates/evm/src/call_helpers.cairo
```diff
@@ -3,7 +3,9 @@ use cmp::min;
 use contracts::kakarot_core::KakarotCore;
 use contracts::kakarot_core::interface::IKakarotCore;
 
-use evm::errors::{EVMError, CALL_GAS_GT_GAS_LIMIT, ACTIVE_MACHINE_STATE_IN_CALL_FINALIZATION};
+use evm::errors::{
```
bajpai244, line 6: Do we use `ensure` anywhere in this file? Can't spot it.

crates/evm/src/create_helpers.cairo
```diff
@@ -83,39 +85,45 @@ impl CreateHelpersImpl of CreateHelpers {
...
+        self.return_data = Default::default().span();
+
+        // The sender in the subcontext is the message's target
+        let sender_address = self.message().target;
+        let mut sender = self.env.state.get_account(sender_address.evm);
+        let sender_current_nonce = sender.nonce();
+        if sender.balance() < create_args.value
+            || sender_current_nonce == integer::BoundedInt::<u64>::max()
             || self.message.depth
-            + 1 == constants::STACK_MAX_DEPTH {
```
bajpai244, line 99: Oh, does this mean earlier we had an error in this logic? If yes, then how did we pass the test?

crates/evm/src/tests/test_stack.cairo
```diff
@@ -275,10 +264,7 @@ mod swap {
         let result = stack.swap_i(1);
 
         assert(result.is_err(), 'should return an EVMError');
-        assert(
-            result.unwrap_err() == EVMError::StackError(STACK_UNDERFLOW),
-            'should return StackUnderflow'
-        );
+        assert(result.unwrap_err() == EVMError::StackUnderflow, 'should return StackUnderflow');
```
bajpai244, line 267: might we should use `assert!`, as we decided it is a better alternative?

@enitrat I can also make a `good first issue` for some to change all `assert` to `assert!` in our codebase, wdu think?

crates/evm/src/tests/test_stack.cairo
```diff
@@ -291,9 +277,6 @@ mod swap {
         let result = stack.swap_i(2);
 
         assert(result.is_err(), 'should return an EVMError');
-        assert(
-            result.unwrap_err() == EVMError::StackError(STACK_UNDERFLOW),
-            'should return StackUnderflow'
-        );
+        assert(result.unwrap_err() == EVMError::StackUnderflow, 'should return StackUnderflow');
```
bajpai244, line 280: same here, we can prefer `assert`.

crates/evm/src/tests/test_stack.cairo
```diff
@@ -43,7 +43,7 @@ fn test_len_should_return_the_length_of_the_stack() {
 
 #[cfg(test)]
 mod push {
-    use evm::errors::{EVMError, STACK_OVERFLOW};
+    use evm::errors::{EVMError};
```
bajpai244, line 46:
```suggestion
    use evm::errors::EVMError;
```

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -46,9 +44,8 @@ impl SystemOperations of SystemOperationsTrait {
         let value = call_args.value;
 
         // Check if current context is read only that value == 0.
-        if read_only && (value != 0) {
-            return Result::Err(EVMError::WriteInStaticContext(VALUE_TRANSFER_IN_STATIC_CALL));
-        }
+        // De Morgan's law: !(read_only && value != 0) == !read_only || value == 0
```
bajpai244, line 47: nice comment, adds more clarity.

crates/evm/src/tests/test_utils.cairo
```diff
@@ -21,7 +21,7 @@ struct VMBuilder {
 #[generate_trait]
 impl VMBuilderImpl of VMBuilderTrait {
     fn new() -> VMBuilder {
-        VMBuilder { vm: Default::default() }
+        VMBuilder { vm: Default::default() }.with_gas_limit(0x100000000000000000000)
```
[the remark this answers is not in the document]
enitrat: high gas value for tests

crates/evm/src/create_helpers.cairo
bajpai244: So, what I am wondering is that:
- `message_depth + 1 == constants::STACK_MAX_DEPTH` &
- `message_depth + 1  > constants::STACK_MAX_DEPTH`

are two different conditions, I think the current one makes sense, i.e where we check `>`, but I guess, `==` is wrong, because we haven't reached the max depth yet right? So, does that mean the old condition is wrong?

Because, if the answer to the above is yes, then we can try to think, how could we have tested this.

crates/evm/src/tests/test_stack.cairo
bajpai244: Gotcha, yeah, because Cairo is based on Rust, the formatter usually does this, but I don't have very strong opinions about it, so fine.

crates/evm/src/tests/test_stack.cairo
bajpai244: gotcha, makes sense!

bajpai244 (approved):
lgtm 🚀

PR #658, opened 2023-12-13
fix: read state on undeployed alive account

enitrat:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #656 

## What is the new behavior?

Accounts can be either
- unalive && undeployed (and thus, "Unkown")
- alive && undeployed (and thus, their type is set - alive means that it's in the state but not commited yet)

In this PR
- Added a check to ensure that when reading from state, the account is __deployed__
- Added a related test

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/658)


greged93 (approved):
tested on ef-tests this fixes the issue

Issue #655, opened 2023-12-13
feat(gas): compute, charge, reimburse create message gas


enitrat:
## Feature Request

Compute the gas used by a create message, and charge it.
Reimburse the gas if 
```rust
 sender.balance() < create_args.value
            || sender_current_nonce == integer::BoundedInt::<u64>::max()
            || self.message.depth == constants::STACK_MAX_DEPTH
 ```

PR #661, opened 2023-12-14
feat(gas): charge-create-message-gas

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #655 

## What is the new behavior?

- Compute charge and reimburse create messages gas
-
-

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/661)

### DIFF
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -86,7 +86,8 @@ impl CreateHelpersImpl of CreateHelpers {
     fn generic_create(ref self: VM, create_args: CreateArgs) -> Result<(), EVMError> {
         self.accessed_addresses.add(create_args.to);
 
-        //TODO(gas) compute and charge create message gas
+        let create_message_gas = gas::max_message_call_gas(self.gas_left);
+        self.gas_left -= create_message_gas;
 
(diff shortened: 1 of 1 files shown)


crates/evm/src/create_helpers.cairo
```diff
@@ -86,7 +86,8 @@ impl CreateHelpersImpl of CreateHelpers {
     fn generic_create(ref self: VM, create_args: CreateArgs) -> Result<(), EVMError> {
         self.accessed_addresses.add(create_args.to);
 
-        //TODO(gas) compute and charge create message gas
+        let create_message_gas = gas::max_message_call_gas(self.gas_left);
```
bajpai244, line 89: If you leave these two lines below 105, then there will be no need to add the charged gas back on line 102!

Lemme know what u think!

enitrat: no, because we run an `ensure` which might early return, and if we execute a `create` that has an exceptional failure it must consume all gas

bajpai244 (approved):

PR #693, opened 2023-12-29
chore: remove prints

enitrat:
Removes print statements, as they're quite gas consuming and they make some EF-tests reach OOG 😢

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/693)

### DIFF
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -392,7 +392,6 @@ mod KakarotCore {
             ) {
                 Result::Ok(_) => {},
                 Result::Err(err) => {
-                    println!("process_transaction: Insufficient balance for fees and transfer");
                     return TransactionResultTrait::exceptional_failure(
                         err.to_bytes(), tx.gas_limit()
                     );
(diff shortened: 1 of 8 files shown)


crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -504,7 +500,7 @@ mod KakarotCore {
             //     ) {
             //     Result::Ok(_) => {},
             //     Result::Err(err) => {
-            //         println!("process_transaction: sequencer couldn't charge gas");
+            //
```
bajpai244, line 503: do we need comment here? I guess we can remove the whole line itself.

crates/evm/src/tests/test_state.cairo
```diff
@@ -20,7 +20,7 @@ fn test_compute_state_key() {
     //         ];
     //     let values_to_hash = [keys[0],keys[1],keys[2]];
     //     let hash = poseidon_hash_many(&values_to_hash);
-    //     println!("hash: {}", hash);
+    //
```
bajpai244, line 23: same, do we need `//` here? we can remove the whole line I guess.

crates/contracts/src/kakarot_core/kakarot.cairo
enitrat: yes, we need, it's explained above

crates/evm/src/tests/test_state.cairo
enitrat: (also for these kind of nitpicks it's easier if you make a suggestion so that I can do it in github)

bajpai244 (approved):

Issue #705, opened 2024-01-22
feat: add `Felt252Vec` and `Span` utils for `modexp`


bajpai244:
## Feature Request

Add all utils used used for `Felt252Vec` and `Span` in pr #699 

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [] No

PR #699, opened 2024-01-17
Feat/add modexp precompile

bajpai244:
## Pull Request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

## What is the current behavior?

Modexp precompile isn't present.

Resolves: #663 

## What is the new behavior?

- Arbitrary precision Modexp is added
- We use [EIP-2565](https://eips.ethereum.org/EIPS/eip-2565) for gas accounting.
## Does this introduce a breaking change? 

- [ ] Yes
- [x] No

The precompile code is taken from [aurora-engine](https://github.com/aurora-is-near/aurora-engine/tree/develop/engine-modexp)  { which is also what revm uses as of now. }

[revm](https://github.com/bluealloy/revm/blob/main/crates/precompile/src/modexp.rs) is also referenced in parts of implementation was the wrapper code.

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/699)

PR #706, opened 2024-01-22
feat: add bytes related utils

bajpai244:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #705 

## What is the new behavior?

- The PR adds utils used for `Felt252Vec` and `Span` in #699 

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)


enitrat (changes requested):
Thanks for the PR! I will not review the Felt252Vec part in this PR, it's probably more appropriate to have it in the next one as there are already quite a lot of stuff to address here.

> Reviewed 2 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-:-Nom1IuI03MH7tYiWQIB:b-r5x2e9)* status: 2 of 6 files reviewed, 17 unresolved discussions (waiting on @bajpai244 and @ClementWalter)

___
*[`crates/utils/src/helpers.cairo` line 11 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoljFZNBuu_TudA5QHY:-NoljFZNBuu_TudA5QHZ:b-xmj35k) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L11)):*
> ```Cairo
> use core::traits::TryInto;
> use integer::{BoundedInt, u32_as_non_zero, U32TryIntoNonZero};
> use integer::{u32_overflowing_add};
> ```

don't we have a trait for this now?
___
*[`crates/utils/src/helpers.cairo` line 642 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoljJI22b9XL3PdFJcs:-NoljJI22b9XL3PdFJct:bwpvbbc) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L642)):*
We need proper function doc also for arguments and returned data
Would be nice to have a visual example of as well, e.g `get_right_padded_span([1,2,3,4], 2,5) -> [2,3,4,0,0]` (right?)

From last review: 
> I don't think the name of the function reflects what it's supposed to return correctly. From what I understand you take a span of size L as input, slice it from \[offset, len\] and then add zeroes until both length match, right?

I think the function name should reflect this behavior. 
1. It produces a slice of the original array
2. You append zeroes at the end of this slice until it matches the `len` provided
3. You return the new object

I propose the name`slice_right_padded` that reflects this behavior well imo.
___
*[`crates/utils/src/helpers.cairo` line 652 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NollWuE24JjN0tSREXq:-NollWuE24JjN0tSREXr:bv69rdx) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L652)):*
> ```Cairo
>         };
> 
>         let tmp = match u32_overflowing_add(start, len) {
> ```

this sounds dangerous. if the result overflows we want to return the same value anyway?

consider this:
start = `0xfffffff6` (2**32 - 10)
len = `12`
=> tmp = `2`

`tmp <= self.len()` is true => start+len overflows and fails. Thus, i'm not quite sure what this is for: why not have `start+len` from the start? (and why not using cmp::max(self.len(), start+len)` if what you're trying to do is simply returning MAX(A,B) ?)
___
*[`crates/utils/src/helpers.cairo` line 662 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolmUo8BNsZzh0QjdqW:-NolmUo8BNsZzh0QjdqX:btu8tz4) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L662)):*
> ```Cairo
>         };
> 
>         // Copy the span
> ```

we can simply slice + clone the slice
___
*[`crates/utils/src/helpers.cairo` line 679 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolo0Eo0kcT_M0S_4-0:-Nolo0Eo0kcT_M0S_4-1:b-5bigxw) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L679)):*
> ```Cairo
>         // Pad the span
>         let diff = len - arr.len();
>         let mut i = 0;
> ```

start with `i = arr.len()` and stop when `i == len`. Makes the stop condition more clear and saves you a substraction
___
*[`crates/utils/src/helpers.cairo` line 692 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolmZIYBQjemHv57Z2j:-NolmZIYBQjemHv57Z2k:b4kcbqz) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L692)):*
> ```Cairo
>     }
> 
>     /// Left padding until `len`. If data is more than len, truncate the right most bytes.
> ```

Proper function doc
___
*[`crates/utils/src/helpers.cairo` line 692 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolnRw49-ZURPfwo29l:-NolnRw49-ZURPfwo29m:undefined) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L692)):*
> ```Cairo
>     }
> 
>     /// Left padding until `len`. If data is more than len, truncate the right most bytes.
> ```





_Suggestion:_
```Cairo
//// Clones and pads the given span to the left with zeroes until it has a length of `len`...
//// # Arguments
?/// ...
```
___
*[`crates/utils/src/helpers.cairo` line 693 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolmabt7qOfAFDQv6Rw:-Nolmabt7qOfAFDQv6Rx:bxb6imy) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L693)):*
> ```Cairo
> 
>     /// Left padding until `len`. If data is more than len, truncate the right most bytes.
>     fn left_padding(self: Span<u8>, len: usize) -> Span<u8> {
> ```

the function performs an action: `pad_left` is more appropriate. To be more precise we could probably even call it `pad_left_with_zeroes`, as it could be padded with any value. The function name should be expressed in active voice
___
*[`crates/utils/src/helpers.cairo` line 702 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolpzFw6r1zxyrNLg96:-NolpzFw6r1zxyrNLg97:bpxoqiv) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L702)):*
> ```Cairo
>         let mut i = 0;
>         loop {
>             if i == (len - self.len()) {
> ```

start at `self.len()` and stop at `i == len`. saves you N substractions.
___
*[`crates/utils/src/helpers.cairo` line 815 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolriQX4O7kSApvinBh:-NolriQX4O7kSApvinBi:bq4h9d9) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L815)):*
> ```Cairo
>         let res = self.to_be_bytes();
> 
>         let mut i = 0;
> ```

end condition should be computed outside of the loop



_Suggestion:_
```Cairo
let mut i = res.len()
...
if i == 4 // break|
```
___
*[`crates/utils/src/helpers.cairo` line 824 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nols2Gj9gvtcnu-iPLE:-Nols2Gj9gvtcnu-iPLF:b-twvoy4) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L824)):*
> ```Cairo
>             i += 1;
>         };
> ```

shouldn't you use your `pad_left` function btw? 
This entire function could be
`pad_left(self.to_be_bytes(), 4)`
___
*[`crates/utils/src/helpers.cairo` line 962 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolsX5C1twYZqCj5kxi:-NolsX5C1twYZqCj5kxj:bnj7hl0) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L962)):*
> ```Cairo
>     /// # Returns
>     /// * The bytes array representation of the value.
>     fn to_be_bytes_padded(mut self: u64) -> Array<u8> {
> ```

same as above
___
*[`crates/utils/src/helpers.cairo` line 1052 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolsq1KEYDsJFZAu9ip:-Nolsq1KEYDsJFZAu9iq:b1stiyf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1052)):*
This sounds overly expansive. Let's try to think about this the other way:
> How many bits are used inside this word?

First, we can state that there are _at most_ `8 * bytes_used` bits used in this word.
Thus, `self.bytes_used()` returns the higher boundary of the amount of used bits in this word.

I know for sure that the actual number of bits used in this word is included in the range [7*bytes_used + 1, 8*bytes_used].

Here, you realise that you actually only need to count the bits used in a _single_ u8, not in a full range uint.
To count how many effective bits this u8 has, you can do a shift_right by (8*(bytes_used -1) to get the upper byte. Which leaves you to only 8 bits to count, and you can do this easily by dichotomy.

Example

0x001F_0FF0_FFFF

bytes_used = 5
shift right by 4*8 = 32 => 0x1F

from here, you can just manually count the bits used with binary search.
and the result is (bytes_used -1)*8 + bits_in_leading_byte.
___
*[`crates/utils/src/helpers.cairo` line 1072 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom0kv22NjSsnRkzWUG:-Nom0kv22NjSsnRkzWUH:b2ss7xm) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1072)):*
> ```Cairo
> 
>     /// Returns the number of bits required to represent `self`, ignoring leading zeros.
>     fn bit_len(self: u64) -> u8 {
> ```

see comment above
___
*[`crates/utils/src/helpers.cairo` line 1150 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom0mdF7okeF61VYDp6:-Nom0mdF7okeF61VYDp7:b-ien078) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1150)):*
> ```Cairo
> 
>     /// Returns the Least signficant 64 bits of a u128
>     fn as_u64(self: u128) -> u64 {
> ```

as_u64 when you're truncating a word doesn't seem like a proper naming. Why not simply use `u128_split` where required?
___
*[`crates/utils/src/helpers.cairo` line 1186 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom13_f-VL_9iJPm9LW:-Nom13_f-VL_9iJPm9LX:b-64qmrs) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1186)):*
> ```Cairo
> 
>         let mut i = 0;
>         loop {
> ```

same as other functions. By the way: why aren't we making this generic? It's always the same. logic.
___
*[`crates/utils/src/helpers.cairo` line 1341 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom1CS85u4-7_RJHqCH:-Nom1CS85u4-7_RJHqCI:bnj7hl0) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1341)):*
> ```Cairo
> 
>     /// Returns the number of bits required to represent `self`, ignoring leading zeros.
>     fn bit_len(self: u256) -> u32 {
> ```

same as above

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 2 of 6 files reviewed, 17 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 11 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoljFZNBuu_TudA5QHY:-NooxgTg4BRZAHCmsP2i:b-rz341l) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L11)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

don't we have a trait for this now?
</blockquote></details>

we have it for multiplication
___
*[`crates/utils/src/helpers.cairo` line 642 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoljJI22b9XL3PdFJcs:-Nooxu585bmvG_hxEegW:be2lv90) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L642)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

We need proper function doc also for arguments and returned data
Would be nice to have a visual example of as well, e.g `get_right_padded_span([1,2,3,4], 2,5) -> [2,3,4,0,0]` (right?)

From last review: 
> I don't think the name of the function reflects what it's supposed to return correctly. From what I understand you take a span of size L as input, slice it from \[offset, len\] and then add zeroes until both length match, right?

I think the function name should reflect this behavior. 
1. It produces a slice of the original array
2. You append zeroes at the end of this slice until it matches the `len` provided
3. You return the new object

I propose the name`slice_right_padded` that reflects this behavior well imo.
</blockquote></details>

The name was borrowed from [revm utils](https://github.com/bluealloy/revm/blob/1ead59db3e858a00f5bb2c85bd7b9ec7cdee826f/crates/precompile/src/utilities.rs#L8) to make the name of functions used in the code of the core lib similar to the source, but I am fine making ur changes as it makes sense.

Done!

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 2 of 8 files reviewed, 17 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 652 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NollWuE24JjN0tSREXq:-Nop7D_04mW5fXc2_Csa:b-hwbrga) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L652)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

this sounds dangerous. if the result overflows we want to return the same value anyway?

consider this:
start = `0xfffffff6` (2**32 - 10)
len = `12`
=> tmp = `2`

`tmp <= self.len()` is true => start+len overflows and fails. Thus, i'm not quite sure what this is for: why not have `start+len` from the start? (and why not using cmp::max(self.len(), start+len)` if what you're trying to do is simply returning MAX(A,B) ?)
</blockquote></details>

I see, this is a mistake, what I needed was [saturating_add](https://doc.rust-lang.org/std/intrinsics/fn.saturating_add.html), but my mistake I used this, thanks for catching this.

I have implemented a generic `saturating_add` which fixes this.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 2 of 8 files reviewed, 17 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 662 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolmUo8BNsZzh0QjdqW:-Nop9w2u4-GHC6h0dd0U:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L662)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

we can simply slice + clone the slice
</blockquote></details>

done
___
*[`crates/utils/src/helpers.cairo` line 679 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolo0Eo0kcT_M0S_4-0:-NopANq654RD5EfcuIki:bqj605s) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L679)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

start with `i = arr.len()` and stop when `i == len`. Makes the stop condition more clear and saves you a substraction
</blockquote></details>

I just realized a more simple approach is to juse check for `arr.len() == len` !

Done.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 2 of 8 files reviewed, 17 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 692 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolnRw49-ZURPfwo29l:-NopClJW5tZaBvH0pVZq:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L692)):*
> ```Cairo
>     }
> 
>     /// Left padding until `len`. If data is more than len, truncate the right most bytes.
> ```

done
___
*[`crates/utils/src/helpers.cairo` line 693 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolmabt7qOfAFDQv6Rw:-NopClwoB8IOF2EPJzst:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L693)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

the function performs an action: `pad_left` is more appropriate. To be more precise we could probably even call it `pad_left_with_zeroes`, as it could be padded with any value. The function name should be expressed in active voice
</blockquote></details>

done

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 2 of 8 files reviewed, 17 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 702 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolpzFw6r1zxyrNLg96:-NopLxCqC2Q0QYSZ7a74:b1pjrew) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L702)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

start at `self.len()` and stop at `i == len`. saves you N substractions.
</blockquote></details>

A simple condition is ` arr.len() == (len - self.len()) ` avoid even addition.

Done

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 2 of 10 files reviewed, 17 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 692 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NolmZIYBQjemHv57Z2j:-NoprQOt1y6814VARiYY:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L692)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Proper function doc
</blockquote></details>

done
___
*[`crates/utils/src/helpers.cairo` line 1052 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolsq1KEYDsJFZAu9ip:-NopTIOGBbNyb4V1aG4n:bqhdhzc) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1052)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

This sounds overly expansive. Let's try to think about this the other way:
> How many bits are used inside this word?

First, we can state that there are _at most_ `8 * bytes_used` bits used in this word.
Thus, `self.bytes_used()` returns the higher boundary of the amount of used bits in this word.

I know for sure that the actual number of bits used in this word is included in the range [7*bytes_used + 1, 8*bytes_used].

Here, you realise that you actually only need to count the bits used in a _single_ u8, not in a full range uint.
To count how many effective bits this u8 has, you can do a shift_right by (8*(bytes_used -1) to get the upper byte. Which leaves you to only 8 bits to count, and you can do this easily by dichotomy.

Example

0x001F_0FF0_FFFF

bytes_used = 5
shift right by 4*8 = 32 => 0x1F

from here, you can just manually count the bits used with binary search.
and the result is (bytes_used -1)*8 + bits_in_leading_byte.
</blockquote></details>

This sounds more effective, I am not sure how we can use a binary search for the last byte, but I am using the shr strategy on the last byte, which should save a lot of steps.

I am using this technique in `bit_len`, and hence making this function return ` 64 - self.bit_len()`
___
*[`crates/utils/src/helpers.cairo` line 1072 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom0kv22NjSsnRkzWUG:-NopURkwEkGjGZ2loGpY:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1072)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

see comment above
</blockquote></details>

done
___
*[`crates/utils/src/helpers.cairo` line 1150 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom0mdF7okeF61VYDp6:-NopVi8y5ul85_DZI9gu:b2pk7h3) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1150)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

as_u64 when you're truncating a word doesn't seem like a proper naming. Why not simply use `u128_split` where required?
</blockquote></details>

This is inspired from the rust syntax, where once can do `25_u128 as u64`, which will return it as a u64, ignoring all the higher bits, we use this a lot in the modexp lib, and hence use a function for this instead.
___
*[`crates/utils/src/helpers.cairo` line 1186 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom13_f-VL_9iJPm9LW:-NoprM7G-40kPkGxgwbh:b-jddzvy) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1186)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

same as other functions. By the way: why aren't we making this generic? It's always the same. logic.
</blockquote></details>

made a generic
___
*[`crates/utils/src/helpers.cairo` line 1341 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nom1CS85u4-7_RJHqCH:-NoprNCJ9d0b-ul1AS0F:b-jddzvy) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1341)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

same as above
</blockquote></details>

made a generic

enitrat (changes requested):
> Reviewed 2 of 6 files at r1, 2 of 3 files at r3, 1 of 3 files at r8, 1 of 3 files at r10, 1 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @bajpai244 and @ClementWalter)

___
*[`crates/utils/src/helpers.cairo` line 652 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NollWuE24JjN0tSREXq:-NoqDObKErqQu7hDZ8yI:b-rjcd9g) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L652)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

I see, this is a mistake, what I needed was [saturating_add](https://doc.rust-lang.org/std/intrinsics/fn.saturating_add.html), but my mistake I used this, thanks for catching this.

I have implemented a generic `saturating_add` which fixes this.
</blockquote></details>

I still think the logic of having `start + len` is not correct. Isn't this simply
`cmp::min(start.saturating_add(len), self.len()`?
___
*[`crates/utils/src/helpers.cairo` line 1052 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolsq1KEYDsJFZAu9ip:-NoqEdeV7GVNZpjWh839:blr7wnb) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1052)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

This sounds more effective, I am not sure how we can use a binary search for the last byte, but I am using the shr strategy on the last byte, which should save a lot of steps.

I am using this technique in `bit_len`, and hence making this function return ` 64 - self.bit_len()`
</blockquote></details>

you can do a binary search similarly to how it's done in the `bytes_used` function, by checking whether the result is inferior to a bound. `<2` means 1 bit used, `<4` means 2 bits used, `<8` means 3 bits used, otherwise 4 bits used
___
*[`crates/utils/src/helpers.cairo` line 697 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqDm1a1Xt_BPuIZt_A:-NoqDm1a1Xt_BPuIZt_B:b-u2wwig) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L697)):*
> ```Cairo
>     ///  let span = array![0x0, 0x01, 0x02, 0x03, 0x04, 0x05].span();
>     ///  let expected = array![0x0, 0x0, 0x0, 0x0, 0x0, 0x01, 0x02, 0x03, 0x04, 0x05].span();
>     ///  let result = span.left_padding(10);
> ```

pad_left_with_zeroes



_Code quote:_
```Cairo
left_padding(
```
___
*[`crates/utils/src/helpers.cairo` line 901 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqEPWn4DDmgT5gPcmk:-NoqEPWn4DDmgT5gPcml:blx1pes) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L901)):*
> ```Cairo
>     /// # Returns
>     /// * The bytes representation of the value.
>     fn to_be_bytes(mut self: u64) -> Span<u8> {
> ```

How necessary is the signature change? If you return an array you can always modify it and append stuff at the end, but not if you return a span
___
*[`crates/utils/src/helpers.cairo` line 971 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqIxRK5clG-riXQw1H:-NoqIxRK5clG-riXQw1I:b-pwbym4) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L971)):*
> ```Cairo
>             };
>             count += 1;
>             n = n.shr(1);
> ```

since multiplication is cheaper than division:
what do you think of, instead of shifting right and doing a BitAnd (1), simply multiplying the mask by 2 at each iteration?

e.g. 0000 1000
start with mask = 1
0000 1000 & 1 => 0
0000 1000 & 10 => 0
0000 1000 * 100 => 0
0000 1000 *& 1000 => 1

you get trailing_zeroes = 3
___
*[`crates/utils/src/helpers.cairo` line 1592 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqH5wQEwYOC-BQT0Qc:-NoqH5wQEwYOC-BQT0Qd:b-ldlen2) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L1592)):*
> ```Cairo
> }
> 
> impl BitLengthTraitImpl<
> ```

super nice :heart_eyes: 
___
*[`crates/utils/src/helpers.cairo` line 1618 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqHDEzAiKdw7mCtg_O:-NoqHDEzAiKdw7mCtg_P:bw6vw2q) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L1618)):*
> ```Cairo
>         /// Count the number of bits in the last byte
>         let mut n = last_byte;
>         loop {
> ```

see comment above with binary search using comparison

enitrat (commented):
Can you remove the Felt252Vec-related features so that we can merge after this round? :pray: 

> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-:-NoqJTp6C17S3Va1HjB2:b-8vp47e)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @bajpai244 and @ClementWalter)

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 652 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NollWuE24JjN0tSREXq:-Noqj8KRAac9qgGeCHR4:bekry7p) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L652)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I still think the logic of having `start + len` is not correct. Isn't this simply
`cmp::min(start.saturating_add(len), self.len()`?
</blockquote></details>

I am not sure how it is logically not correct, as it does return the min value, but wasn't aware that cmp::min existed when I was writing this code, so yeah this make it easier, refactored.

Done.
___
*[`crates/utils/src/helpers.cairo` line 697 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqDm1a1Xt_BPuIZt_A:-NoqjdXz4hR9t-YvoAas:b-umq0ci) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L697)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

pad_left_with_zeroes
</blockquote></details>

Ah, yeah, wrote the comment before updating the name, corrected.
___
*[`crates/utils/src/helpers.cairo` line 901 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqEPWn4DDmgT5gPcmk:-Noqjs9S5o8GIWb5C2to:bz0ym19) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L901)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

How necessary is the signature change? If you return an array you can always modify it and append stuff at the end, but not if you return a span
</blockquote></details>

So, everywhere I was using this function, I was using `.span()`, and I think this function signature is better, till we actually come to use-cases where we need the array rather than the Span, because right now everywhere we call it, we just need the Span.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 971 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqIxRK5clG-riXQw1H:-Noqv6Ki4JMDZyWZxI1N:bd2wmm0) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L971)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

since multiplication is cheaper than division:
what do you think of, instead of shifting right and doing a BitAnd (1), simply multiplying the mask by 2 at each iteration?

e.g. 0000 1000
start with mask = 1
0000 1000 & 1 => 0
0000 1000 & 10 => 0
0000 1000 * 100 => 0
0000 1000 *& 1000 => 1

you get trailing_zeroes = 3
</blockquote></details>

But, we are interested in leading zeroes and not trailing one.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1052 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-Nolsq1KEYDsJFZAu9ip:-NoqvnxYB-whSiTLh8aj:bdbfwah) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/29c485f3205bcfa3e61142488fc8eb830670d2f0/crates/utils/src/helpers.cairo#L1052)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

you can do a binary search similarly to how it's done in the `bytes_used` function, by checking whether the result is inferior to a bound. `<2` means 1 bit used, `<4` means 2 bits used, `<8` means 3 bits used, otherwise 4 bits used
</blockquote></details>

I don't think what we do `bytes_used` can be called a binary search,  I think it is easier to do for number of bytes, but harder to do it for number of bits.

+ I don't think we should use the same technique here, the code we have is simple to read to and small, and even in the worst case will do 8 shr.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1592 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqH5wQEwYOC-BQT0Qc:-NoqxMfH-zZTBdlGHsxL:b-zcjvte) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L1592)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

super nice :heart_eyes:
</blockquote></details>

Thanks
___
*[`crates/utils/src/helpers.cairo` line 1618 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqHDEzAiKdw7mCtg_O:-NoqxO3N37jr7YoxDSfP:b-u7b4ta) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L1618)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

see comment above with binary search using comparison
</blockquote></details>

I have replied on it, I think that this more readable and simple, + even in worst case will do 8 shr.

bajpai244 (commented):
removed.

> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-:-Noqx_Ox4JSzw5dxLkwO:brpmqc3)* status: 9 of 11 files reviewed, 6 unresolved discussions (waiting on @ClementWalter and @enitrat)

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 10 of 11 files reviewed, 3 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 971 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqIxRK5clG-riXQw1H:-NouCsP8A_6p6Sd43Ci1:b6srhj1) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L971)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I don't get it: the name of the function is `trailing_zeroes` ?
</blockquote></details>

Ah, mistake on my end, on reviewable I wasn't able to see the function, and thought u are suggesting this for leaning zeroes, fixed.
___
*[`crates/utils/src/helpers.cairo` line 1618 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqHDEzAiKdw7mCtg_O:-NouE_HhAYmCH4AF0DW9:b-l7h6hd) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L1618)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

> Binary Search is defined as **a searching algorithm used in a sorted array by repeatedly dividing the search interval in half**.

It is a sort of binary search :)

> I think it is easier to do for number of bytes, but harder to do it for number of bits.

How is it different? :thinking: 

> and even in the worst case will do 8 shr.

which remains a more expensive than doing 2 range checks for no overhead in complexity given that this logic is already used previously

```rust
        if self < 0b0100 {
            if self < 0b0010 {
                if self == 0 {
                    return 0;
                } else {
                    return 1;
                };
            }
            return 2;
        } else {
            if self < 0b1000 {
                return 3;
            } else {
                return 4;
            }
        };
```
</blockquote></details>

Hmm, this only seems to be able to do it till 4 bits, what if the u8 is `255`? The algorithm will return `4` for it, whereas the answer is `8`.

We would have to add more conditions here and I don't find this way of doing it elegant, if u still think that adding for all those conditions is better than what we have then I will do it.

enitrat (changes requested):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: all files reviewed, 1 unresolved discussion (waiting on @bajpai244 and @ClementWalter)

___
*[`crates/utils/src/helpers.cairo` line 1618 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqHDEzAiKdw7mCtg_O:-NouVoQ_-ylsqc9FgGwN:b-8qkf9y) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L1618)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

Hmm, this only seems to be able to do it till 4 bits, what if the u8 is `255`? The algorithm will return `4` for it, whereas the answer is `8`.

We would have to add more conditions here and I don't find this way of doing it elegant, if u still think that adding for all those conditions is better than what we have then I will do it.
</blockquote></details>

yeah, you need another branch of binary search for answers between 5 and 8, please do it

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706)* status: 9 of 11 files reviewed, 1 unresolved discussion (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1618 at r11](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-NoqHDEzAiKdw7mCtg_O:-NoutVkzEPuE5aSY7CYs:b-896fix) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/eaf8531ccb2968d86f8a8e4fc6ce8b7c1ed79001/crates/utils/src/helpers.cairo#L1618)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

yeah, you need another branch of binary search for answers between 5 and 8, please do it
</blockquote></details>

Done.

enitrat (approved):
lgtm

> Reviewed 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/706#-:-NovgASB4CP5PfWtzyrX:bb7k0qh)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ClementWalter)

Issue #707, opened 2024-01-22
bug: is_account_alive


enitrat:
## Bug Report

**Kakarot version:**

**Current behavior:**
Reads the local state to check whether an account is alive or not - returns false if the account is not in the local state, meaning not cached during the transaction

**Expected behavior:**
Should return `false` only if the account doesn't exist _at all_, meaning has no code, nonce nor balance _in the global state_

TAdev0:
@enitrat  isn't it already the case?

```
 fn is_account_alive(ref self: State, evm_address: EthAddress) -> bool {
        let account = self.accounts.read(evm_address.into());
        match account {
            Option::Some(acc) => {
                return !(acc.nonce == 0 && acc.code.len() == 0 && acc.balance == 0);
            },
            Option::None => { return false; }
        }
    }
    ```

the function checs for nonce, codesize and balance. But i guess the ref self is a local state right? how can we grab the global state?

TAdev0:
does it mean kakarot core storage needs to be modified, to store all Account structs for all ethAddress in it?

enitrat:
if you call `get_account` you will get an account object that reflects its "true" state at the current time, not only the cached one

TAdev0:
oh yeah i get it, with `fetch_or_create` , you reconstruct the Account struct corresponding to the evm_address

TAdev0:
why? if you call `get_account`,  the return value is `self.accounts.read(evm_address.into()).unwrap()` if it is in the cache,
and `is_account_alive` does the same

TAdev0:
@enitrat  ok its all clear now. Spent time to understand how bytecode and nonce were stored in contracts, using low level methods.

PR #708, opened 2024-01-22
bug: fix `is_account_alive` function

TAdev0:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #707 

## What is the new behavior?

`is_account_alive` checks in the global state if the contract account exists.

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/708)

### DIFF
--- a/crates/evm/src/state.cairo
+++ b/crates/evm/src/state.cairo
@@ -206,13 +206,8 @@ impl StateImpl of StateTrait {
 
     // Check whether is an account is both in the state and non empty.
     fn is_account_alive(ref self: State, evm_address: EthAddress) -> bool {
-        let account = self.accounts.read(evm_address.into());
-        match account {
-            Option::Some(acc) => {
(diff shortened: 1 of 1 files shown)


crates/evm/src/state.cairo
```diff
@@ -206,13 +206,8 @@ impl StateImpl of StateTrait {
 
     // Check whether is an account is both in the state and non empty.
```
enitrat, line 207: """
    Check whether is an account is both in the state and non empty.

    Parameters
    ----------
    state:
        The state
    address:
        Address of the account that needs to be checked.

   Returns
    -------
    is_alive : `bool`
        True if the account is alive.
    """
    
    
    can you update the function doc to the equivalent of this, perhaps replacing "in the state" which is ambiguous for us by 'in the global state'

TAdev0:
@enitrat  ready to be merged

enitrat (approved):

PR #710, opened 2024-01-24
Add a Precompiles contract

ClementWalter:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Precompiles are available within an EVM run only.

Resolves: https://github.com/kkrt-labs/kakarot/issues/909

## What is the new behavior?

Added a Precompiles contract with one `exec_precompile` entrypoint to run a precompile.

Note: I've just put the SHA256 has it's the only one missing currently for Kakarot0. I don't know if it's worth adding them all.

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/710)

### DIFF
--- a/crates/contracts/src/precompiles.cairo
+++ b/crates/contracts/src/precompiles.cairo
@@ -1,0 +1,31 @@
+#[starknet::interface]
+trait IPrecompiles<T> {
+    fn exec_precompile(self: @T, address: felt252, data: Span<u8>) -> Span<u8>;
+}
+
+#[starknet::contract]
+mod Precompiles {
+    use alexandria_math::sha256::sha256;
+
+    use super::IPrecompiles;
+
+    #[storage]
+    struct Storage {}
+
+    #[abi(embed_v0)]
+    impl Precompiles of IPrecompiles<ContractState> {
+        fn exec_precompile(self: @ContractState, address: felt252, data: Span<u8>) -> Span<u8> {
+            let result = match address {
+                0 => panic!("Precompile address can't be 0"),
+                1 => panic!("Precompile ecRecover not available"),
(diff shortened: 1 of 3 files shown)


bajpai244:
@ClementWalter aftrer we merge #712, we can also add ModExp alongwith SHA256, as from what I know, the ModExp of KKRT-0 isn't arbitrary precision!

bajpai244:
@ClementWalter You need to do:
```
scarb fmt
```

and then push.

crates/contracts/src/precompiles.cairo
```diff
@@ -0,0 +1,31 @@
...
+
+    #[abi(embed_v0)]
+    impl Precompiles of IPrecompiles<ContractState> {
+        fn exec_precompile(self: @ContractState, address: felt252, data: Span<u8>) -> Span<u8> {
+            let result = match address {
+                0 => panic!("Precompile address can't be 0"),
+                1 => panic!("Precompile ecRecover not available"),
+                2 => {
+                    let mut input = array![];
+                    input.append_span(data);
```
enitrat, line 23: you're doing a deep clone of `data` just to have it as an `Array<u8>` (mutable) instead of `Span<u8>` (immutable). You can just change the input type of the function `data :Span<u8>` to `data: Array<u8>` (in the interface as well), then call `sha256(data).span()`

ClementWalter: I noticed it but I thought that inputs where by design immutable (like `calldata` storage in solidity is immutable)

enitrat (approved):

Issue #713, opened 2024-01-25
feat: add Felt252Vec utils


bajpai244:
### Describe the Feature Request

Add Felt252Vec utils used in #699.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [x] Yes
- [ ] No

PR #714, opened 2024-01-25
feat: add vec utils

bajpai244:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Adds Felt252Vec utils used in #699.

Resolves: #713 

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)

### DIFF
--- a/crates/alexandria_data_structures/src/vec.cairo
+++ b/crates/alexandria_data_structures/src/vec.cairo
@@ -63,9 +63,16 @@ impl VecIndex<V, T, impl VecTraitImpl: VecTrait<V, T>> of Index<V, usize, T> {
     }
 }
 
+
 struct Felt252Vec<T> {
     items: Felt252Dict<T>,
     len: usize,
+}
+
+impl DefaultFeltVec<T, +Drop<T>, +Copy<T>, +Felt252DictValue<T>> of Default<Felt252Vec<T>> {
+    fn default() -> Felt252Vec<T> {
+        Felt252VecImpl::<T>::new()
+    }
 }
 
 impl DestructFeltVec<T, +Drop<T>, +Felt252DictValue<T>> of Destruct<Felt252Vec<T>> {
--- a/crates/evm/src/errors.cairo
+++ b/crates/evm/src/errors.cairo
@@ -1,5 +1,5 @@
 use core::fmt::{Debug, Formatter, Error, Display};
-use utils::helpers::U256Trait;
+use utils::helpers::{U256Trait, ToBytes};
 
 // STACK
 
--- a/crates/evm/src/precompiles/blake2f.cairo
+++ b/crates/evm/src/precompiles/blake2f.cairo
@@ -5,7 +5,7 @@ use evm::errors::{EVMError, ensure};
 use evm::model::vm::{VM, VMTrait};
 use starknet::EthAddress;
 use utils::crypto::blake2_compress::compress;
-use utils::helpers::{U32Trait, U64Trait};
+use utils::helpers::{U32Trait, U64Trait, ToBytes};
 
 const GF_ROUND: u64 = 1;
 const INPUT_LENGTH: usize = 213;
--- a/crates/evm/src/precompiles/ec_recover.cairo
+++ b/crates/evm/src/precompiles/ec_recover.cairo
@@ -7,7 +7,7 @@ use starknet::{
     secp256k1::{Secp256k1Point}
 };
 
-use utils::helpers::{U256Trait, BoolIntoNumeric};
+use utils::helpers::{U256Trait, BoolIntoNumeric, ToBytes};
 use utils::traits::EthAddressIntoU256;
 
 const EC_RECOVER_PRECOMPILE_GAS_COST: u128 = 3000;
@@ -67,7 +67,7 @@ impl EcRecoverPrecompileTraitImpl of EcRecoverPrecompileTrait {
         };
 
         let eth_address: u256 = public_key_point_to_eth_address(recovered_public_key).into();
-        let eth_address = eth_address.to_padded_bytes();
+        let eth_address = eth_address.to_be_bytes_padded();
 
         vm.return_data = eth_address;
 
--- a/crates/evm/src/tests/test_gas.cairo
+++ b/crates/evm/src/tests/test_gas.cairo
@@ -10,7 +10,7 @@ use utils::eth_transaction::{
     EthereumTransaction, LegacyTransaction, AccessListTransaction, EthereumTransactionTrait,
     AccessListItem
 };
-use utils::helpers::U256Trait;
+use utils::helpers::{U256Trait, ToBytes};
 
 #[test]
 fn test_calculate_intrinsic_gas_cost() {
--- a/crates/evm/src/tests/test_precompiles/test_ec_recover.cairo
+++ b/crates/evm/src/tests/test_precompiles/test_ec_recover.cairo
@@ -7,7 +7,8 @@ use evm::memory::MemoryTrait;
 use evm::precompiles::ec_recover::EcRecoverPrecompileTrait;
 use evm::stack::StackTrait;
 use evm::tests::test_utils::{VMBuilderTrait, native_token, other_starknet_address};
-use utils::helpers::U256Trait;
+use utils::helpers::{U256Trait, ToBytes};
+
 
 // source: <https://www.evm.codes/playground?unit=Wei&codeType=Mnemonic&code='jFirsNplace_parameters%20in%20memoryZ456e9aea5e197a1f1af7a3e85a3212fa4049a3ba34c2289b4c860fc0b0c64ef3whash~Y~28wvX2YZ9242685bf161793cc25603c231bc2f568eb630ea16aa137d2664ac8038825608wrX4YZ4f8ae3bd7535248d0bd448298cc2e2071e56992d0774dc340c368ae950852adawsX6YqqjDo_call~32JSizeX80JOffsetX8VSize~VOffset~1waddressW4QFFFFFFFFwgasqSTATICCALLqqjPut_resulNalonKon_stackqPOPX80qMLOAD'~W1%20w%20jq%5Cnj//%20_%20thKZW32QY0qMSTOREX~0xWqPUSHV0wargsQ%200xNt%20Ke%20Jwret%01JKNQVWXYZ_jqw~_>
 #[test]
@@ -16,16 +17,13 @@ fn test_ec_recover_precompile() {
 
     let mut vm = VMBuilderTrait::new_with_presets().build();
 
-    let msg_hash = U256Trait::to_padded_bytes(
-        0x456e9aea5e197a1f1af7a3e85a3212fa4049a3ba34c2289b4c860fc0b0c64ef3
-    );
-    let v = U256Trait::to_padded_bytes(28);
-    let r = U256Trait::to_padded_bytes(
-        0x9242685bf161793cc25603c231bc2f568eb630ea16aa137d2664ac8038825608
-    );
-    let s = U256Trait::to_padded_bytes(
-        0x4f8ae3bd7535248d0bd448298cc2e2071e56992d0774dc340c368ae950852ada
-    );
+    let msg_hash = 0x456e9aea5e197a1f1af7a3e85a3212fa4049a3ba34c2289b4c860fc0b0c64ef3_u256
+        .to_be_bytes_padded();
+    let v = 28_u256.to_be_bytes_padded();
+    let r = 0x9242685bf161793cc25603c231bc2f568eb630ea16aa137d2664ac8038825608_u256
+        .to_be_bytes_padded();
+    let s = 0x4f8ae3bd7535248d0bd448298cc2e2071e56992d0774dc340c368ae950852ada_u256
+        .to_be_bytes_padded();
 
     let mut calldata = array![];
     calldata.append_span(msg_hash);
--- a/crates/utils/src/address.cairo
+++ b/crates/utils/src/address.cairo
@@ -4,7 +4,7 @@ use core::traits::TryInto;
 use evm::errors::EVMError;
 use starknet::EthAddress;
 use utils::errors::RLPErrorTrait;
-use utils::helpers::{U8SpanExTrait, U64Trait, U256Trait, EthAddressExTrait, ArrayExtTrait};
+use utils::helpers::{U8SpanExTrait, U64Trait, U256Trait, EthAddressExTrait, ArrayExtTrait, ToBytes};
 use utils::math::WrappingBitshift;
 use utils::rlp::{RLPTrait, RLPItem};
 use utils::traits::{TryIntoResult};
@@ -44,11 +44,11 @@ fn compute_contract_address(sender_address: EthAddress, sender_nonce: u64) -> Et
 fn compute_create2_contract_address(
     sender_address: EthAddress, salt: u256, bytecode: Span<u8>
 ) -> Result<EthAddress, EVMError> {
-    let hash = bytecode.compute_keccak256_hash().to_padded_bytes();
+    let hash = bytecode.compute_keccak256_hash().to_be_bytes_padded();
 
     let sender_address = sender_address.to_bytes().span();
 
-    let salt = salt.to_padded_bytes();
+    let salt = salt.to_be_bytes_padded();
 
     let mut preimage: Array<u8> = array![];
 
@@ -57,7 +57,7 @@ fn compute_create2_contract_address(
     preimage.append_span(salt);
     preimage.append_span(hash);
 
-    let address_hash = preimage.span().compute_keccak256_hash().to_padded_bytes();
+    let address_hash = preimage.span().compute_keccak256_hash().to_be_bytes_padded();
 
     let address: EthAddress = address_hash.slice(12, 20).try_into_result()?;
 
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -1819,39 +1819,7 @@ impl Felt252VecTraitImpl<
     /// * `new_len` the new length of the Felt252Vec
     /// * `value` the value to fill the new elements with
     fn resize(ref self: Felt252Vec<T>, new_len: usize, value: T) {
-        if self.len() == new_len {
-            return;
-        }
-
-        let current_len = self.len();
-        let diff = current_len - new_len;
-        if current_len < new_len {
-            let mut i = 0;
-            loop {
-                if i == diff {
-                    break;
-                }
-
-                self.push(value);
-                i += 1;
-            };
-
-            return;
-        };
-
-        let mut new_vec = Default::default();
-        let mut i = 0;
-        loop {
-            if i == new_len {
-                break;
-            }
-
-            new_vec.push(self[i]);
-
-            i += 1;
-        };
-
-        self = new_vec;
+        self.len = new_len;
     }
 
 
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -1505,6 +1505,16 @@ trait BytesUsedTrait<T> {
     fn bytes_used(self: T) -> u8;
 }
 
+impl U8BytesUsedTraitImpl of BytesUsedTrait<u8> {
+    fn bytes_used(self: u8) -> u8 {
+        if self == 0 {
+            return 0;
+        }
+
+        return 1;
+    }
+}
+
 impl USizeBytesUsedTraitImpl of BytesUsedTrait<usize> {
     fn bytes_used(self: usize) -> u8 {
         if self < 0x10000 { // 256^2
@@ -1626,31 +1636,6 @@ enum Felt252VecTraitErrors {
     Overflow,
     LengthIsNotSame,
     SizeLessThanCurrentLength
-}
-
-#[generate_trait]
-impl Felt252VecU8TraitImpl of Felt252VecU8Trait {
-    /// Returns Felt252Vec<u8> as a Span<8>
-    /// The endianess of the bytes will be same as the endianess of the `self`
-    /// # Arguments
-    /// * `self` a ref Felt252Vec<u8>
-    /// # Returns
-    /// * A Span<u8> of len == self.len() with same endianess as `self`
-    fn to_bytes(ref self: Felt252Vec<u8>) -> Span<u8> {
-        let mut arr: Array<u8> = Default::default();
-
-        let mut i = 0;
-        loop {
-            if i == self.len() {
-                break;
-            }
-
-            arr.append(self[i]);
-            i += 1;
-        };
-
-        arr.span()
-    }
 }
 
 #[generate_trait]
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -1857,7 +1857,7 @@ impl Felt252VecTraitImpl<
             return;
         };
 
-        let mut new_vec = Felt252VecImpl::new();
+        let mut new_vec = Default::default();
         let mut i = 0;
         loop {
             if i == new_len {
@@ -1971,7 +1971,7 @@ impl Felt252VecTraitImpl<
     /// * `input` a ref Felt252Vec<T>
     /// Note: this is an expensive operation, as it will create a new Felt252Vec
     fn remove_trailing_zeroes(ref self: Felt252Vec<T>) {
-        let mut vec: Felt252Vec<T> = Felt252VecImpl::new();
+        let mut vec: Felt252Vec<T> = Default::default();
 
         let mut i = self.len;
         let mut num_of_trailing_zeroes = 0;
@@ -2016,7 +2016,7 @@ impl Felt252VecTraitImpl<
             return Option::None;
         }
 
-        let mut new_vec = Felt252VecImpl::new();
+        let mut new_vec = Default::default();
         let popped_ele = self[self.len() - 1];
 
         let mut i = 0;
@@ -2040,7 +2040,7 @@ impl Felt252VecTraitImpl<
     /// # Returns
     /// * A new Felt252Vec<T> with the same elements
     fn duplicate(ref self: Felt252Vec<T>) -> Felt252Vec<T> {
-        let mut new_vec = Felt252VecImpl::new();
+        let mut new_vec = Default::default();
 
         let mut i: u32 = 0;
 
@@ -2077,7 +2077,7 @@ impl Felt252VecTraitImpl<
             return Option::None;
         };
 
-        let mut new_vec = Felt252VecImpl::new();
+        let mut new_vec = Default::default();
 
         let mut i: u32 = 0;
 
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -1698,12 +1698,18 @@ impl Felt252VecTraitImpl<
     +Into<u8, T>,
     +PartialEq<T>
 > of Felt252VecTrait<T> {
-    /// Expands a Felt252Vec to a new length by appending zeroes, if the new length is less than the current length, it will return an error
+    /// Expands a Felt252Vec to a new length by appending zeroes
+    ///
+    /// This function will mutate the Felt252Vec in-place and will expand its length to the new length by appending zeroes.
+    /// If the new length is less than the current length, it will return an error.
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `new_length` the new length of the Felt252Vec
+    ///
     /// # Returns
     /// * Result::<(), Felt252VecTraitErrors>
+    ///
     /// # Errors
     /// * Felt252VecTraitErrors::SizeLessThanCurrentLength if the new length is less than the current length
     fn expand(ref self: Felt252Vec<T>, new_length: usize) -> Result<(), Felt252VecTraitErrors> {
@@ -1717,6 +1723,7 @@ impl Felt252VecTraitImpl<
     }
 
     /// Sets all elements of the Felt252Vec to zero, mutates the Felt252Vec in-place
+    ///
     /// # Arguments
     /// self a ref Felt252Vec<T>
     fn reset(ref self: Felt252Vec<T>) {
@@ -1726,8 +1733,10 @@ impl Felt252VecTraitImpl<
     }
 
     /// Returns the leading zeroes of a Felt252Vec<T>
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
+    ///
     /// # Returns
     /// * The number of leading zeroes in `self`.
     fn count_leading_zeroes(ref self: Felt252Vec<T>) -> usize {
@@ -1744,7 +1753,10 @@ impl Felt252VecTraitImpl<
     }
 
     /// Resizes the Felt252Vec<T> in-place so that len is equal to new_len.
+    ///
+    /// This function will mutate the Felt252Vec in-place and will resize its length to the new length.
     /// If new_len is greater than len, the Vec is extended by the difference, with each additional slot filled with `value`. If new_len is less than len, the Vec is simply truncated from the right.
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `new_len` the new length of the Felt252Vec
@@ -1755,10 +1767,12 @@ impl Felt252VecTraitImpl<
 
 
     /// Copies the elements from a Span<u8> into the Felt252Vec<T> in little endian format, in case of overflow or index being out of bounds, an error is returned
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `index` the index at `self` to start copying from
     /// * `slice` a Span<u8>
+    ///
     /// # Errors
     /// * Felt252VecTraitErrors::IndexOutOfBound if the index is out of bounds
     /// * Felt252VecTraitErrors::Overflow if the Span is too big to fit in the Felt252Vec
@@ -1789,9 +1803,11 @@ impl Felt252VecTraitImpl<
     }
 
     /// Copies the elements from a Felt252Vec<T> into the Felt252Vec<T> in little endian format, If length of both Felt252Vecs are not same, it will return an error
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `vec` a ref Felt252Vec<T>
+    ///
     /// # Errors
     /// * Felt252VecTraitErrors::LengthIsNotSame if the length of both Felt252Vecs are not same
     fn copy_from_vec_le(
@@ -1815,10 +1831,12 @@ impl Felt252VecTraitImpl<
     }
 
     /// Insert elements of Felt252Vec into another Felt252Vec at a given index, in case of overflow or index being out of bounds, an error is returned
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `idx` the index at `self` to start inserting from
     /// * `vec` a ref Felt252Vec<T>
+    ///
     /// # Errors
     /// * Felt252VecTraitErrors::IndexOutOfBound if the index is out of bounds
     /// * Felt252VecTraitErrors::Overflow if the Felt252Vec is too big to fit in the Felt252Vec
@@ -1849,6 +1867,7 @@ impl Felt252VecTraitImpl<
 
 
     /// Removes trailing zeroes from a Felt252Vec<T>
+    ///
     /// # Arguments
     /// * `input` a ref Felt252Vec<T>
     fn remove_trailing_zeroes(ref self: Felt252Vec<T>) {
@@ -1867,9 +1886,11 @@ impl Felt252VecTraitImpl<
     }
 
     /// Pops an element out of the vector, returns Option::None if the vector is empty
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// # Returns
+    ///
     /// * Option::Some(T), returns the last element or Option::None if the vector is empty
     fn pop(ref self: Felt252Vec<T>) -> Option<T> {
         if (self.len) == 0 {
@@ -1882,9 +1903,12 @@ impl Felt252VecTraitImpl<
     }
 
     /// takes a Felt252Vec<T> and returns a new Felt252Vec<T> with the same elements
+    ///
     /// Note: this is an expensive operation, as it will create a new Felt252Vec
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
+    ///
     /// # Returns
     /// * A new Felt252Vec<T> with the same elements
     fn duplicate(ref self: Felt252Vec<T>) -> Felt252Vec<T> {
@@ -1906,7 +1930,10 @@ impl Felt252VecTraitImpl<
     }
 
     /// Returns a new Felt252Vec<T> with elements starting from `idx` to `idx + len`
+    ///
+    /// This function will start slicing from `idx` and will slice `len` elements, it will firstly clone the elements and then return a new Felt252Vec<T>
     /// In case of overflow return Option::None
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `idx` the index to start slicing from
@@ -1946,9 +1973,11 @@ impl Felt252VecTraitImpl<
     }
 
     /// Returns whether two Felt252Vec<T> are equal after removing trailing_zeroes
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `rhs` a ref Felt252Vec<T>
+    ///
     /// # Returns
     /// * bool, returns true if both Felt252Vecs are equal, false otherwise
     fn equal_remove_trailing_zeroes(ref self: Felt252Vec<T>, ref rhs: Felt252Vec<T>) -> bool {
@@ -1978,13 +2007,16 @@ impl Felt252VecTraitImpl<
 
     /// Fills a Felt252Vec<T> with a given `value` starting from `start_idx` to `start_idx + len`
     /// In case of index out of bounds or overflow, error is returned
+    ///
     /// # Arguments
     /// * `self` a ref Felt252Vec<T>
     /// * `start_idx` the index to start filling from
     /// * `len` the length of the fill
     /// * `value` the value to fill the Felt252Vec with
+    ///
     /// # Returns
     /// * Result::<(), Felt252VecTraitErrors>
+    ///
     /// # Errors
     /// * Felt252VecTraitErrors::IndexOutOfBound if the index is out of bounds
     /// * Felt252VecTraitErrors::Overflow if the Felt252Vec is too big to fit in the Felt252Vec
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -1902,9 +1902,10 @@ impl Felt252VecTraitImpl<
             return Result::Err(Felt252VecTraitErrors::Overflow);
         }
 
+        let stop = idx + vec.len();
         let mut i = idx;
         loop {
-            if i == idx + vec.len() {
+            if i == stop {
                 break;
             }
 
@@ -1919,10 +1920,7 @@ impl Felt252VecTraitImpl<
     /// Removes trailing zeroes from a Felt252Vec<T>
     /// # Arguments
     /// * `input` a ref Felt252Vec<T>
-    /// Note: this is an expensive operation, as it will create a new Felt252Vec
     fn remove_trailing_zeroes(ref self: Felt252Vec<T>) {
-        let mut vec: Felt252Vec<T> = Default::default();
-
         let mut i = self.len;
         let mut num_of_trailing_zeroes = 0;
         loop {
@@ -1934,25 +1932,7 @@ impl Felt252VecTraitImpl<
             num_of_trailing_zeroes += 1;
         };
 
-        if num_of_trailing_zeroes == 0 {
-            return;
-        }
-
-        let mut i = 0;
-        loop {
-            if i == (self.len - num_of_trailing_zeroes) {
-                break;
-            }
-
-            vec.push(self[i]);
-            i += 1;
-        };
-
-        if vec.len == 0 {
-            vec.push(Zero::zero());
-        };
-
-        self = vec;
+        self.len = self.len - num_of_trailing_zeroes;
     }
 
     /// Pops an element out of the vector, returns Option::None if the vector is empty
@@ -1960,26 +1940,13 @@ impl Felt252VecTraitImpl<
     /// * `self` a ref Felt252Vec<T>
     /// # Returns
     /// * Option::Some(T), returns the last element or Option::None if the vector is empty
-    /// Note: this is an expensive operation, as it will create a new Felt252Vec
     fn pop(ref self: Felt252Vec<T>) -> Option<T> {
         if (self.len) == 0 {
             return Option::None;
         }
 
-        let mut new_vec = Default::default();
         let popped_ele = self[self.len() - 1];
-
-        let mut i = 0;
-        loop {
-            if i == self.len - 1 {
-                break;
-            }
-            new_vec.push(self[i]);
-            i += 1;
-        };
-
-        self = new_vec;
-
+        self.len = self.len - 1;
         Option::Some(popped_ele)
     }
 
--- a/crates/utils/src/helpers.cairo
+++ b/crates/utils/src/helpers.cairo
@@ -1664,3 +1664,503 @@ impl BitLengthTraitImpl<
         BitSize::<T>::bits() - self.bit_len()
     }
 }
+
+#[derive(Drop, Debug, PartialEq)]
+enum Felt252VecTraitErrors {
+    IndexOutOfBound,
+    Overflow,
+    LengthIsNotSame,
+    SizeLessThanCurrentLength
+}
+
+#[generate_trait]
+impl Felt252VecU8TraitImpl of Felt252VecU8Trait {
+    /// Returns Felt252Vec<u8> as a Span<8>
+    /// The endianess of the bytes will be same as the endianess of the `self`
+    /// # Arguments
+    /// * `self` a ref Felt252Vec<u8>
+    /// # Returns
+    /// * A Span<u8> of len == self.len() with same endianess as `self`
+    fn to_bytes(ref self: Felt252Vec<u8>) -> Span<u8> {
+        let mut arr: Array<u8> = Default::default();
+
+        let mut i = 0;
+        loop {
+            if i == self.len() {
+                break;
+            }
+
+            arr.append(self[i]);
+            i += 1;
+        };
+
+        arr.span()
+    }
+}
+
+#[generate_trait]
+impl Felt252VecU64TraitImpl of Felt252VecU64Trait {
+    /// Returns Felt252Vec<u64> as a Span<8>, Where self is in little endian format, and the returned Span is in big endian format
+    /// # Arguments
+    /// * `self` a ref Felt252Vec<u64>
+    /// # Returns
+    /// * A Span<u8> representing bytes conversion of `self` in big endian format
+    fn from_le_to_be_bytes(ref self: Felt252Vec<u64>) -> Span<u8> {
+        let mut res: Array<u8> = array![];
+
+        self.remove_trailing_zeroes();
+
+        let mut i = self.len();
+
+        loop {
+            if i == 0 {
+                break;
+            }
+
+            let j = i - 1;
+
+            res.append_span(self[j].to_be_bytes_padded());
+
+            i -= 1;
+        };
+
+        res.span()
+    }
+
+    /// Returns Felt252Vec<u64> as a Span<8>, Where self is in little endian format, and the returned Span is in little endian format
+    /// # Arguments
+    /// * `self` a ref Felt252Vec<u64>
+    /// # Returns
+    /// * A Span<u8> representing bytes conversion of `self` in little endian format
+    fn from_le_to_le_bytes(ref self: Felt252Vec<u64>) -> Span<u8> {
+        let mut res: Array<u8> = array![];
+        let mut i = 0;
+
+        loop {
+            if i == self.len() {
+                break;
+            }
+
+            if self[i] == 0 {
+                res.append(0);
+            } else {
+                res.append_span(self[i].to_le_bytes());
+            }
+
+            i += 1;
+        };
+
+        res.span()
+    }
+}
+
+#[generate_trait]
+impl Felt252VecTraitImpl<
+    T,
+    +Drop<T>,
+    +Copy<T>,
+    +Felt252DictValue<T>,
+    +Zero<T>,
+    +Add<T>,
+    +Sub<T>,
+    +Div<T>,
+    +Mul<T>,
+    +Exponentiation<T>,
+    +PartialOrd<T>,
+    +Into<u8, T>,
+    +PartialEq<T>
+> of Felt252VecTrait<T> {
+    /// Expands a Felt252Vec to a new length by appending zeroes, if the new length is less than the current length, it will return an error
+    /// # Arguments
+    /// * `self` a ref Felt252Vec<T>
+    /// * `new_length` the new length of the Felt252Vec
+    /// # Returns
+    /// * Result::<(), Felt252VecTraitErrors>
+    /// # Errors
+    /// * Felt252VecTraitErrors::SizeLessThanCurrentLength if the new length is less than the current length
+    fn expand(ref self: Felt252Vec<T>, new_length: usize) -> Result<(), Felt252VecTraitErrors> {
+        if (new_length < self.len) {
+            return Result::Err(Felt252VecTraitErrors::SizeLessThanCurrentLength);
+        };
+
+        let mut i = new_length - self.len();
+        loop {
+            if i == 0 {
+                break;
+            }
+
+            self.push(Zero::zero());
+            i -= 1;
+        };
+
+        Result::Ok(())
+    }
+
+    /// Sets all elements of the Felt252Vec to zero, mutates the Felt252Vec in-place
+    /// # Arguments
+    /// self a ref Felt252Vec<T>
+    fn reset(ref self: Felt252Vec<T>) {
+        let mut i = 0;
+
+        loop {
+            if i == self.len() {
+                break;
+            };
+
+            self.set(i, Zero::zero());
+
+            i += 1;
+        };
+    }
+
+    /// Returns the leading zeroes of a Felt252Vec<T>
+    /// # Arguments
+    /// * `self` a ref Felt252Vec<T>
+    /// # Returns
+    /// * The number of leading zeroes in `self`.
+    fn count_leading_zeroes(ref self: Felt252Vec<T>) -> usize {
+        let mut i = 0;
+        loop {
+            if i == self.len || self[i] != Zero::zero() {
+                break;
+            }
+
+            i += 1;
+        };
+
+        i
+    }
+
+    /// Resizes the Felt252Vec<T> in-place so that len is equal to new_len.
+    /// If new_len is greater than len, the Vec is extended by the difference, with each additional slot filled with `value`. If new_len is less than len, the Vec is simply truncated from the right.
+    /// # Arguments
+    /// * `self` a ref Felt252Vec<T>
+    /// * `new_len` the new length of the Felt252Vec
+    /// * `value` the value to fill the new elements with
+    fn resize(ref self: Felt252Vec<T>, new_len: usize, value: T) {
+        if self.len() == new_len {
+            return;
+        }
+
+        let s = self.len();
+        if s < new_len {
+            let mut i = 0;
+            loop {
+                if i == new_len - s {
+                    break;
+                }
+
+                self.push(value);
+                i += 1;
+            };
+
+            return;
+        };
+
+        let mut new_vec = Felt252VecImpl::new();
+        let mut i = 0;
+        loop {
+            if i == new_len {
+                break;
+            }
+
+            new_vec.push(self[i]);
+
+            i += 1;
+        };
+
+        self = new_vec;
+    }
+
+
+    /// Copies the elements from a Span<u8> into the Felt252Vec<T> in little endian format, in case of overflow or index being out of bounds, an error is returned
+    /// # Arguments
+    /// * `self` a ref Felt252Vec<T>
+    /// * `index` the index at `self` to start copying from
+    /// * `slice` a Span<u8>
+    /// # Errors
+    /// * Felt252VecTraitErrors::IndexOutOfBound if the index is out of bounds
+    /// * Felt252VecTraitErrors::Overflow if the Span is too big to fit in the Felt252Vec
+    fn copy_from_bytes_le(
+        ref self: Felt252Vec<T>, index: usize, mut slice: Span<u8>
+    ) -> Result<(), Felt252VecTraitErrors> {
+        if (index >= self.len) {
+            return Result::Err(Felt252VecTraitErrors::IndexOutOfBound);
+        }
+
+        if ((slice.len() + index) > self.len()) {
+            return Result::Err(Felt252VecTraitErrors::Overflow);
+        }
+
+        let mut i = index;
+        loop {
+            let val = slice.pop_front();
+            if val.is_none() {
+                break;
+            }
+
+            // safe unwrap, as in case of none, we will never reach this branch
+            self.set(i, (*(val.unwrap())).into());
+            i += 1;
+        };
+
+        Result::Ok(())
+    }
+
+    /// Copies the elements from a Felt252Vec<T> into the Felt252Vec<T> in little endian format, If length of both Felt252Vecs are not same, it will return an error
+    /// # Arguments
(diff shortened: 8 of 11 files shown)


enitrat (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 1 of 2 files reviewed, 24 unresolved discussions (waiting on @bajpai244 and @ClementWalter)

___
*[`crates/utils/src/helpers.cairo` line 1678 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozsmq42hwGnK4Mdpig:-Nozsmq42hwGnK4Mdpih:b-g1sz9v) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1678)):*
> ```Cairo
> #[generate_trait]
> impl Felt252VecU8TraitImpl of Felt252VecU8Trait {
>     /// Returns Felt252Vec<u8> as a Span<8>
> ```

I think this should be either Into::< Span<u8>,Felt252Vec<u8>> or From::<Span<u8>,Felt252Vec<u8>> no ?

Also: when is this going to be used and why is the conversion required? If it's only a problem of the type system inside the crypto library, if we make a generic collection trait, this will be avoided. If it needs to be part of the data model, then yes there will be a requirement for a conversion. But we must avoid deep clones as much as possible
___
*[`crates/utils/src/helpers.cairo` line 1684 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-NozsDhpC2GbvsH7YBsK:-NozsDhpC2GbvsH7YBsL:b-ize6ah) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1684)):*
> ```Cairo
>     /// # Returns
>     /// * A Span<u8> of len == self.len() with same endianess as `self`
>     fn to_bytes(ref self: Felt252Vec<u8>) -> Span<u8> {
> ```

Just a thought for now: this is exactly the same as `to_bytes` on an Array. We could use the trait system to implement this over a generic type T representing a Collection, that Array and Vec would implement. Will think about it later once the core concerns of this PR are addressed
___
*[`crates/utils/src/helpers.cairo` line 1702 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozub_S7l4a4iq6hzm2:-Nozub_S7l4a4iq6hzm3:bkwfgi0) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1702)):*
> ```Cairo
> 
> #[generate_trait]
> impl Felt252VecU64TraitImpl of Felt252VecU64Trait {
> ```

I think it would be better if we had a generic Felt252VecExTrait rather than type-specific traits that will need multiple imports
___
*[`crates/utils/src/helpers.cairo` line 1708 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-NozytjeEOw8FyiRfzzx:-NozytjeEOw8FyiRfzzy:bm0e63c) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1708)):*
> ```Cairo
>     /// # Returns
>     /// * A Span<u8> representing bytes conversion of `self` in big endian format
>     fn from_le_to_be_bytes(ref self: Felt252Vec<u64>) -> Span<u8> {
> ```

not a fan of the function name: what do you think of words64_to_be_bytes ?



_Code quote:_
```Cairo
from_le_to_be_bytes
```
___
*[`crates/utils/src/helpers.cairo` line 1720 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozyk5g5IMCLFTKoxMh:-Nozyk5h22xOGCFzzPBd:blaotn7) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1720)):*
> ```Cairo
>             }
> 
>             let j = i - 1;
> ```

i-=1
res.append_span(self[i].to_be_bytes_padded());

no need to have `j` then
___
*[`crates/utils/src/helpers.cairo` line 1735 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozz3AmBQtQK54Rlytj:-Nozz3AmBQtQK54Rlytk:b-ahlhte) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1735)):*
> ```Cairo
>     /// # Returns
>     /// * A Span<u8> representing bytes conversion of `self` in little endian format
>     fn from_le_to_le_bytes(ref self: Felt252Vec<u64>) -> Span<u8> {
> ```

not a fan of the function name: what do you think of words64_to_le_bytes ?
___
*[`crates/utils/src/helpers.cairo` line 1792 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-NozzK4M5qc3Fl6Y2Cqi:-NozzK4M5qc3Fl6Y2Cqj:bsf3wom) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1792)):*
> ```Cairo
>             }
> 
>             self.push(Zero::zero());
> ```

I think this is highly ineffective when we think of the underlying data structure that we're using.

We're using a Dictionary to simulate a mutable data structure. A dictionary has a default value. If I have a dictionary with KV: `{'1': "10"}` and call dict.get(1), I get `10` in return. If I call `dict.get(100)`, I get `0` in return

Therefore, if I just extend the length of the underlying Vec by overriding `self.len`, I am _actually_ going to replicate the behavior of pushing n zeroes for free.
___
*[`crates/utils/src/helpers.cairo` line 1802 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np--lel3Ac8FCpJ69Xg:-Np--lel3Ac8FCpJ69Xh:beq79kl) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1802)):*
> ```Cairo
>     /// # Arguments
>     /// self a ref Felt252Vec<T>
>     fn reset(ref self: Felt252Vec<T>) {
> ```

highly ineffective as well: setting all the elements of the vec to zero is strictly equivalent to replacing the underlying dict with a new one.
___
*[`crates/utils/src/helpers.cairo` line 1845 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-0daP5q5yORv81bpY:-Np-0daP5q5yORv81bpZ:beldjnt) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1845)):*
> ```Cairo
>         }
> 
>         let s = self.len();
> ```

current_len ?
___
*[`crates/utils/src/helpers.cairo` line 1849 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-1686CNx0VNKdGvrs:-Np-1686CNx0VNKdGvrt:b-nv8qlf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1849)):*
> ```Cairo
>             let mut i = 0;
>             loop {
>                 if i == new_len - s {
> ```

stop condition computed outside of loop
___
*[`crates/utils/src/helpers.cairo` line 1860 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-1EgM2hVQPnLIIkRA:-Np-1EgM2hVQPnLIIkRB:b7mq46m) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1860)):*
> ```Cairo
>         };
> 
>         let mut new_vec = Felt252VecImpl::new();
> ```

same as previous comments: instead of doing a deep clone, just reduce the self.len to the desired length
___
*[`crates/utils/src/helpers.cairo` line 1916 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-1gBc6G1BDuFWpPao:-Np-1gBc6G1BDuFWpPap:bi18bwz) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1916)):*
> ```Cairo
>     /// # Errors
>     /// * Felt252VecTraitErrors::LengthIsNotSame if the length of both Felt252Vecs are not same
>     fn copy_from_vec_le(
> ```

what will this be used for? Could iterating over the same data structure, but in help achieve the same result?

I would be extremely careful about any operation that does deep copies so just trying to understand here
___
*[`crates/utils/src/helpers.cairo` line 1929 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-2-KQD7g0lgeK0-DJ:-Np-2-KQD7g0lgeK0-DK:b-5mxbgm) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1929)):*
> ```Cairo
>             }
> 
>             self.set(i, vec[i]);
> ```

if you're effectively overriding every single value held in the `vec` (which should be called `dst`) destination, why is it an input parameter of the function instead of being initialized empty in this function
___
*[`crates/utils/src/helpers.cairo` line 1957 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-2rVhEaZiX_krQbLH:-Np-2rVhEaZiX_krQbLI:bxbk4bn) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1957)):*
> ```Cairo
>         let mut i = idx;
>         loop {
>             if i == idx + vec.len() {
> ```

stop condition computed outside the loop
___
*[`crates/utils/src/helpers.cairo` line 1973 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-3-Rw-0pZJ3914rjT:-Np-3-Rw-0pZJ3914rjU:bq4dhao) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1973)):*
> ```Cairo
>     /// * `input` a ref Felt252Vec<T>
>     /// Note: this is an expensive operation, as it will create a new Felt252Vec
>     fn remove_trailing_zeroes(ref self: Felt252Vec<T>) {
> ```

same remark as previously: just change the `self.len` value to the required new len.
___
*[`crates/utils/src/helpers.cairo` line 2013 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-3HBx9ytpmRVRn9A6:-Np-3HBx9ytpmRVRn9A7:bse047h) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2013)):*
> ```Cairo
>     /// # Returns
>     /// * Option::Some(T), returns the last element or Option::None if the vector is empty
>     /// Note: this is an expensive operation, as it will create a new Felt252Vec
> ```

and magically its not longer expensive :)
___
*[`crates/utils/src/helpers.cairo` line 2031 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-3E2L7_t2lEHkBE3R:-Np-3E2L7_t2lEHkBE3S:b-129928) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2031)):*
> ```Cairo
>         };
> 
>         self = new_vec;
> ```

same, just decrease self.len



_Code quote:_
```Cairo
        let mut i = 0;
        loop {
            if i == self.len - 1 {
                break;
            }
            new_vec.push(self[i]);
            i += 1;
        };

        self = new_vec;
```
___
*[`crates/utils/src/helpers.cairo` line 2042 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-3_bIEBrHwLElvL9Z:-Np-3_bIEBrHwLElvL9_:bok5nc3) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2042)):*
> ```Cairo
>     /// # Returns
>     /// * A new Felt252Vec<T> with the same elements
>     fn duplicate(ref self: Felt252Vec<T>) -> Felt252Vec<T> {
> ```

clone ? (and thus would be in the Clone trait if possible, which im not sure)
___
*[`crates/utils/src/helpers.cairo` line 2069 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-3n7Q9vhlOST-dDiV:-Np-3n7Q9vhlOST-dDiW:b-swzx67) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2069)):*
> ```Cairo
>     /// * Option::Some(Felt252Vec<T>), returns Option::None if the index is out of bounds or if the slice overflows
>     /// Note: this is an expensive operation, as it will create a new Felt252Vec
>     fn slice(ref self: Felt252Vec<T>, idx: usize, len: usize) -> Option<Felt252Vec<T>> {
> ```

clone_slice

all clone / deep copy operations should reflect it in the name



_Code quote:_
```Cairo
slice
```
___
*[`crates/utils/src/helpers.cairo` line 2072 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-44_CCcbyW0Dw1sQ5:-Np-44_CCcbyW0Dw1sQ6:b-w66o6c) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2072)):*
> ```Cairo
>         // Index out of bounds
>         if idx >= self.len {
>             return Option::None;
> ```

shouldnt this panic if done out of bounds (similar to array.slice - let's try to have uniform signatures if this is not detrimental.)
___
*[`crates/utils/src/helpers.cairo` line 2103 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5REVDOUGMz7sPike:-Np-5REVDOUGMz7sPikf:b-rghtj) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2103)):*
> ```Cairo
>     /// # Returns
>     /// * bool, returns true if both Felt252Vecs are equal, false otherwise
>     fn equal(ref self: Felt252Vec<T>, ref rhs: Felt252Vec<T>) -> bool {
> ```

could be in the PartialEq trait ?
___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np-5ZjO1USRrlW6KCtr:b-q50z8q) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
> ```Cairo
>     fn equal(ref self: Felt252Vec<T>, ref rhs: Felt252Vec<T>) -> bool {
>         let mut lhs = self.duplicate();
>         lhs.remove_trailing_zeroes();
> ```

a function named `equal` should compare full vectors, not processed ones after removing trailing_zeroes
___
*[`crates/utils/src/tests/test_helpers.cairo` line 892 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5rciBdAcrKz5qhyc:-Np-5rciBdAcrKz5qhyd:b-l6mwxj) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/tests/test_helpers.cairo#L892)):*
> ```Cairo
> 
> mod felt252_vec_u64_test {
>     #[available_gas(1000000000000)]
> ```

remove
___
*[`crates/utils/src/tests/test_helpers.cairo` line 898 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5tI61aJWcw1LwDCS:-Np-5tI61aJWcw1LwDCT:b-ocmw23) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/tests/test_helpers.cairo#L898)):*
> ```Cairo
>     #[test]
>     fn test_felt252_vec_u64_from_le_to_le_bytes() {
>         let mut vec: Felt252Vec<u64> = Felt252VecImpl::new();
> ```

Default::default();

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 1 of 2 files reviewed, 24 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np-IHQ30GFihDfKb2gy:b-l6p4ef) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

a function named `equal` should compare full vectors, not processed ones after removing trailing_zeroes
</blockquote></details>

So, this is used for tests to compare two Felt252Vec which are supposed to representating two big nums, so I guess we do need the ability to remove trailing_zeroes.

So, lemme rename this function to:
```
equal_remove_trailing_zeroes
```

Which should convey the fact that it will trim the trailing zeroes, and the do the comparison between them as Big Nums.
___
*[`crates/utils/src/tests/test_helpers.cairo` line 892 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5rciBdAcrKz5qhyc:-Np-GS0A1rbox62Yicmw:b-ktq4er) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/tests/test_helpers.cairo#L892)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

remove
</blockquote></details>

strange how this not giving an error on compilation, remvoed
___
*[`crates/utils/src/tests/test_helpers.cairo` line 898 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5tI61aJWcw1LwDCS:-Np-Hyo33GYcVs2NyoA8:bb6un1j) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/tests/test_helpers.cairo#L898)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Default::default();
</blockquote></details>

Done

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 1 of 2 files reviewed, 24 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 2103 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5REVDOUGMz7sPike:-Np-IiW5CHL-6QD_6DCO:bi5gdyl) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2103)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

could be in the PartialEq trait ?
</blockquote></details>

As explained in the previous commit, we need its ability to compare two Vec after removing trailing zeroes, so I guess better if we keep it is a seperate function.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 1 of 2 files reviewed, 24 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1678 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozsmq42hwGnK4Mdpig:-Np-QXTi1s9AgMcaVEor:bq6j5c5) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1678)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I think this should be either Into::< Span<u8>,Felt252Vec<u8>> or From::<Span<u8>,Felt252Vec<u8>> no ?

Also: when is this going to be used and why is the conversion required? If it's only a problem of the type system inside the crypto library, if we make a generic collection trait, this will be avoided. If it needs to be part of the data model, then yes there will be a requirement for a conversion. But we must avoid deep clones as much as possible
</blockquote></details>

So `Into` is defined as

fn into(self: T) -> S

Now, all methods of Felt252Vec<T> are defined on a reference, and not on the type itself, and hence the issue arises when we want to use `Into`
___
*[`crates/utils/src/helpers.cairo` line 1708 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-NozytjeEOw8FyiRfzzx:-Np-Sjig4kgutCUXF_4W:b-7rl0jq) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1708)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

not a fan of the function name: what do you think of words64_to_be_bytes ?
</blockquote></details>

done!
___
*[`crates/utils/src/helpers.cairo` line 1720 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozyk5g5IMCLFTKoxMh:-Np-TQet4wlZNgvTJPaX:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1720)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

i-=1
res.append_span(self[i].to_be_bytes_padded());

no need to have `j` then
</blockquote></details>

done
___
*[`crates/utils/src/helpers.cairo` line 1735 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozz3AmBQtQK54Rlytj:-Np-TSJo4LagOAS3SOIS:b-n06y0k) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1735)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

not a fan of the function name: what do you think of words64_to_le_bytes ?
</blockquote></details>

I can go with it! I guess we can assume that words64 is always little endian.
___
*[`crates/utils/src/helpers.cairo` line 2042 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-3_bIEBrHwLElvL9Z:-Np-NuIN4OdcMuLn4yNP:btwc8f7) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2042)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

clone ? (and thus would be in the Clone trait if possible, which im not sure)
</blockquote></details>

We can't implement clone, as clone is defined over snapshot of a typear
___
*[`crates/utils/src/helpers.cairo` line 2069 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-3n7Q9vhlOST-dDiV:-Np-L5425N_GzW1hIyzK:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2069)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

clone_slice

all clone / deep copy operations should reflect it in the name
</blockquote></details>

done
___
*[`crates/utils/src/helpers.cairo` line 2072 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-44_CCcbyW0Dw1sQ5:-Np-LIfS-3o12Ws_NNP6:b-t91l5s) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2072)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

shouldnt this panic if done out of bounds (similar to array.slice - let's try to have uniform signatures if this is not detrimental.)
</blockquote></details>

we can panic! I am fine with it, I guess everywhere we are unwraps.

done!
___
*[`crates/utils/src/helpers.cairo` line 2103 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5REVDOUGMz7sPike:-Np-Kb4y4CEUz7sA7mRP:b-582rt8) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2103)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

As explained in the previous commit, we need its ability to compare two Vec after removing trailing zeroes, so I guess better if we keep it is a seperate function.
</blockquote></details>

+ ParitalEq is defined on a snapshot, and the get methods doesn't work on snapshots, so we can't implement it.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 0 of 3 files reviewed, 24 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1792 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-NozzK4M5qc3Fl6Y2Cqi:-Np-c58FDXRAltxw-z4r:bwoyrgz) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1792)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I think this is highly ineffective when we think of the underlying data structure that we're using.

We're using a Dictionary to simulate a mutable data structure. A dictionary has a default value. If I have a dictionary with KV: `{'1': "10"}` and call dict.get(1), I get `10` in return. If I call `dict.get(100)`, I get `0` in return

Therefore, if I just extend the length of the underlying Vec by overriding `self.len`, I am _actually_ going to replicate the behavior of pushing n zeroes for free.
</blockquote></details>

Oh! Cool trick, refactored!
___
*[`crates/utils/src/helpers.cairo` line 1802 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np--lel3Ac8FCpJ69Xg:-Np-cWzU8fHNUclCwcjY:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1802)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

highly ineffective as well: setting all the elements of the vec to zero is strictly equivalent to replacing the underlying dict with a new one.
</blockquote></details>

done
___
*[`crates/utils/src/helpers.cairo` line 1845 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-0daP5q5yORv81bpY:-Np-ciMG5Nznc83TeDsH:bb6tzbb) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1845)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

current_len ?
</blockquote></details>

DOne
___
*[`crates/utils/src/helpers.cairo` line 1849 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-1686CNx0VNKdGvrs:-Np-cuywAUjAZYMIp_lF:bmuys85) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1849)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

stop condition computed outside of loop
</blockquote></details>

Done, but is there a reason for this?
___
*[`crates/utils/src/helpers.cairo` line 1860 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-1EgM2hVQPnLIIkRA:-Np-dYIh2qikPKrQh2pu:b-n8v0a0) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1860)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

same as previous comments: instead of doing a deep clone, just reduce the self.len to the desired length
</blockquote></details>

Lol! When I see I could have done that, feels like my whole life was a lie, done!

enitrat (changes requested):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: all files reviewed, 13 unresolved discussions (waiting on @bajpai244 and @ClementWalter)

___
*[`crates/utils/src/helpers.cairo` line 1849 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-1686CNx0VNKdGvrs:-Np-l8Sk4NN-BtXHvhGw:b-trxvlp) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1849)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

Done, but is there a reason for this?
</blockquote></details>

computing N times 1 + 1 is N times more intensive than computing one time 1 + 1
___
*[`crates/utils/src/helpers.cairo` line 2072 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-44_CCcbyW0Dw1sQ5:-Np-m7qz02srFXQSZL8n:b-tig5w6) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2072)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

we can panic! I am fine with it, I guess everywhere we are unwraps.

done!
</blockquote></details>

we can remove this as self[idx + i] will panic with out of bounds anyway
___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np-mU7Q4kK7BFCMnFKp:ble0qk7) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

So, this is used for tests to compare two Felt252Vec which are supposed to representating two big nums, so I guess we do need the ability to remove trailing_zeroes.

So, lemme rename this function to:
```
equal_remove_trailing_zeroes
```

Which should convey the fact that it will trim the trailing zeroes, and the do the comparison between them as Big Nums.
</blockquote></details>

perhaps the sanitizing should be done outside of the function? Otherwise, i'm fine with the renaming. My principal concern here is that you're mutating the object pass in a function that checks equality, and this shouldn't happen.
___
*[`crates/utils/src/helpers.cairo` line 1771 at r3](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-ksodANoBlBRT75jQ:-Np-ksodANoBlBRT75jR:b3okmfw) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/b56f703d243466d2b61255be61b3895ebe485e68/crates/utils/src/helpers.cairo#L1771)):*
> ```Cairo
> > of Felt252VecTrait<T> {
>     /// Expands a Felt252Vec to a new length by appending zeroes, if the new length is less than the current length, it will return an error
>     /// # Arguments
> ```

Can you explain this behavior as a dev note under (skip a line under the description and add it before the arguments)
___
*[`crates/utils/src/helpers.cairo` line 1816 at r3](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-lGX65xpiujjryhvV:-Np-lGX74GWzamlX40rz:b-vzd0da) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/b56f703d243466d2b61255be61b3895ebe485e68/crates/utils/src/helpers.cairo#L1816)):*
> ```Cairo
> 
>     /// Resizes the Felt252Vec<T> in-place so that len is equal to new_len.
>     /// If new_len is greater than len, the Vec is extended by the difference, with each additional slot filled with `value`. If new_len is less than len, the Vec is simply truncated from the right.
> ```

Add dev comment (skip a line and before #arguments)
___
*[`crates/utils/src/helpers.cairo` line 1924 at r3](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-la3V25Gq03cx4Rok:-Np-la3V25Gq03cx4Rol:b-k6klkf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/b56f703d243466d2b61255be61b3895ebe485e68/crates/utils/src/helpers.cairo#L1924)):*
> ```Cairo
>     /// * `input` a ref Felt252Vec<T>
>     fn remove_trailing_zeroes(ref self: Felt252Vec<T>) {
>         let mut i = self.len;
> ```

rename `i` to `new_len`, and then just `decrement `new_len`, at the end `self.len = new_len`
___
*[`crates/utils/src/helpers.cairo` line 1997 at r3](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-mCy29PCH-YdqYpnp:-Np-mCy29PCH-YdqYpnq:bb7od0r) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/b56f703d243466d2b61255be61b3895ebe485e68/crates/utils/src/helpers.cairo#L1997)):*
> ```Cairo
>         // Overflow
>         if (idx + len) > self.len {
>             panic(array!['Overflow']);
> ```

same

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 1 of 10 files reviewed, 13 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 2072 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-44_CCcbyW0Dw1sQ5:-Np0CkWa0KQtBxBm86ZX:bb6un1j) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2072)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

we can remove this as self[idx + i] will panic with out of bounds anyway
</blockquote></details>

Done
___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np0Cyk2ENKf5QeIW3qB:b-i7zbv0) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

perhaps the sanitizing should be done outside of the function? Otherwise, i'm fine with the renaming. My principal concern here is that you're mutating the object pass in a function that checks equality, and this shouldn't happen.
</blockquote></details>

This is used mostly as a testing util,  so I think renaming is fine, in the next PR we see that we don't use it anywhere else apart from testing, I will just refactor this as a test util.
___
*[`crates/utils/src/helpers.cairo` line 1924 at r3](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-la3V25Gq03cx4Rok:-Np0CVh17QpER4crK8Dh:b-p75r0b) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/b56f703d243466d2b61255be61b3895ebe485e68/crates/utils/src/helpers.cairo#L1924)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

rename `i` to `new_len`, and then just `decrement `new_len`, at the end `self.len = new_len`
</blockquote></details>

Smart! Done
___
*[`crates/utils/src/helpers.cairo` line 1997 at r3](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-mCy29PCH-YdqYpnp:-Np0CrwY62Sm2EDdZcO_:bb6un1j) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/b56f703d243466d2b61255be61b3895ebe485e68/crates/utils/src/helpers.cairo#L1997)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

same
</blockquote></details>

Done

enitrat (changes requested):
Great idea to have the ToBytes trait!

> Reviewed 7 of 9 files at r4, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-:-Np0NklW0CGZBXt7Rs9Q:b872ud2)* status: all files reviewed, 14 unresolved discussions (waiting on @bajpai244 and @ClementWalter)

___
*[`crates/utils/src/helpers.cairo` line 1929 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-2-KQD7g0lgeK0-DJ:-Np0NT6ZAWLmIGEXT88h:b-llrzls) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1929)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

This is how I use it as of now.

Now, I can do the same with duplicate as well! But I don't think that would be significantly optimized.


</blockquote></details>

I do not understand how this function is different than a clone (duplicate in your case)
___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np0Lwko3Xd6kTo9cKd8:b-9p4or8) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

This is used mostly as a testing util,  so I think renaming is fine, in the next PR we see that we don't use it anywhere else apart from testing, I will just refactor this as a test util.
</blockquote></details>

If it's used as a test utils and not in the source code, it should be in a `tests/` subdirectory
___
*[`crates/utils/src/helpers.cairo` line 1127 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0Maef-PBTC4kl-J4X:-Np0Maef-PBTC4kl-J4Y:bvwetjw) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1127)):*
> ```Cairo
> > of ToBytes<T> {
>     fn to_be_bytes(self: T) -> Span<u8> {
>         let bytes_used: T = self.bytes_used().into();
> ```

is `into()` required? can we keep bytes_used: u8?
___
*[`crates/utils/src/helpers.cairo` line 1131 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0MR5M-xK2EC5yjKwg:-Np0MR5M-xK2EC5yjKwh:b-fs8is5) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1131)):*
> ```Cairo
>         let one = One::<T>::one();
>         let two = one + one;
>         let eight = one + one + one + one + one + one + one + one;
> ```

two + two + two + two ?
___
*[`crates/utils/src/helpers.cairo` line 1134 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0MHon7fkYpTfAD2zn:-Np0MHon7fkYpTfAD2zo:b-c0iylo) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1134)):*
> ```Cairo
> 
>         // 0xFF
>         let mask = (eight * eight * two * two) - one;
> ```

Can you check if you can do
let mask = BoundedInt::<u8>::max().into()
___
*[`crates/utils/src/helpers.cairo` line 1137 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0MvjS5CWRDXENX4fA:-Np0MvjS5CWRDXENX4fB:b5vxpz0) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1137)):*
> ```Cairo
> 
>         let mut bytes: Array<u8> = Default::default();
>         let mut i: T = Zero::zero();
> ```

here we could have i: u8 as well
___
*[`crates/utils/src/helpers.cairo` line 1143 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0N-9E-t-DM7Uf7WBG:-Np0N-9E-t-DM7Uf7WBH:bi9rfxg) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1143)):*
> ```Cairo
>             }
> 
>             let val = Bitshift::<T>::shr(self, eight * (bytes_used.try_into().unwrap() - i - one));
> ```

and the casting of bytes_used and i to `T` could be done here with a simple `into()`, as we know for sure that nothing is smaller than `u8`
___
*[`crates/utils/src/helpers.cairo` line 1163 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0NAYC7iEYRitB-8We:-Np0NAYC7iEYRitB-8Wf:b-xctsrj) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1163)):*
> ```Cairo
> 
>         // 0xFF
>         let mask = (eight * eight * two * two) - one;
> ```

all same as above for be_bytes
___
*[`crates/utils/src/helpers.cairo` line 1724 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0NJZnDliXnH1ohURx:-Np0NJZo8KcmDJjkQyX0:b7sxh82) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1724)):*
> ```Cairo
>     /// Expands a Felt252Vec to a new length by appending zeroes
>     ///
>     /// This function will mutate the Felt252Vec in-place and will expand its length to the new length by appending zeroes.
> ```

Not really! We don't append zeroes. We just leverage the fact that the default value for a dict is 0.



_Code quote:_
```Cairo
    /// This function will mutate the Felt252Vec in-place and will expand its length to the new length by appending zeroes.
```
___
*[`crates/utils/src/helpers.cairo` line 1953 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0Nh7c4OJH2z0slgoR:-Np0Nh7d2gJgqjh9VG41:undefined) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1953)):*
> ```Cairo
>     /// Returns a new Felt252Vec<T> with elements starting from `idx` to `idx + len`
>     ///
>     /// This function will start slicing from `idx` and will slice `len` elements, it will firstly clone the elements and then return a new Felt252Vec<T>
> ```





_Suggestion:_
```Cairo
cloning from `idx` and will clone
```

bajpai244 (commented):
Thanks

> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-:-Np0O-_ZD38ibuRhyC-H:b-zcjvte)* status: all files reviewed, 14 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1127 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0Maef-PBTC4kl-J4X:-Np0OXAt60hKzHn7_W4g:byuzxef) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1127)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

is `into()` required? can we keep bytes_used: u8?
</blockquote></details>

we use bytes_used later on to calculate the val, which is done by doing operations on type T
___
*[`crates/utils/src/helpers.cairo` line 1131 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0MR5M-xK2EC5yjKwg:-Np0O76j8nDeJS3pngcs:btuwrzr) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1131)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

two + two + two + two ?
</blockquote></details>

haha, two* two * two 

Done
___
*[`crates/utils/src/helpers.cairo` line 1134 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0MHon7fkYpTfAD2zn:-Np0OGWe5iHpYU_NhhAD:b8nbk90) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1134)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Can you check if you can do
let mask = BoundedInt::<u8>::max().into()
</blockquote></details>

Makes sense, done
___
*[`crates/utils/src/helpers.cairo` line 1137 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0MvjS5CWRDXENX4fA:-Np0Ogc99BZPGKcuqImu:b-bbgsfs) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1137)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

here we could have i: u8 as well
</blockquote></details>

Won't solve much as  we will still need TyInto<T, u8> for bytes.append()


___
*[`crates/utils/src/helpers.cairo` line 1143 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0N-9E-t-DM7Uf7WBG:-Np0PyNR8WCy-A1rrGwf:b-xxy54z) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1143)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

and the casting of bytes_used and i to `T` could be done here with a simple `into()`, as we know for sure that nothing is smaller than `u8`
</blockquote></details>

See the above comment, even after this, we will still need TryInto<T, u8>
___
*[`crates/utils/src/helpers.cairo` line 1163 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0NAYC7iEYRitB-8We:-Np0QCZK1QYJKC2IIMnl:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1163)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

all same as above for be_bytes
</blockquote></details>

done

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: all files reviewed, 14 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1137 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0MvjS5CWRDXENX4fA:-Np0QtR77FK2P0uouSoU:ba01z5s) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1137)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

Won't solve much as  we will still need TyInto<T, u8> for bytes.append()


</blockquote></details>

I have made the refactor
___
*[`crates/utils/src/helpers.cairo` line 1143 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0N-9E-t-DM7Uf7WBG:-Np0QqiW8dfoE52mjVzk:ba01z5s) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1143)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

See the above comment, even after this, we will still need TryInto<T, u8>
</blockquote></details>

I have made the refactor

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 9 of 11 files reviewed, 14 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1684 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-NozsDhpC2GbvsH7YBsK:-Np0R8Bv76cr-GqJkP-7:b4w9x18) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1684)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Just a thought for now: this is exactly the same as `to_bytes` on an Array. We could use the trait system to implement this over a generic type T representing a Collection, that Array and Vec would implement. Will think about it later once the core concerns of this PR are addressed
</blockquote></details>

Refactored into a generic as of now, we can look into making an abstraction for this similar to as an Array later on as well, I guess for now we have been able to refactor this into a generic
___
*[`crates/utils/src/helpers.cairo` line 1702 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozub_S7l4a4iq6hzm2:-Np0RMr472eSKmQvBQDu:bb6tzbb) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1702)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I think it would be better if we had a generic Felt252VecExTrait rather than type-specific traits that will need multiple imports
</blockquote></details>

DOne
___
*[`crates/utils/src/helpers.cairo` line 1929 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-2-KQD7g0lgeK0-DJ:-Np0RdO683emY-VNTJcJ:b-7ifz29) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1929)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I do not understand how this function is different than a clone (duplicate in your case)
</blockquote></details>

This make it easier to do then do:

result = scratch_space.duplicate()

But, it makes sense to use duplicate under the hood for its implementation! 

Done
___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np0Siu17GJsqm8oe9Sc:b-2j08bm) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

If it's used as a test utils and not in the source code, it should be in a `tests/` subdirectory
</blockquote></details>

Yeah, but I would wait till next PR to confirm that is not used internally anywhere else, and if not then I will refactor it uo
___
*[`crates/utils/src/helpers.cairo` line 1724 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0NJZnDliXnH1ohURx:-Np0RS7C2LDtj1lRj2yP:b-r3ay1v) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1724)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

Not really! We don't append zeroes. We just leverage the fact that the default value for a dict is 0.
</blockquote></details>

But, for the consumer of this function does this detail matters? for them the behaviour is as if the it got filled by 0 as a default value for new indexes
___
*[`crates/utils/src/helpers.cairo` line 1953 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0Nh7c4OJH2z0slgoR:-Np0Shwf51hjzoGxtTYB:bb6tzbb) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1953)):*
> ```Cairo
>     /// Returns a new Felt252Vec<T> with elements starting from `idx` to `idx + len`
>     ///
>     /// This function will start slicing from `idx` and will slice `len` elements, it will firstly clone the elements and then return a new Felt252Vec<T>
> ```

DOne

enitrat (changes requested):
very nice to see how the use of smart generics and performance improvements have been integrated, :dancers: 

> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-:-Np1Yw2yBiU5UX6RCODi:b-mbrgar)* status: all files reviewed, 7 unresolved discussions (waiting on @bajpai244 and @ClementWalter)

___
*[`crates/utils/src/helpers.cairo` line 1702 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozub_S7l4a4iq6hzm2:-Np1XZgd8EfSfKGArT0X:b-2c4hbn) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1702)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

DOne
</blockquote></details>

This was not necessarily what I had in mind, I expressed myself badly: I meant something like 
```rust
trait Felt252VecExTrait{

  fn to_bytes(ref self: Felt252Vec<u8>) -> Span<u8>;
  fn words64_to_le_bytes(ref self: Felt252Vec<u64>) -> Span<u8>;
  fn resize...
```

but, I guess this is even better!
___
*[`crates/utils/src/helpers.cairo` line 1929 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-2-KQD7g0lgeK0-DJ:-Np1YVj1DBxjiuMVa9Og:bqbl6xv) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1929)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

This make it easier to do then do:

result = scratch_space.duplicate()

But, it makes sense to use duplicate under the hood for its implementation! 

Done
</blockquote></details>

I still don't understand :thinking: If it's a function that does nothing else than a clone I don't think it should exist, we should use clone instead.
from the example shown you could just do
```rust
    result.digits = scratch_space.duplicate()
```
___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np1YqKq7HK7vjtSri1K:bsqc73w) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
<details><summary><i>Previously, bajpai244 (Harsh Bajpai) wrote…</i></summary><blockquote>

Yeah, but I would wait till next PR to confirm that is not used internally anywhere else, and if not then I will refactor it uo
</blockquote></details>

ok, then add a TODO comment please otherwise this will be forgotten
___
*[`crates/utils/src/helpers.cairo` line 1724 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0NJZnDliXnH1ohURx:-Np1XuWk85pxfFRvSVE0:bdfjw9h) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1724)):*
For the consumer indeed it doesn't matter. But this line is for the dev, who will look into this function in 6 months and will not understand why it says that it's appending zeroes when it's not. 

> Expands a Felt252Vec to a new length by appending zeroes

This line already gives the information to the consumer: now we need to give extra info about implementation details.

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: all files reviewed, 7 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1702 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Nozub_S7l4a4iq6hzm2:-Np41TId7x49VrFaWTdU:b-q8ilnv) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1702)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

This was not necessarily what I had in mind, I expressed myself badly: I meant something like 
```rust
trait Felt252VecExTrait{

  fn to_bytes(ref self: Felt252Vec<u8>) -> Span<u8>;
  fn words64_to_le_bytes(ref self: Felt252Vec<u64>) -> Span<u8>;
  fn resize...
```

but, I guess this is even better!
</blockquote></details>

gotcha
___
*[`crates/utils/src/helpers.cairo` line 1929 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-2-KQD7g0lgeK0-DJ:-Np41WotAvrPYsSaPNW-:b-klylo9) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L1929)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

I still don't understand :thinking: If it's a function that does nothing else than a clone I don't think it should exist, we should use clone instead.
from the example shown you could just do
```rust
    result.digits = scratch_space.duplicate()
```
</blockquote></details>

For me in the core lib this check is quite important! I want to duplicate given that they are both not the same length, if in any case this isn't true then the code should panic as it is an assumption in the code base!



In rust they have a function for this called copy_from_slice, which does the same


For now, I would like to have this, as I wouldn't want to manually check that both have vecs have same length before duplicating
___
*[`crates/utils/src/helpers.cairo` line 2105 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np-5ZjO1USRrlW6KCtq:-Np42rYo-mml0kp64tL0:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/c8fc28e2ff3b9a1d729f7d076b5001325ee364d8/crates/utils/src/helpers.cairo#L2105)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

ok, then add a TODO comment please otherwise this will be forgotten
</blockquote></details>

done

bajpai244 (commented):
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: 10 of 11 files reviewed, 7 unresolved discussions (waiting on @ClementWalter and @enitrat)

___
*[`crates/utils/src/helpers.cairo` line 1724 at r6](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714#-Np0NJZnDliXnH1ohURx:-Np42tbm-MeN-fYxuzAj:bb7f2mf) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/e21d82c274420ac8086adb5c5a0c8cae4ffc354a/crates/utils/src/helpers.cairo#L1724)):*
<details><summary><i>Previously, enitrat (Mathieu) wrote…</i></summary><blockquote>

For the consumer indeed it doesn't matter. But this line is for the dev, who will look into this function in 6 months and will not understand why it says that it's appending zeroes when it's not. 

> Expands a Felt252Vec to a new length by appending zeroes

This line already gives the information to the consumer: now we need to give extra info about implementation details.
</blockquote></details>

done

enitrat (approved):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/714)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ClementWalter)

PR #734, opened 2024-02-09
fix: bits_used when value is 0

enitrat:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

- Fix `bits_used`
- add Modexp to Precompiles library
-

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/734)

### DIFF
--- a/crates/evm/src/precompiles/modexp.cairo
+++ b/crates/evm/src/precompiles/modexp.cairo
@@ -12,7 +12,7 @@ use evm::precompiles::Precompile;
 use integer::{u32_overflowing_add, BoundedInt};
 use starknet::EthAddress;
 use utils::crypto::modexp::lib::modexp;
-use utils::helpers::{U256Trait, U8SpanExTrait, U64Trait, FromBytes, BitLengthTrait};
+use utils::helpers::{U256Trait, U8SpanExTrait, U64Trait, FromBytes, BitsUsed};
 
 const HEADER_LENGTH: usize = 96;
 const MIN_GAS: u128 = 200;
@@ -62,7 +62,7 @@ impl ModExp of Precompile {
         // Used to extract ADJUSTED_EXPONENT_LENGTH.
         let exp_highp_len = cmp::min(exp_len, 32);
 
-        let input = if input.len() >= 96 {
+        let input = if input.len() >= HEADER_LENGTH {
             input.slice(HEADER_LENGTH, input.len() - HEADER_LENGTH)
         } else {
             array![].span()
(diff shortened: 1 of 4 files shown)


crates/contracts/src/precompiles.cairo
```diff
@@ -40,6 +41,8 @@ mod Precompiles {
                 0 => Result::Err(EVMError::NotImplemented),
                 1 => Result::Err(EVMError::NotImplemented),
                 2 => Sha256::exec(data),
+                3 | 4 => Result::Err(EVMError::NotImplemented),
```
bajpai244, line 44: Didn't knew u could do that! So cool!

enitrat: https://book.cairo-lang.org/ch06-02-the-match-control-flow-construct.html#multiple-patterns-with-the--operator hehe

crates/evm/src/precompiles/modexp.cairo
```diff
@@ -130,14 +130,21 @@ fn calculate_multiplication_complexity(base_length: u64, mod_length: u64) -> u25
...
+        if exp_highp == 0 {
+            0
+        } else {
+            (exp_highp.bits_used() - 1).into()
+        }
     } else {
-        let max: u64 = cmp::max(exp_highp.bit_len().into(), 1);
+        let length_part = 8 * (exp_length - 32);
+        let bits_part = if exp_highp != 0 {
+            exp_highp.bits_used() - 1
```
bajpai244, line 142: It is `bits_used` and `bit_len`

+ I would prefer to still make use of cmp::max and remove the if and else!

i.e:
`bits_part` = cmp::max(exp_highp.bit_len().into(), 1) - 1

crates/utils/src/helpers.cairo
```diff
@@ -1477,22 +1436,65 @@ impl BitLengthTraitImpl<
...
+        bits_used.into() + 8 * (bytes_used - 1).into()
     }
 
     fn count_leading_zeroes(self: T) -> u32 {
-        BitSize::<T>::bits() - self.bit_len()
+        BitSize::<T>::bits() - self.bits_used()
+    }
+}
+
+mod bits_used_internal {
```
bajpai244, line 1461: Why make this a `mod` and not a trait?

crates/evm/src/precompiles/modexp.cairo
enitrat: cmp::max makes useless range checks, we can just check if exp_highp == 0

crates/utils/src/helpers.cairo
enitrat: why would this be a trait? There's already a trait that defines the set of functions. I put it in a mod to hide it from external usage and only enable it in this module

enitrat: If I make it a trait then `bits_used` is ambiguous would be ambiguous (u8 ? generic? )
Also, since it's only meant to be used as an internal function of the BitsUsed trait, I see little value in implementing it on a type

crates/evm/src/precompiles/modexp.cairo
bajpai244: Oh! Then makes sense!

crates/utils/src/helpers.cairo
bajpai244: Makes sense!

enitrat: And generally I think *globally* we've had the wrong approach on traits, which should be a set of functionalities shared between types rather than functionalities over a single type. For example, you were very right in creating FromBytes and ToBytes traits, rather than having these functionalities inside the "trait" of each type (which in reality was only done to have the "method" syntax... anyway I'm still thinking about this)

bajpai244 (approved):
LGTM 🚀

PR #754, opened 2024-04-03
dev: CairoZero crate

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

- CairoZero-specific crate
-
-

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/754)


crates/cairo_zero_contracts/src/uninitialized_account.cairo
```diff
@@ -0,0 +1,59 @@
...
+        library_call_syscall
+    };
+    use super::{
+        IUninitializedAccount, IKakarotCoreDispatcher, IKakarotCoreDispatcherTrait,
+        INITIALIZE_SELECTOR
+    };
+
+    #[storage]
+    struct Storage {
+        Account_evm_address: EthAddress,
```
ClementWalter, line 33: are we sure that storage slots are computed the same ? maybe we should use const hard coded storage keys for these two?

crates/cairo_zero_contracts/src/uninitialized_account.cairo
```diff
@@ -0,0 +1,59 @@
...
+    #[abi(embed_v0)]
+    impl UninitializedAccountImpl of IUninitializedAccount<ContractState> {
+        fn initialize(ref self: ContractState) {
+            let kakarot = IKakarotCoreDispatcher {
+                contract_address: self.Account_kakarot_address.read()
+            };
+            let implementation_class = kakarot.get_account_contract_class_hash();
+            replace_class_syscall(implementation_class).unwrap_syscall();
+
+            let calldata = array![implementation_class.into()];
```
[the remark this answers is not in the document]
enitrat: yes so actually that's something I thought about yesterday and outlined the reasoning in https://github.com/kkrt-labs/kakarot/pull/1056/commits/2af523c124522a1ad732379e432597d6edf111a2

Ideally we would not need to store an account's own class if we had access to a get_class_hash_at syscall but that sounded like an elegant possibility

crates/cairo_zero_contracts/src/uninitialized_account.cairo
enitrat: yes, they're computed the same

Issue #755, opened 2024-04-11
feat: add RIP-7212 to the list of precompiles Cairo1Helpers module


Eikix:
### Describe the Feature Request

The people demand [RIP-7212](https://ethereum-magicians.org/t/eip-7212-precompiled-for-secp256r1-curve-support/14789)! 

What RIP7212: https://www.alchemy.com/blog/what-is-rip-7212
Why: for easier usage of TEE signer
How: use the new shiny Starknet secp256r1 syscall!

### Describe Preferred Solution

- Go through the EIP -> https://eips.ethereum.org/EIPS/eip-7212 and RIP -> https://github.com/ethereum/RIPs/blob/master/RIPS/rip-7212.md to understand the API of the precompile, the gas usage and its permanent address.
- Go through the documentation of the secp256r1 syscall in Starknet
- Implement the precompile
- Test the precompile

Difficulty level: intermediate
Estimated time: 1 day of reading, 1 day of implem, 1 day of back and forth with maintainers

Quentash:
Hello,
I could try to take care about this ticket !

PR #779, opened 2024-04-18
Feat: add RIP-7212 to the list of precompiles Cairo1Helpers module

Quentash:
Add the precompiled contracts p256verify to the list of precompiled contracts at address 0x0b following this implementation : https://github.com/ethereum/go-ethereum/pull/27540 .

I have chosen the 0x0b instead of 0x100 because of the amount of lines that the second one would have added to the precompiles.cairo and I am currently unaware of a precompiled contract at address 0x0b.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #755 

## What is the new behavior?

- p256verify is now added to the list of precompiled contract at address 0x0b

- [ ] Yes
- [X] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779)

### DIFF
--- a/crates/evm/src/call_helpers.cairo
+++ b/crates/evm/src/call_helpers.cairo
@@ -121,5 +121,5 @@ impl CallHelpersImpl of CallHelpers {
 /// Check whether an address for a call-family opcode is a precompile.
 fn is_precompile(self: EthAddress) -> bool {
     let self: felt252 = self.into();
-    return (self != 0 && self.into() < 257_u256);
+    return (self != 0 && (self.into() < 10_u256 || self.into() == 256_u256));
 }
--- a/crates/evm/src/call_helpers.cairo
+++ b/crates/evm/src/call_helpers.cairo
@@ -121,5 +121,5 @@ impl CallHelpersImpl of CallHelpers {
 /// Check whether an address for a call-family opcode is a precompile.
 fn is_precompile(self: EthAddress) -> bool {
     let self: felt252 = self.into();
-    return (self != 0 && self.into() < 10_u256);
+    return (self != 0 && self.into() < 12_u256);
 }
--- a/crates/evm/src/call_helpers.cairo
+++ b/crates/evm/src/call_helpers.cairo
@@ -121,5 +121,5 @@ impl CallHelpersImpl of CallHelpers {
 /// Check whether an address for a call-family opcode is a precompile.
 fn is_precompile(self: EthAddress) -> bool {
     let self: felt252 = self.into();
-    return (self != 0 && self.into() < 12_u256);
+    return (self != 0 && self.into() < 257_u256);
 }
--- a/crates/evm/src/precompiles.cairo
+++ b/crates/evm/src/precompiles.cairo
@@ -29,50 +29,51 @@ impl PrecompilesImpl of Precompiles {
         let precompile_address = vm.message.target.evm;
         let input = vm.message().data;
 
-        let (gas, result) = match precompile_address.address {
-            0 => {
-                // we should never reach this branch!
-                panic!("pre-compile address can't be 0")
-            },
-            1 => { EcRecover::exec(input)? },
-            2 => { Sha256::exec(input)? },
-            3 => {
-                // we should never reach this branch!
-                panic!(
-                    "pre-compile at address {} isn't implemented yet", precompile_address.address
-                )
-            },
-            4 => { Identity::exec(input)? },
-            5 => { ModExp::exec(input)? },
-            6 => {
-                // we should never reach this branch!
-                panic!(
-                    "pre-compile at address {} isn't implemented yet", precompile_address.address
-                )
-            },
-            7 => {
-                // we should never reach this branch!
-                panic!(
-                    "pre-compile at address {} isn't implemented yet", precompile_address.address
-                )
-            },
-            8 => {
-                // we should never reach this branch!
-                panic!(
-                    "pre-compile at address {} isn't implemented yet", precompile_address.address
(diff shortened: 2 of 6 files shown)


enitrat:
> I have chosen the 0x0b instead of 0x100 because of the amount of lines that the second one would have added to the precompiles.cairo and I am currently unaware of a precompiled contract at address 0x0b.

You can just add a specific if / else before the match table handling this specific case for 0x100. This should not be an issue when deciding a precompile address. Given that it seems a consensus was found on using 0x100 - 0x1ff for rollup precompiles, let's go with 0x100

enitrat (changes requested):
> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @ClementWalter and @Quentash)

___
*[`crates/evm/src/precompiles.cairo` line 72 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779#-Nvlby87Dlj5WKxUVsz7:-Nvlby87Dlj5WKxUVsz8:b-11knjb) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/16fc6f633b97fa82a57b192eed1ee605c33f3695/crates/evm/src/precompiles.cairo#L72)):*
> ```Cairo
>                 )
>             },
>             11 => { P256Verify::exec(input)? },
> ```

see main comment - we can still move use it as 0x100
___
*[`crates/evm/src/precompiles/p256verify.cairo` line 20 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779#-NvlcnFU6uyzeyUkv6j5:-NvlcnFU6uyzeyUkv6j6:b-ipv4a1) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/16fc6f633b97fa82a57b192eed1ee605c33f3695/crates/evm/src/precompiles/p256verify.cairo#L20)):*
> ```Cairo
>     fn exec(input: Span<u8>) -> Result<(u128, Span<u8>), EVMError> {
>         let gas: u128 = P256VERIFY_PRECOMPILE_GAS_COST;
> ```

missing input size check. if size is 160 then you can assume the precompile input is correct and simply `message_hash.from_be_bytes().unwrap()` (same for all conversions)

globally I think you can remove all the manual error handling as long as size is ok. if size !=160 return 0
___
*[`crates/evm/src/tests/test_precompiles/test_p256verify.cairo` line 16 at r1](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779#-NvlddS95FviOt68DNVI:-NvlddS95FviOt68DNVJ:bnll29b) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/16fc6f633b97fa82a57b192eed1ee605c33f3695/crates/evm/src/tests/test_precompiles/test_p256verify.cairo#L16)):*
> ```Cairo
> #[test]
> fn test_p256verify_precompile() {
>     let (_, _) = setup_contracts_for_testing();
> ```

do you need to setup a contract? Can this not be a unit test?

Quentash:
- Chose 0x100 as precompiled address
- Added length check ( nice catch )
- Added tests concerning length check
- Removed setup_contracts for unit tests

enitrat (changes requested):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779)* status: all files reviewed, 3 unresolved discussions (waiting on @ClementWalter and @Quentash)

___
*[`crates/evm/src/call_helpers.cairo` line 124 at r2](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779#-Nvls41sD-vK91EmBmvv:-Nvls41sD-vK91EmBmvw:b-rnf1pp) ([raw file](https://github.com/kkrt-labs/kakarot-ssj/blob/759c8f76dd4f239e02c2242774cd5028922fe284/crates/evm/src/call_helpers.cairo#L124)):*
> ```Cairo
> fn is_precompile(self: EthAddress) -> bool {
>     let self: felt252 = self.into();
>     return (self != 0 && self.into() < 257_u256);
> ```

only [0, 9] and 0x100 are valid addresses, this needs to be modified accordingly

Quentash:
Corrected precompiled address range

enitrat:
some tests are failing, can you verify is `256` or `0x100` is an address used in tests?

Quentash:
Alright, my bad, I missed those since I struggle running every tests and didn't notice it on github.
But indeed, some tests were deploying on the 0x100 address which created conflicts. I changed them to deploy their test contract on 0x101.

enitrat:
> Alright, my bad, I missed those since I struggle running every tests and didn't notice it on github.
> But indeed, some tests were deploying on the 0x100 address which created conflicts. I changed them to deploy their test contract on 0x101.

0x101 is still too "close" we should ideally have a random value like 0xabfa740ccd

Quentash:
You're right ! 0xabfa740ccd it is then !

enitrat (approved):
:lgtm:

> Reviewed 1 of 1 files at r3, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/779#-:-Nvplc8805D4NDB_IRhS:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @ClementWalter and @Quentash)

PR #790, opened 2024-04-26
fix: rip-7212 implementation incorrect return value

ulerdogan:
## Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes

The current implementation doesn't meet with actual RIP-7212 behavior for the return values. It returns unwanted 0 value for some cases.

Resolves: #

## What is the new behavior?

The changes apply true behavior for the RIP-7212.

```
- **Output data:**
    - If the signature verification process succeeds, it returns 1 in 32 bytes format.
    - If the signature verification process fails, it does not return any output data.
```

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/790)


enitrat (approved):
lgtm ❤️

Issue #808, opened 2024-07-26
dev: use `Deref` to trivially access `T`'s members from `Box<T>`


enitrat:
Current architecture will use some more `Box<T>` types to pass big structs to functions and reduce the step usage. By implementing Deref<Box<T>> and DerefMut<Box<T>> with T being the internal target type we could abstract the `var.unbox().member` accesses and simply do `var.member`, where `var: Box<T>`.

Also use for all Nullables

remybar:
Can I have a look at this task ?

PR #817, opened 2024-08-08
implement DeRefMut for Box<T>

remybar:
See issue https://github.com/kkrt-labs/kakarot-ssj/issues/808.

Note that DerefBox is already implemented in corelib::box (https://github.com/starkware-libs/cairo/blob/dd86a0eb2707d0067939e5051e04b25d93f515f3/corelib/src/box.cairo#L56).

But I left the DerefBox unit test to be sure that it still works in the next Cairo versions.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/817)

### DIFF
--- a/crates/utils/src/traits.cairo
+++ b/crates/utils/src/traits.cairo
@@ -127,6 +127,15 @@ pub impl U8IntoEthAddress of Into<u8, EthAddress> {
     }
 }
 
+// Note: DerefBox is already implemented in corelib::box
+
+impl DerefMutBox<T, +Copy<T>, +Drop<T>> of core::ops::DerefMut<Box<T>> {
+    type Target = T;
+    fn deref_mut(ref self: Box<T>) -> T {
+        self.unbox()
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use core::starknet::{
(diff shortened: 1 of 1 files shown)


enitrat:
> Note that DerefBox is already implemented in corelib::box 

Does that mean that I can access all members of `T` from a `Box<T>` natively? If so I dont think there's anything to do here, as deref_mut is to use deref only on mutable variables

remybar:
> > Note that DerefBox is already implemented in corelib::box
> 
> Does that mean that I can access all members of `T` from a `Box<T>` natively? If so I dont think there's anything to do here, as deref_mut is to use deref only on mutable variables

Yes, you can do things like that directly: https://github.com/kkrt-labs/kakarot-ssj/blob/6268cd56792199c5c11cb85bcd1af1901e8a7090/crates/utils/src/traits.cairo#L174

remybar:
> If so I dont think there's anything to do here, as deref_mut is to use deref only on mutable variables

Yes, I'm not sure there is a use case to deref mutable boxes 😅 
I close this PR so you can close the associated issue.

Issue #806, opened 2024-07-24
dev: use const fixed-size arrays for lookup tables


enitrat:

remybar:
I'm looking at this one if it's ok for you ;)

PR #822, opened 2024-08-08
Feat: const fixed size arrays

remybar:
See issue https://github.com/kkrt-labs/kakarot-ssj/issues/806.

The first commit uses const fixed-size arrays in the code while the second one uses const fixed-size arrays in tests.

## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

- [ ] Yes
- [X] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/822)


crates/contracts/src/account_contract.cairo
```diff
@@ -81,6 +81,41 @@ pub mod AccountContract {
 
     const VERSION: u32 = 000_001_000;
 
+    const KAKAROT_VALIDATION_FAILED: [
```
enitrat, line 84: can you put this inside the `contracts::errors` module please

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -649,7 +653,7 @@ mod tests {
 
     #[test]
     fn test_calldatacopy_basic() {
-        test_calldatacopy(32, 0, 3, array![4, 5, 6].span());
+        test_calldatacopy(32, 0, 3, ARRAY_3.span());
```
enitrat, line 656: can this not be inlined as 

```suggestion
        test_calldatacopy(32, 0, 3, [4,5,6].span());
```
?

If yes, do this for all the tests. If not, revert to the previous implementation, as we don't really about perf for tests.

enitrat (commented):
can you also adapt `pow2` and `pow256_rev` ?

remybar:
> can you also adapt `pow2` and `pow256_rev` ?

Yes, done 👍

enitrat (approved):

Issue #841, opened 2024-08-21
feat: use scarb macro to generate const powers


enitrat:
### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

juankohler:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
I have over 7 years of experience in backend development, primarily using Go. While my expertise is in building robust backend systems, I’m currently learning Cairo in Dojo Coding and am eager to apply my skills to contribute effectively to projects using this language.

enitrat:
@ClementWalter the Language Server has some issues with procedural macros (they're undetected so considered as errors) - as such I don't want to merge it yet, I reported the issue and proposed them to simply mute diagnostics for macros.

maxslimb:
@enitrat is this issue completed ?

PR #865, opened 2024-08-23
feat: use scarb macro to generate const powers

juankohler:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #841

## What is the new behavior?

- Uses pow! macro instead of hardcoded constants.
-
-

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/865)


juankohler:
hi @enitrat 
I am importing the alexandria_macros package in crates/utils/Scarb.toml, but when I try to use it in utils/src/helpers.cairo, it gives me an 'Identifier not found.' error on the line 'use alexandria_macros::pow'. Can you help me with this please?

juankohler:
Hi @enitrat, I have a problem. When I took the issue, I interpreted that I needed to export the Alexandria package, but then I realized that it was in Rust. When I looked into how to calculate the power of 2 at compile time in Cairo, I couldn't find any solution. The current utils only calculate at runtime up to 128. The only way I can think of is to hardcode the constants up to that number and create a function that returns them depending on the argument passed, but this doesn't seem like an optimal solution. Could you help me with this?

enitrat:
I don't understand the problem. Can't you use the macro?

enitrat:
merge blocked until snfoundry allows use of scarb macros

Issue #853, opened 2024-08-22
feat: implement `execute_starknet_call` in account contract


enitrat:
### Describe the Feature Request
Add an entrypoint that allows Kakarot to call any contract on behalf of an account.

### Additional Context

Reference: https://github.com/kkrt-labs/kakarot/blob/65e333a0d3475e81b0a6367b0f00561c763da0a0/src/kakarot/accounts/account_contract.cairo#L310-L330

PR #871, opened 2024-08-26
feat: execute_starknet_call

obatirou:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #853 

## What is the new behavior?

- add execute_starknet_call

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/871)


crates/contracts/src/account_contract.cairo
```diff
@@ -271,6 +272,23 @@ pub mod AccountContract {
...
+            self.ownable.assert_only_owner();
+            let kakarot_address = self.ownable.owner();
+            if call.to == kakarot_address && call.selector != selector!("get_starknet_address") {
+                return (false, KAKAROT_REENTRANCY.span());
+            }
+            let response = call_contract_syscall(call.to, call.selector, call.calldata);
+            if response.is_ok() {
+                return (true, response.unwrap().into());
+            }
+            return (true, response.unwrap_err().into());
```
enitrat, line 290: should success be false here?

```suggestion
            return (false, response.unwrap_err().into());
```

crates/contracts/src/account_contract.cairo
```diff
@@ -271,6 +272,23 @@ pub mod AccountContract {
             self.ownable.assert_only_owner();
             self.Account_nonce.write(nonce);
         }
+
+        // @notice Used to preserve caller in Cairo Precompiles
+        // @dev Reentrency check is done for Kakarot contract, only get_starknet_address is allowed
+        //      for Solidity contracts to be able to get the corresponding Starknet address in their
+        //      calldata.
```
enitrat, line 279: we prefer using rustdoc format, so no `@notice` nor `@dev` tags, and three slashes

```
/// Used to preserve caller in Cairo Precompiles.
///
```

crates/contracts/tests/test_contract_account.cairo
```diff
@@ -63,3 +70,62 @@ fn test_ca_storage() {
 
     assert(storage == expected_storage, 'wrong contract storage');
 }
+
+#[test]
```
enitrat, line 74: i am going to wait a bit before merging this as i am currently migrating to sn-foundry and this will change

crates/contracts/src/account_contract.cairo
obatirou: Feature parity with kkrt0 is no check on the return of the call https://github.com/kkrt-labs/kakarot/blob/ebcdf0b0167cb5b5d1cb2934761809cfc1d63902/src/kakarot/accounts/account_contract.cairo#L339 and return what is sent back

enitrat: In kkrt0 we cannot catch the result of the syscall, which is why we cannot return FALSE.
In practice, if the syscall fails, the transaction will revert. But later in the future, we will be able to catch them, and I think it is better to have this entrypoint return false in this case.

tldr; the err case will never be triggered until call errors can be caught, as it will make the tx fail

enitrat (approved):

Issue #856, opened 2024-08-22
feat: create EVM Precompile wrapper around ec_mul and ec_add


enitrat:
### Describe the Feature Request
EVM Precompiles take bytes as input and output bytes. Wrap ec_mul and ec_add to match the evm specs.

Depends on https://github.com/kkrt-labs/kakarot-ssj/pull/855
### Additional Context

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

PR #880, opened 2024-08-28
feat: ecAdd / ecMul

obatirou:
## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #856

## What is the new behavior?

Wrap ecAdd / ecMul

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/880)

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
 scarb 2.7.1
-starknet-foundry 0.28.0
+starknet-foundry 0.30.0
--- a/Scarb.lock
+++ b/Scarb.lock
@@ -46,12 +46,12 @@ dependencies = [
 [[package]]
 name = "snforge_scarb_plugin"
 version = "0.1.0"
-source = "git+https://github.com/foundry-rs/starknet-foundry.git?tag=v0.28.0#4dfe39d96690ed6b3d56971512700de3f58288ea"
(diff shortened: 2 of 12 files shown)


obatirou:
blocked by https://github.com/foundry-rs/starknet-foundry/issues/2397

feltroidprime:
Some note : if ec_add or ec_mul returns None, it shouldn't be interpreted as (0,0) which a valid way of encoding the point at inifinity. It should be a precompile error

obatirou:
> Some note : if ec_add or ec_mul returns None, it shouldn't be interpreted as (0,0) which is a valid way of encoding the point at inifinity. It should be a precompile error
> 
> See https://www.evm.codes/precompiled ` If the input is not valid, all gas provided is consumed and there is no return data, indicating a precompile contract error. Also, if not enough gas was given, there is no return data.`
> 
> The Option::None case is reflecting the input is not valid .

Thank you for spotting this !
Will make sure to take this into account when working on it again.
This draft PR is not ready as it does not compile due to the blocking issue.

enitrat:
@obatirou https://github.com/foundry-rs/starknet-foundry/releases/tag/v0.30.0

Issue #923, opened 2024-09-11
feat: add `eth_chain_id` entrypoint


enitrat:
- Add an `eth_chain_id` entrypoint to the `KakarotCore` contract
- This entrypoint will return the `chain_id` field of the `tx_info`, modulo `2**53`
- Replace all references to tx_info.chain_id by this value returned from the Kakarot Contract

Closes #920

bitfalt:
hii! Is this issue open for contributors? If so I'd love to tackle this issue. I assume I would need to use `get_tx_info().unbox()` and then return the `chain_id` field after applying modulo `2**53`, then refactor the code with the value returned from it.

enitrat:
Hi! Yes, you can do it.

To help you, here's an existing implementation in old cairo.

https://github.com/kkrt-labs/kakarot/blob/7a83ddf6cfb7a69d995245291fe11c1d2444e408/src/kakarot/library.cairo#L124-L131

You could generally just find and replace all occurences of use of this tx_info.chain_id by a call to this function

bitfalt:
Okay! Thank you so much for providing a link to the existing implementation, will start ASAP!

PR #932, opened 2024-09-12
feat: add eth_chain_id entrypoint

bitfalt:
Add `eth_chain_id` entrypoint to the `KakarotCore` contract. Add a new constant to compute 2**53.  Draft PR since replacing of all references is still needed, also want to ask feedback regarding implementation (correct types) to @enitrat.

## Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Resolves: #923 

## What is the new behavior?

Add an entrypoint to the `KakarotCore` contract get the `eth_chain_id` from the transaction

- Add a new constant in `utils/src/constants.cairo` for 2**53
- Add `eth_chain_id` entrypoint

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/932)


bitfalt:
I've checked the codebase and found 3 places where we can call `kakarot.eth_chain_id()`. I have some questions regarding how to call the function in some places, and would appreciate any guidance or feedback towards my solution!

https://github.com/kkrt-labs/kakarot-ssj/blob/eeac8bb3f92fde514e46d9ab9cb30814bd83a761/crates/evm/src/backend/starknet_backend.cairo#L82

I have question regarding the code above, since it is doing `chain_id: get_tx_info().unbox().chaind_id.try_into().unwrap()` which is similar to what the implementation of the function I did. 
However, I'd like to know if it would be viable to call `eth_chain_id`. If the refactor is necessary here, I'd like to know how to call the function, since I'm used to work with dispatchers, but I don't see any address and the only reference to `KakarotCore` I see is `kakarot_state`. 

https://github.com/kkrt-labs/kakarot-ssj/blob/eeac8bb3f92fde514e46d9ab9cb30814bd83a761/crates/contracts/src/account_contract.cairo#L243-L259

The code above is easy to refactor, my solution would be move the line 259 (`let kakarot = IKakarotCoreDispatcher { contract_address: self.ownable.owner() };`) to be above line 243 (`let chain_id: u64 = tx_info.chain_id.try_into().unwrap() % POW_2_32.try_into().unwrap();`). Then just call the implemented function to set the `chain_id`. In this way: `let chain_id: u64 = kakarot.eth_chain_id()`.


https://github.com/kkrt-labs/kakarot-ssj/blob/eeac8bb3f92fde514e46d9ab9cb30814bd83a761/crates/evm/src/backend/validation.cairo#L19-L24

Regarding the code above, I have the same question as the first one I made. I'm thinking that we could call `eth_chain_id` function in the `KakarotCore` contract. However, I'm facing the same issue were this is no address, but in this case we have the parameter `kakarot_state`.  However, I still don't really get how to call the function `eth_chain_id` implemented in `KakarotCore` without the dispatcher. 

Any guidance toward the questions above is appreciated!

enitrat:
regarding 1. yes you should call eth_chain_id

for 1. and 3. the `unsafe_new_contract_state` function gives you the `ContractState` object of the contract you call in on - in this case, `KakarotCore::ContractState`. This means that any functions defined with `self: ContractState` in the kakarot contract can be called through this object. Make sure you correctly import the `IKakarotCore` trait, in that defines functions callable on the `KakarotCore::ContractState` type.

example:

```
/// Populate an Environment with Starknet syscalls.
pub fn get_env(origin: EthAddress, gas_price: u128) -> Environment {
    let kakarot_state = KakarotCore::unsafe_new_contract_state();
    let kakarot_storage = kakarot_state.snapshot_deref().storage();
    // let chain_id = kakarot_state.eth_chain_id();
    let block_info = get_block_info().unbox();

    // tx.gas_price and env.gas_price have the same values here
    // - this is not always true in EVM transactions
    Environment {
        origin: origin,
        gas_price,
        chain_id: get_tx_info().unbox().chain_id.try_into().unwrap(),
        prevrandao: kakarot_storage.Kakarot_prev_randao.read(),
        block_number: block_info.block_number,
        block_gas_limit: constants::BLOCK_GAS_LIMIT,
        block_timestamp: block_info.block_timestamp,
        coinbase: kakarot_storage.Kakarot_coinbase.read(),
        base_fee: kakarot_storage.Kakarot_base_fee.read(),
        state: Default::default(),
    }
}
```

enitrat:
@bitfalt can you move the entrypoint to the `EthRPC` trait to sync with changes made in https://github.com/kkrt-labs/kakarot-ssj/pull/937 and rebase on main?

bitfalt:
@enitrat sure! I'll modify the code accordingly.

crates/contracts/src/kakarot_core/eth_rpc.cairo
```diff
@@ -130,7 +131,9 @@ pub impl EthRPC<
     }
 
     fn eth_chain_id(self: @TContractState) -> u64 {
-        panic!("unimplemented")
+        let tx_info = get_tx_info().unbox();
```
enitrat, line 134: can you add a unit test?

enitrat (commented):
lgtm, just add a unit test in the same file :)

bitfalt:
okayy, wouldn't it be better to have the test in `crates/contracts/tests/test_kakarot_core.cairo` along with all the other tests?

enitrat:
yes, in the same module, under a `#[cfg(test)]` module, as we're unit testing

crates/contracts/src/kakarot_core/eth_rpc.cairo
bitfalt: done!

crates/contracts/src/kakarot_core/eth_rpc.cairo
```diff
@@ -210,3 +213,27 @@ fn is_view(self: @KakarotCore::ContractState) -> bool {
...
+
+        // Set up environment
+        start_cheat_chain_id_global(1);
+
+        kakarot_state
+    }
+
+
+    #[test]
+    fn test_eth_chain_id() {
```
enitrat, line 235: Your test correctly checks that calling `eth_chain_id` returns the chain_id in the environment, but doesn't check whether that value is taken mod 2**53.

Also, in this case, the need for a `set_up` is not required, as you will define set the chain_id in a case-by-case basis

A good practice is also to have more descriptive test names, e.g:

```
    #[test]
    fn test_eth_chain_id_returns_input_when_less_than_pow_2_53() {
        let kakarot_state = set_up();
        let chain_id: u64 = POW_2_53 - 1;
        start_cheat_chain_id_global(chain_id);
        assert_eq!(kakarot_state.eth_chain_id(), chain_id, "Should return original chain ID when below 2^53");
        tear_down();
    }

    #[test]
    fn test_eth_chain_id_returns_modulo_when_greater_than_or_equal_to_pow_2_53() {
        let kakarot_state = set_up();
        let chain_id: u64 = POW_2_53;
        start_cheat_chain_id_global(chain_id);
        assert_eq!(kakarot_state.eth_chain_id(), 0, "Should return 0 when chain ID is 2^53");

        let chain_id: u64 = POW_2_53 + 42;
        start_cheat_chain_id_global(chain_id);
        assert_eq!(kakarot_state.eth_chain_id(), 42, "Should return correct value after modulo operation");
        tear_down();
    }
 ```

bitfalt: I've addressed what you said, didn't consider adding more cases. Thank you for your feedback and the advice to give more significant names to tests!

enitrat:
@bitfalt there is a merge conflict; can you rebase?

bitfalt:
Just rebased everything, I think I didn't mess up the rebase this time. I'm fairly new to doing rebase instead of merge, so it is a bit confusing to me haha. Let me know if there is anything else!

enitrat (approved):

PR #948, opened 2024-09-16
fix: returndata buffer emptied on stop

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

- Fixes an issue where the returndata buffer was not correctly emptied on stop.
- Fixes an issue where the returndata was copied in memory for exceptional halts.
- Adapted the CALL opcodes tests accordingly - no longer checking returndata, but rather checking the context in which an SSTORE was executed...

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/948)

### DIFF
--- a/crates/evm/src/instructions/stop_and_arithmetic_operations.cairo
+++ b/crates/evm/src/instructions/stop_and_arithmetic_operations.cairo
@@ -14,9 +14,13 @@ pub impl StopAndArithmeticOperations of StopAndArithmeticOperationsTrait {
     /// 0x00 - STOP
     /// Halts the execution of the current program.
     /// # Specification: https://www.evm.codes/#00?fork=shanghai
-    fn exec_stop(ref self: VM) -> Result<(), EVMError> {
+    fn exec_stop(ref self: VM) {
+        // return_data stored the return_data for the last executed sub context
+        // see CALLs opcodes. When we run the STOP opcode, we stop the current
+        // execution context with *no* return data (unlike RETURN and REVERT).
+        // hence we just clear the return_data and stop.
+        self.return_data = [].span();
         self.stop();
-        Result::Ok(())
     }
 
     /// 0x01 - ADD
@@ -268,15 +272,16 @@ mod tests {
 
 
     #[test]
-    fn test_exec_stop() {
+    fn test_exec_stop_should_stop_and_empty_return_data() {
         // Given
         let mut vm = VMBuilderTrait::new_with_presets().build();
-
+        vm.return_data = [1, 2, 3].span();
         // When
-        vm.exec_stop().expect('exec_stop failed');
+        vm.exec_stop();
 
         // Then
-        assert(!vm.is_running(), 'ctx not stopped');
+        assert!(!vm.is_running());
+        assert_eq!(vm.return_data, [].span());
     }
 
     #[test]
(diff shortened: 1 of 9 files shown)


crates/evm/src/instructions/stop_and_arithmetic_operations.cairo
```diff
@@ -14,9 +14,13 @@ pub impl StopAndArithmeticOperations of StopAndArithmeticOperationsTrait {
     /// 0x00 - STOP
     /// Halts the execution of the current program.
     /// # Specification: https://www.evm.codes/#00?fork=shanghai
-    fn exec_stop(ref self: VM) -> Result<(), EVMError> {
+    fn exec_stop(ref self: VM) {
+        // return_data stored the return_data for the last executed sub context
+        // see CALLs opcodes. When we run the STOP opcode, we stop the current
+        // execution context with *no* return data (unlike RETURN and REVERT).
+        // hence we just clear the return_data and stop.
```
obatirou, line 21:
```suggestion
        // return_data store the return_data for the last executed sub context
        // see CALLs opcodes. When it runs the STOP opcode, it stops the current
        // execution context with *no* return data (unlike RETURN and REVERT).
        // hence it just clear the return_data and stop.
```
We should avoid using the "we" in comments

crates/evm/src/instructions/system_operations.cairo
```diff
@@ -587,14 +534,16 @@ mod tests {
...
-        assert(vm.return_data().is_empty(), 'Wrong return_data len');
-        assert(!vm.is_running(), 'vm should be stopped');
+        assert!(!vm.is_error());
+        assert!(!vm.is_running());
+        let storage_val = vm.env.state.read_state(contract_account.address.evm, 0x42);
+        assert_eq!(storage_val, 0x42);
     }
 
     #[test]
     fn test_exec_staticcall() {
```
obatirou, line 544: This is becoming a failing case right ?
It should be renamed and a new passing test should be added
```suggestion
    fn test_should_fail_exec_staticcall() {
```

crates/evm/src/interpreter.cairo
```diff
@@ -372,7 +377,7 @@ pub impl EVMImpl of EVMTrait {
         // Call the appropriate function based on the opcode.
         if opcode == 0x00 {
             // STOP
-            return self.exec_stop();
+            return Result::Ok(self.exec_stop());
```
obatirou, line 380: Why breaking the pattern and remove the` Result<(), EVMError>` from `exec_stop` ?
It seems be here only to be able to do 
```
if pc >= bytecode.len() {
    vm.exec_stop();
}
```
It is like executing stop but seems a little off to exec an opcode for a special case
imo it should be a 
```
// If PC is out of bounds, empty the returndata and stop the VM
if pc >= bytecode.len() {
     self.return_data = [].span();
     self.stop();
 }
```

What are your thoughts on this ?

enitrat: So I had a bit of trouble debugging this with REVM & Cast because what happend was that

bytecode_len = 21
pc_len = 17, opcode = push25
-> next pc = 43
-> It showed the execution of the STOP opcode

I later realized that no, it was not executing the STOP opcode - but rather, the behavior when PC goes out of bound is exactly the same as executing STOP. Thus the idea was to simply call `exec_stop` here.

And regarding the Result<T,E>, here, exec_stop cannot fail, thus the returned type should not be wrapped in an error

obatirou: I see, thanks for the explanation

obatirou (approved):

PR #950, opened 2024-09-16
fix: saturate memory offset taken from stack

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

- In some cases, it was possible that a memory offset taken from the stack was absurdely high, but combined with a null size. In this case, since no memory was modified, no gas was charged for that, and the memory was not resized
- This ensures that, for these high values, the program doesn't fail with a type execution error - rather, we saturate at the max_usize value, s.t. if size != 0, the memory expansion will create a failure because of a too high cost - and otherwise, the tx will pass.
-

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/950)


enitrat:
Failing test will be solved upon rebasing on #949

obatirou (approved):

PR #955, opened 2024-09-17
fix: increment pc post-opcode execution

enitrat:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes

Resolves: #

## What is the new behavior?

- Fixes an issue where PC should be incremented AFTER opcode execution, not before
- (skip increment for JUMP family opcodes)
-

- [ ] Yes
- [ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/955)

### DIFF
--- a/crates/evm/src/instructions/memory_operations.cairo
+++ b/crates/evm/src/instructions/memory_operations.cairo
@@ -713,10 +713,8 @@ mod tests {
 
         // Then
         let pc = vm.pc();
-        // ideally we should assert that it incremented, but incrementing is done by
-        // `decode_and_execute`
-        // so we can assume that will be done
-        assert(pc == old_pc, 'PC should be same');
+        // If the jump is not taken, the PC should be incremented by 1
+        assert_eq!(pc, old_pc + 1);
     }
 
     #[test]
@@ -758,10 +756,8 @@ mod tests {
 
         // Then
         let pc = vm.pc();
-        // ideally we should assert that it incremented, but incrementing is done by
-        // `decode_and_execute`
-        // so we can assume that will be done
-        assert(pc == old_pc, 'PC should be same');
+        // If the jump is not taken, the PC should be incremented by 1
+        assert_eq!(pc, old_pc + 1);
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


crates/evm/src/interpreter.cairo
```diff
@@ -334,11 +334,14 @@ pub impl EVMImpl of EVMTrait {
...
         let opcode: u8 = *bytecode.at(pc);
-        // Increment pc
-        vm.set_pc(pc + 1);
 
         match Self::execute_opcode(ref vm, opcode) {
             Result::Ok(_) => {
+                if opcode != 0x56 && opcode != 0x57 {
+                    // Increment pc if not a JUMP family opcode
+                    vm.set_pc(vm.pc() + 1);
+                }
```
obatirou, line 343: If the jump is taken, it will land on a jumpdest but PC will not be incremented.
Why not always increment post-opcode execution ?

enitrat: JUMPDEST is an opcode that needs to be executed as well; it takes 1 gas but does nothing.

crates/evm/src/instructions/memory_operations.cairo
```diff
@@ -203,6 +203,10 @@ pub impl MemoryOperation of MemoryOperationTrait {
             // remove both `value` and `b`
             self.stack.pop()?;
             self.stack.pop()?;
+
+            // Return with a PC incremented by one - as JUMP and JUMPi increments
+            // are skipped in the main `execute_code` loop
+            self.set_pc(self.pc() + 1);
```
obatirou, line 209: Comments need update but cannot make suggestion on non modified code
```
// Peek the value so we don't need to push it back again in case we want to call `exec_jump`
```
Should be `Peek b ...`

`// remove both value and b`

Should be 
`// remove both counter and b`


Also 
```
// counter would have been already popped by `exec_jump`
// so we just remove `b`
```
no sure about this comment: `jump` is executed not `exec_jump` hence 
From my understanding index == counter 
Why no pop the 2 before `jump` ?

crates/evm/src/interpreter.cairo
obatirou: right, missed the function jumpdest

crates/evm/src/instructions/memory_operations.cairo
enitrat: I think I can just do 2 pops yeah

obatirou (approved):

Issue #942, opened 2024-09-13
feat: implement `eth_get_transaction_count`


enitrat:
https://github.com/kkrt-labs/kakarot-ssj/blob/b1f7f8c13be38edc081f0738c188561240c098b4/crates/contracts/src/kakarot_core/eth_rpc.cairo#L128-L130

Given an evm address, this will return the transaction count (none) of this address.

Use it where required in the codebase, for example:

https://github.com/kkrt-labs/kakarot-ssj/blob/b1f7f8c13be38edc081f0738c188561240c098b4/crates/evm/src/backend/validation.cairo#L35-L38

reference:

https://github.com/kkrt-labs/kakarot/blob/main/src/kakarot/eth_rpc.cairo#L43-L50

Gerson2102:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
---

### Hey! I'm Gerson. 👋

Member of **Dojo Coding**. I have contributed before to other Kakarot projects like:

- `kakarot-rpc`
- `ef-tests`

Can I work on this issue?

Check my OnlyDust profile, I've contributed to many projects: [Profile](https://app.onlydust.com/u/Gerson2102)



### How I plan on tackling this issue
---

### Plan to Solve the Issue:

1. **Study the codebase** to understand the current structure and functionality.
2. **Understand the problem** that needs to be solved by thoroughly reading the issue details.
3. **Leverage all provided resources** within the issue to get a better grasp of the task at hand.
4. If I encounter roadblocks, I will **ask questions ASAP** in the Telegram group to move forward effectively.
5. **I will open a PR** and wait for ur feedback

---

mubarak23:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
I am a Experience Cairo smart contract developer with experience working on projects such as Just Art Peace, Dojo, Kart, TBA, and Shinigami. Before transitioning to Cairo development, I was a backend developer specializing in Rust.



My recent work with cairo starknet
- https://github.com/keep-starknet-strange/raito/pull/128
- https://github.com/keep-starknet-strange/raito/pull/143
- https://github.com/horuslabsio/TBA/pull/45
- https://github.com/horuslabsio/karst-core/pull/60
- https://github.com/keep-starknet-strange/art-peace/pull/121
[...]

PR #983, opened 2024-09-26
Implementation of eth_get_transaction_count function

Gerson2102:
This PR include changes on the following file `crates/contracts/src/kakarot_core/eth_rpc.cairo`. I'm doing the implementation of the eth_get_transaction_count function. The issue is not done yet.

## Pull Request type
Normal

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #942 

- [ ] Yes
- [X] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/983)

### DIFF
--- a/crates/contracts/src/kakarot_core/eth_rpc.cairo
+++ b/crates/contracts/src/kakarot_core/eth_rpc.cairo
@@ -132,7 +132,11 @@ pub impl EthRPC<
     }
 
     fn eth_get_transaction_count(self: @TContractState, address: EthAddress) -> u64 {
-        panic!("unimplemented")
+        let kakarot_state = KakarotState::get_state();
+        let starknet_address = kakarot_state.compute_starknet_address(address);
+        let account = IAccountDispatcher { contract_address: starknet_address };
+        let nonce = account.get_nonce();
+        nonce
     }
 
     fn eth_chain_id(self: @TContractState) -> u64 {
--- a/crates/contracts/src/kakarot_core/eth_rpc.cairo
+++ b/crates/contracts/src/kakarot_core/eth_rpc.cairo
@@ -133,7 +133,7 @@ pub impl EthRPC<
 
     fn eth_get_transaction_count(self: @TContractState, address: EthAddress) -> u64 {
         let kakarot_state = KakarotState::get_state();
-        let starknet_address = kakarot_state.compute_starknet_address(address);
+        let starknet_address = kakarot_state.get_starknet_address(address);
         let account = IAccountDispatcher { contract_address: starknet_address };
         let nonce = account.get_nonce();
         nonce
--- a/crates/contracts/src/kakarot_core/eth_rpc.cairo
+++ b/crates/contracts/src/kakarot_core/eth_rpc.cairo
@@ -232,6 +232,7 @@ mod tests {
     use evm::test_utils::{sequencer_evm_address, evm_address};
     use snforge_std::{start_cheat_chain_id_global, stop_cheat_chain_id_global};
     use utils::constants::POW_2_53;
+    use evm::test_utils::evm_address;
 
     fn set_up() -> KakarotCore::ContractState {
         // Define the kakarot state to access contract functions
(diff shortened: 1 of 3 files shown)


Gerson2102:
I was checking in the codebase for more places where we can now use the `eth_get_transaction_count` but I didn't find any to be honest. Let me know if there is any that is missing. @enitrat

crates/contracts/src/kakarot_core/eth_rpc.cairo
```diff
@@ -127,7 +127,11 @@ pub impl EthRPC<
     }
 
     fn eth_get_transaction_count(self: @TContractState, address: EthAddress) -> u64 {
-        panic!("unimplemented")
+        let kakarot_state = KakarotState::get_state();
+        let starknet_address = kakarot_state.compute_starknet_address(address);
```
enitrat, line 131: use `get_starknet_address` instead (the other one will be deprecated)

enitrat: this is likely the reason why the test fails, no worries

crates/evm/src/backend/validation.cairo
```diff
@@ -32,7 +32,9 @@ pub fn validate_eth_tx(kakarot_state: @KakarotCore::ContractState, tx: Transacti
     // Validate nonce
     let starknet_caller_address = get_caller_address();
     let account = IAccountDispatcher { contract_address: starknet_caller_address };
-    assert(account.get_nonce() == tx.nonce(), 'Invalid nonce');
+    let eth_address = account.get_evm_address();
+    let nonce = kakarot_state.eth_get_transaction_count(eth_address);
+    assert(nonce == tx.nonce(), 'Invalid nonce');
```
enitrat, line 37: actually, after giving it a second though, let's remove this change.
The tests here fails because we don't mock the call to `get_evm_address`, on an address that is not deployed.

But in the end, since we already have the account instance, using `eth_get_transaction_count` is just adding extra steps for nothing, so we could keep the original.

`eth_get_transaction_count` would be called by external services like the RPC to query the nonce of an account, for example.

```suggestion
    assert(account.get_nonce() == tx.nonce(), 'Invalid nonce');
```

crates/contracts/src/kakarot_core/eth_rpc.cairo
```diff
@@ -127,7 +127,11 @@ pub impl EthRPC<
     }
 
     fn eth_get_transaction_count(self: @TContractState, address: EthAddress) -> u64 {
-        panic!("unimplemented")
+        let kakarot_state = KakarotState::get_state();
```
enitrat, line 130: can you add a test, similar to what was done for chain_id ?

crates/evm/src/backend/validation.cairo
Gerson2102: Got it man!

Gerson2102:
I added one test. But I was thinking about change the nonce and check it again. But I dont really know how to do it...

enitrat:
1. don't deploy contracts with setup_contracts_for_testing, just do a local, internal test with 
https://github.com/kkrt-labs/kakarot-ssj/pull/983/files#diff-843e16e046c7baf8fcae8e8d2c2bfc5f8aecfed6d81af61768352c08e6ac83f7R231

2. you can use mock_call from starknet foundry to mock the account call to `get_nonce` to return an arbitrary value - you can check existing usages in the codebase

Gerson2102:
This is the test that I'm diong now:

```
#[test]
    fn test_eth_get_transaction_count() {
        let kakarot_state = set_up();
        // Deployed eoa should return a zero nonce
        let starknet_address = compute_starknet_address(
            test_address(), evm_address(), uninitialized_account()
        );
        start_mock_call::<u256>(starknet_address, selector!("get_nonce"), 0);
        assert_eq!(kakarot_state.eth_get_transaction_count(evm_address()), 0);
    }
```

And I'm getting this error:

```
[FAIL] contracts::kakarot_core::eth_rpc::tests::test_eth_get_transaction_count

Failure data:
    "Contract not deployed at address: 0x0"

```
That was the reason that I was using deploy_eoa() function before.

enitrat:
the issue was that the setup function does not intitialize anyu storage var, thus the class used for address computation was 0

enitrat (dismissed):

Issue #905, opened 2024-09-05
dev: optimize bitshifts by using a lookup table for powers of two


enitrat:
for the powers of two up to 256, we can define a lookup table and use that lookup table to execute bitshift operations, instead of calling the expensive cairo function `pow`. 

- If p <=256, use the lookup table
- if p>256, fallback on using `pow`.

manlikeHB:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
Hi, I am Cairo dev with lots of experience contributing to Cairo projects, my OD profile is a witness to this. This feat is right within my comfort zone.

I have had experience working on projects were bitshift were implemented I can use those as a reference for implementing this.

### How I plan on tackling this issue
After implementing this, I will write a robust unit test and ensure all test case are covered and the feat behaves as expected.

MPSxDev:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
Hello, I am Manuel, a process engineer and web3 developer. I have participated in Starknet Bootcamps and am an Elite winner of Speedrunstark. I have a high capacity to solve problems. I am a member of the DojoCoding community.
I hope this issue is assigned to me. I am available to work immediately to achieve what is required in the shortest time possible.

### How I plan on tackling this issue
To address the requirements of the issue, I will:

1. Create a lookup table `POW_2` containing powers of two up to 256.
2. Implement logic to check if the exponent `p` is less than or equal to 255:
   - If `p <= 255`, use the corresponding value from the `POW_2` lookup table for optimization.
   - If `p > 255`, fallback to using the `pow` function.
3. Ensure that the implementation utilizes the correct types by calculating the value 255 with `BitSize::<u256>::bits() - One::<T>::one()`.
4. Test the new implementation to verify correctness and performance improvements.

guha-rahul:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
I have been a rust and zk dev before and i have implemented lookup tables for zk cryptography . I think that will help me in this issue

### How I plan on tackling this issue
1. Read the codebase
2.implement the lookup table
3. replace the expensive cairo math with these tables
4. Ask questions if i am stuck anywhere

saimeunt:
[...]

PR #984, opened 2024-09-26
dev: optimized bitshifts by using a lookup table for powers of two

augustin-v:
## Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [✅ ] Other (please describe): dev

The bit shifting operations in the existing implementation are not optimized, leading to potential inefficiencies.

Resolves: #905 

## What is the new behavior?
- Introduced a new implementation for bit shifting using a lookup table for powers of two.
- Enhanced efficiency by minimizing multiplications and leveraging early returns for large shift values.

- [ ] Yes
- [✅ ] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/984)


enitrat (approved):
Great!

For curiosity, if you're interested, you can do some profiling to see how many steps are saved when doing bitshifts. The biggest difference should be when shifting 255 bits as it saves a lot of steps by skipping the expensive exponenciation

https://foundry-rs.github.io/starknet-foundry/snforge-advanced-features/profiling.html

enitrat:
The easiest way would be to use python to generate cairo code, like so:
```
def generate_powers_of_two():
    print("pub const POW_2: [")
    print("    u256")
    print("    ; 256] = [")

    for i in range(256):
        value = 2 ** i
        hex_value = f"0x{value:x}"
        print(f"    {hex_value},")

    print("];")

if __name__ == "__main__":
    generate_powers_of_two()
 ```

**Comment by augustin-v** [comment]
Thank you for the review, I'll add a new constant as instructed and try profiling as well!

File: crates/utils/src/math.cairo [hunk 1]
```diff
@@ -236,13 +237,21 @@ impl BitshiftImpl<
...
 > of Bitshift<T> {
     fn shl(self: T, shift: T) -> T {
         // if we shift by more than nb_bits of T, the result is 0
         // we early return to save gas and prevent unexpected behavior
         if shift > BitSize::<T>::bits().try_into().unwrap() - One::one() {
             panic_with_felt252('mul Overflow');
         }
+        // if the shift is within the bit size of u256 (<= 255 bits),
+        // use the POW_2 lookup table to get 2^shift for efficient multiplication
+        if shift <= BitSize::<u256>::bits().try_into().unwrap() - One::<T>::one() {
```
enitrat, line 251:
```suggestion
        if shift <= BitSize::<u256>::bits().try_into().unwrap() - One::<T>::one() {
```

```suggestion
        if shift.try_into().unwrap() <= BitSize::<u256>::bits() - One::<u32>::one() {
```
This change modifies the comparison to avoid a potential unwrap panic. Instead of converting the result of BitSize::<u256>::bits() to T, which might fail, it converts shift to usize. This is safer because we know that shift of type T will always fit in a `usize`

enitrat (dismissed):

augustin-v:
Hi @enitrat, understood. Thank you for the update, and please let me know if there's anything I can do to assist with resolving the issue

enitrat:
hey @augustin-v i ran some profiling on `test_execute_from_outside_eip1559` - which is just an empty execution, so it only performs validation steps:

Before:
image


After:image


It's nearly 10x better :)

augustin-v:
@enitrat wow it improved way more than what I thought it would, bitshift operations looking good

Issue #940, opened 2024-09-13
dev: optimize conversion of bytes to words


enitrat:
Some benchmarks:

```rust
pub fn ceil32(value: usize) -> usize {
    let ceiling = 32_u32;
    let (_q, r) = DivRem::div_rem(value, ceiling.try_into().unwrap());
    if r == 0_u8.into() {
        return value;
    } else {
        return (value + ceiling - r).into();
    }
}

#[inline(never)]
pub fn words_size(value: usize) -> usize {
    return (value + 31) / 32;
}

#[inline(never)]
pub fn words_size_ceil(size: usize) -> u64 {
    let words_size = (ceil32(size) / 32).into();
    return words_size;
}

#[cfg(test)]
mod tests{
    use super::{words_size, words_size_ceil};

    #[test]
    fn test_words_size(){
        let x = words_size(1);
        let x = words_size(32);
        let x = words_size(33);
    }

    #[test]
    fn test_words_size_ceil(){
        let x = words_size_ceil(1);
        let x = words_size_ceil(32);
        let x = words_size_ceil(33);
    }

}

```

<img width="326" alt="image" src="https://github.com/user-attachments/assets/467d572c-e8af-46d2-8548-7da83a094ed3">
<img width="745" alt="image" src="https://github.com/user-attachments/assets/f3056f72-4506-4234-b2d9-688374b0cb46">

It's almost twice cheaper to to `(bytes+31) / 32` then to `(ceil32(bytes) /32)` to compute the amount of 32bytes words required to represent `size` bytes.

- Locate all `ceil32(size)` usage throughout the codebase
- Replace with a function `32bytes_words_size` marked as `#[inline(always)]`

mubarak23:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
I am a Experience Cairo smart contract developer with experience working on projects such as Just Art Peace, Dojo, Kart, TBA, and Shinigami. Before transitioning to Cairo development, I was a backend developer specializing in Rust.



My recent work with cairo starknet
- https://github.com/keep-starknet-strange/raito/pull/128
- https://github.com/keep-starknet-strange/raito/pull/143
- https://github.com/horuslabsio/TBA/pull/45
- https://github.com/horuslabsio/karst-core/pull/60
- https://github.com/keep-starknet-strange/art-peace/pull/121
[...]

PR #985, opened 2024-09-27
Optimizes conversion of bytes to words

lordshashank:
## Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes

Resolves: #940 

## What is the new behavior?

- Introduces new function `bytes_32_words_size` which return number of 32 byte words required to represent `size`
- Replaces the `ceil32` function with `bytes_32_word_size` wherever applicable.

- [ ] Yes
- [X] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/985)

### DIFF
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -48,7 +48,7 @@ pub impl CreateHelpersImpl of CreateHelpers {
         let charged_gas = match create_type {
             CreateType::Create => gas::CREATE + memory_expansion.expansion_cost + init_code_gas,
             CreateType::Create2 => {
-                let calldata_words = bytes_32_words_size(size) / 32;
+                let calldata_words = bytes_32_words_size(size);
                 gas::CREATE
                     + gas::KECCAK256WORD * calldata_words.into()
                     + memory_expansion.expansion_cost
--- a/crates/evm/src/create_helpers.cairo
+++ b/crates/evm/src/create_helpers.cairo
@@ -13,7 +13,7 @@ use crate::stack::StackTrait;
 use crate::state::StateTrait;
 use utils::address::{compute_contract_address, compute_create2_contract_address};
 use utils::constants;
-use utils::helpers::ceil32;
+use utils::helpers::bytes_32_words_size;
 use utils::set::SetTrait;
 use utils::traits::{
     BoolIntoNumeric, EthAddressIntoU256, U256TryIntoResult, SpanU8TryIntoResultEthAddress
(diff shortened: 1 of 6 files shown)


crates/evm/src/create_helpers.cairo
```diff
@@ -48,7 +48,7 @@ pub impl CreateHelpersImpl of CreateHelpers {
         let charged_gas = match create_type {
             CreateType::Create => gas::CREATE + memory_expansion.expansion_cost + init_code_gas,
             CreateType::Create2 => {
-                let calldata_words = ceil32(size) / 32;
+                let calldata_words = bytes_32_words_size(size) / 32;
```
enitrat, line 51:
```suggestion
                let calldata_words = bytes_32_words_size(size);
```

you have one extra `/32` here

crates/evm/src/gas.cairo
```diff
@@ -194,7 +194,7 @@ pub fn memory_expansion(
         }
     }?;
 
-    let new_size = helpers::ceil32(max_size);
+    let new_size = helpers::bytes_32_words_size(max_size) * 32;
```
enitrat, line 197: for this specific case, we want the size in _bytes_, not in 32-bytes words. Thus we can keep using ceil32

```suggestion
    let new_size = helpers::ceil32(max_size);
```

crates/utils/src/helpers.cairo
```diff
@@ -55,6 +55,22 @@ pub fn ceil32(value: usize) -> usize {
...
+/// The number of 32-byte words required to represent `size` bytes
+///
+/// # Examples
+/// bytes_32_words_size(2) = 1
+/// bytes_32_words_size(34) = 2
+#[inline(always)]
+pub fn bytes_32_words_size(size: usize) -> usize {
+    (size + 31) / 32
+}
+
```
enitrat, line 73: what happens if `size + 31` > MAX_USIZE ?

In that case, the function will panic. What we can do to avoid this is use the `SaturatingAdd` trait to saturate the result to the max size of a usize.

Also, please add a test for this edge case

crates/evm/src/instructions/environmental_information.cairo
```diff
@@ -238,8 +238,8 @@ pub impl EnvironmentInformationImpl of EnvironmentInformationTrait {
         }
         ensure(!(last_returndata_index > return_data.len()), EVMError::ReturnDataOutOfBounds)?;
 
-        //TODO: handle overflow in ceil32 function.
-        let words_size = (ceil32(size.into()) / 32).into();
+        //TODO: handle overflow in bytes_32_words_size function.
```
enitrat, line 241:
```suggestion
```

Once handled in `bytes_32_words_size` we can remove this TODO

crates/utils/src/helpers.cairo
enitrat: we can also delete ceil32 if it's not longer used anywhere :)

enitrat (approved):

Issue #941, opened 2024-09-13
feat: implement `eth_get_balance`


enitrat:
Implement the method `eth_get_balance` of the `EthRPC` trait

https://github.com/kkrt-labs/kakarot-ssj/blob/b1f7f8c13be38edc081f0738c188561240c098b4/crates/contracts/src/kakarot_core/eth_rpc.cairo#L124-L126

Given an evm address, this will return the balance of this address.

Use this where required inside the codebase, e.g.

https://github.com/kkrt-labs/kakarot-ssj/blob/b1f7f8c13be38edc081f0738c188561240c098b4/crates/evm/src/backend/validation.cairo#L52-L56

https://github.com/kkrt-labs/kakarot-ssj/blob/b1f7f8c13be38edc081f0738c188561240c098b4/crates/evm/src/backend/starknet_backend.cairo#L120-L125

stevencartavia:
Hi, I would like to work on this!

enitrat:
Hi, this will be for the ODHack week

bitfalt:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
Hii! My name is Daniel, I’m a Computer Science student and web3 developer. I’ve done some contributions, finished a Starknet Bootcamp, exercises on NodeGuardians, finished Starklings and I have participated in multiple Hackathons. I’m also a member of the Dojo Coding community.I’ve contributed to the project in the past.

### How I plan on tackling this issue
I would approach this issue in the following way:
- Create the implementation for the eth_get_balance
- Refactor the code to import the EthRPC trait properly (i’ve already done it in the past contribution)
- Refactor the codebase to call eth_get_balance where it is necessary
- I would ask the maintainers if any questions come up during my implementation

manlikeHB:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
Hi, I am Cairo dev with lots of experience contributing to Cairo projects, my OD profile is a witness to this. This feat is right within my comfort zone.

### How I plan on tackling this issue
I would go through the code base and understand the functionality expected, after that I would write a robust test ensuring all test cases are covered and the feat behaves as expected.

mahesh-dalle:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

[...]

PR #990, opened 2024-09-28
Implement eth_get_balance

saimeunt:
This PR implements `eth_get_balance` and use it instead of querying the balanceOf function of the `native_token` in the codebase. It also fixes a suspected bug in `get_starknet_address` from `KakarotCore`.

## Pull Request type

- [x] Bugfix
- [x] Feature
- [ ] Build-related changes

The `eth_get_balance` is not implemented and thus we need to query the `native_token` directly in several parts of the codebase.

Resolves: #941 

## What is the new behavior?

The `eth_get_balance` is now implemented and replicates its `eth_getBalance` counterpart in the EVM.

- The implementation of `eth_get_balance` leverages the `get_starknet_address` function from `KakarotCore` to get the corresponding starknet address from an evm address. Once we have the underlying starknet address, we query the `native_token` balance. The function has also been added to `ExtendedKakarotCore` interface.
- When reading about the description of the `get_starknet_address`, I suspected a bug and from my understanding we should return the registered starknet address only if it is found and thus non-zero, so I think the if condition should be inversed, this is why my PR introduces a bugfix.
- The occurences of direct `native_token` querying have been replaced by `eth_get_balance`, in particular for the `eth_validate_tx` test helper I had to mock additional calls so `account.get_evm_address` returns the expected `account_starknet_address`.
- A simple test has been added to ensure the correct behavior of `eth_get_balance` both for uninitialized and initialized accounts.

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/990)

### DIFF
--- a/crates/contracts/src/kakarot_core/kakarot.cairo
+++ b/crates/contracts/src/kakarot_core/kakarot.cairo
@@ -211,7 +211,7 @@ pub mod KakarotCore {
         // @return starknet_address The Starknet Account Contract address
         fn get_starknet_address(self: @ContractState, evm_address: EthAddress) -> ContractAddress {
             let registered_starknet_address = self.address_registry(evm_address);
-            if (registered_starknet_address.is_zero()) {
+            if (!registered_starknet_address.is_zero()) {
                 return registered_starknet_address;
             }
 
--- a/crates/evm/src/backend/validation.cairo
+++ b/crates/evm/src/backend/validation.cairo
@@ -57,8 +57,8 @@ mod tests {
     use contracts::kakarot_core::KakarotCore;
     use core::num::traits::Bounded;
     use core::ops::SnapshotDeref;
-    use core::starknet::{EthAddress, ContractAddress};
     use core::starknet::storage::{StorageTrait, StoragePathEntry};
+    use core::starknet::{EthAddress, ContractAddress};
     use evm::gas;
     use snforge_std::cheatcodes::storage::{store_felt252};
     use snforge_std::{
--- a/crates/evm/src/backend/validation.cairo
+++ b/crates/evm/src/backend/validation.cairo
@@ -5,7 +5,6 @@ use core::ops::SnapshotDeref;
 use core::starknet::storage::{StoragePointerReadAccess};
 use core::starknet::{get_caller_address};
 use crate::gas;
-use openzeppelin::token::erc20::interface::{IERC20CamelDispatcher, IERC20CamelDispatcherTrait};
 use starknet::storage::StorageTrait;
 use utils::eth_transaction::check_gas_fee;
 use utils::eth_transaction::transaction::{Transaction, TransactionTrait};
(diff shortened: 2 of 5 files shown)


crates/contracts/src/kakarot_core/kakarot.cairo
```diff
@@ -211,7 +211,7 @@ pub mod KakarotCore {
         // @return starknet_address The Starknet Account Contract address
         fn get_starknet_address(self: @ContractState, evm_address: EthAddress) -> ContractAddress {
             let registered_starknet_address = self.address_registry(evm_address);
-            if (registered_starknet_address.is_zero()) {
+            if (!registered_starknet_address.is_zero()) {
```
enitrat, line 214: !! thanks for spotting this!

you can take https://github.com/kkrt-labs/kakarot-ssj/issues/991 if you want

crates/evm/src/backend/validation.cairo
```diff
@@ -81,7 +78,10 @@ mod tests {
         let kakarot_state = KakarotCore::unsafe_new_contract_state();
         let kakarot_storage = kakarot_state.snapshot_deref().storage();
         let kakarot_address = test_address();
-        let account_starknet_address = 'account_starknet_address'.try_into().unwrap();
+        let account_evm_address_felt = 'account_evm_address';
+        let account_evm_address: EthAddress = account_evm_address_felt.try_into().unwrap();
```
enitrat, line 82:
```suggestion
        let account_evm_address: EthAddress = 'account_evm_address'.try_into().unwrap();
```

crates/evm/src/backend/validation.cairo
```diff
@@ -81,7 +78,10 @@ mod tests {
...
+        let account_starknet_address_felt = 'account_starknet_address';
+        let account_starknet_address = account_starknet_address_felt.try_into().unwrap();
```
enitrat, line 84:
```suggestion
        let account_starknet_address = 'account_starknet_address'.try_into().unwrap();
```

crates/evm/src/backend/validation.cairo
```diff
@@ -94,10 +94,21 @@ mod tests {
         store_felt252(kakarot_address, base_fee_storage, 1_000_000_000); // 1 Gwei
         store_felt252(kakarot_address, block_gas_limit_storage, BLOCK_GAS_LIMIT.into());
         store_felt252(kakarot_address, native_token_storage_address, native_token_address.into());
+        store(
+            kakarot_address,
+            map_entry_address(
```
enitrat, line 99: I didn't know about this function from starknet-foundry, nice!

I found another way of achieving the same thing here:
https://github.com/kkrt-labs/kakarot-ssj/blob/main/crates/contracts/tests/test_kakarot_core.cairo#L118-L126

I prefer this version, because it has guarantees to be type safe.

enitrat (commented):
great work!

saimeunt:
@enitrat Thank you for your quick review and your Cairo tips, I made the requested changes by following your best practices insights.

I will apply to #991 and I'd like to continue my work on Kakarot so this is appreciated! 🙌

enitrat (dismissed):

enitrat (approved):

Issue #991, opened 2024-09-28
test: `fn get_starknet_address`


enitrat:
test `fn get_starknet_address` to ensure
1. if the address is registered in the mapping, it's returned
2. if not, we return the result of the computation

the test will use starknet-foundry's "store" and "load" cheatcodes, as already used in:
https://github.com/lordshashank/kakarot-ssj/blob/ed7e4145a0bb2517c60c6904fd317d11d1b00625/crates/contracts/tests/test_kakarot_core.cairo#L96-L131

The logic should be:
for 1. you can just put any value in the mapping and verify it's returned
for 2., make sure the mapping is empty, and that `get_starknet_address` returns the same as `compute_starknet_address`.

reserved for @saimeunt

Damilola145:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
Hello,I'm a front-end developer with a resolve to confront and solve new challenges. This alongside my experience will aid me in resolving this issue

### How I plan on tackling this issue
I would carefully go through the github link highlighted to aid the implementation of the cheatcodes to resolve the issue, and also carry out the test as required

ScottyDavies:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
I am a blockchaindevelopereveloper

### How I plan on tackling this issue
I would approach this issue with the following steps:

For Testing Address Registration:
Add a value to the mapping.
Verify that the expected value is returned when calling get_starknet_address.

For Testing Computation:
Ensure the mapping is empty.
Check that get_starknet_address returns the same value as compute_starknet_address.

saimeunt:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
I'm a returning contributor and I have some experience in Cairo testing.

### How I plan on tackling this issue
I will carefully implement the test as explained in the issue description.

akintewe:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/kakarot-zkevm).

### My background and how it can be leveraged
[...]

PR #1006, opened 2024-10-01
test: get_starknet_address

saimeunt:
## Pull Request type

Unit test of the KakarotCore::get_starknet_address function

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes

Function not tested.

Resolves: #991 

## What is the new behavior?

Function tested with full coverage of the 2 possible outcomes.

- [ ] Yes
- [x] No

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/1006)

### DIFF
--- a/crates/contracts/tests/test_kakarot_core.cairo
+++ b/crates/contracts/tests/test_kakarot_core.cairo
@@ -165,7 +165,7 @@ fn test_kakarot_core_get_starknet_address() {
     let mut kakarot_state = KakarotCore::unsafe_new_contract_state();
 
     let registered_evm_address = test_utils::evm_address();
-    let registered_starknet_address = kakarot_core.compute_starknet_address(registered_evm_address);
+    let registered_starknet_address = test_utils::starknet_address();
     let registered_map_entry_address = kakarot_state
         .snapshot_deref()
         .storage()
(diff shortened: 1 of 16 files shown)


crates/contracts/tests/test_kakarot_core.cairo
```diff
@@ -159,6 +159,49 @@ fn test_kakarot_core_upgrade_contract() {
...
+        .__storage_pointer_address__;
+    // store the registered address in the mapping
+    store(
+        kakarot_core.contract_address,
+        registered_map_entry_address.into(),
+        [registered_starknet_address.into()].span()
+    );
+    // when the address is registered in the mapping, it's returned
+    assert_eq!(
+        kakarot_core.get_starknet_address(registered_evm_address), registered_starknet_address
```
enitrat, line 184: There is an issue with this approach, which is that you're 
1. computing the expected starknet address
2. registring this address in the mapping
3. verifying that the value returned is the one read in the mapping.

however, given that the value in the mapping is the one expected to be returned by the compute_starknet_address, you can't actually verify this behavior.

You should just store an arbitrary value, and verify it's returned.

saimeunt:
@enitrat Thank you for your review and thorough explanation on what was problematic with my test. I've made the requested changes please tell me if that's OK now.

enitrat (approved):
great! thank you !

PR #1019, opened 2024-10-04
refactor: use a match table for execute_opcode"

enitrat:
more optimized than if / elses

- - -
This change is [Reviewable](https://reviewable.io/reviews/kkrt-labs/kakarot-ssj/1019)


crates/evm/src/interpreter.cairo
```diff
@@ -410,605 +410,414 @@ pub impl EVMImpl of EVMTrait {
     }
 
     fn execute_opcode(ref self: VM, opcode: u8) -> Result<(), EVMError> {
-        // Call the appropriate function based on the opcode.
-        if opcode == 0x00 {
-            // STOP
-            return Result::Ok(self.exec_stop());
+        match opcode {
+            0 => // STOP
```
obatirou, line 414: Would be better in hex notation no ?

crates/evm/src/interpreter.cairo
```diff
@@ -410,605 +410,414 @@ pub impl EVMImpl of EVMTrait {
...
+            9 => // MULMOD
+            self.exec_mulmod(),
+            10 => // EXP
+            self.exec_exp(),
+            11 => // SIGNEXTEND
+            self.exec_signextend(),
+            12 => Result::Err(EVMError::InvalidOpcode(opcode)),
+            13 => Result::Err(EVMError::InvalidOpcode(opcode)),
+            14 => Result::Err(EVMError::InvalidOpcode(opcode)),
+            15 => Result::Err(EVMError::InvalidOpcode(opcode)),
```
obatirou, line 441: Why adding those ?
Removing it might be easier to read and will go into the default case

enitrat: I can't skip numbers unfortunately they have to be sequential 😮‍💨

PR #1023, opened 2024-11-19
[KGA-19] feat: add new_call_contract_syscall to Cairo1Helpers class

enitrat:
required for https://github.com/kkrt-labs/kakarot/pull/1625

### DIFF
--- a/crates/contracts/src/cairo1_helpers.cairo
+++ b/crates/contracts/src/cairo1_helpers.cairo
@@ -1,4 +1,4 @@
-use core::starknet::{EthAddress, secp256_trait::Signature};
+use core::starknet::{EthAddress, secp256_trait::Signature, ContractAddress};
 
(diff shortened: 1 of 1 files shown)


crates/contracts/src/cairo1_helpers.cairo
```diff
@@ -217,6 +238,22 @@ pub mod embeddable_impls {
 
             return is_valid_signature(msg_hash, r, s, public_key);
         }
+
+        fn new_call_contract_syscall(
```
obatirou, line 242: no tests ?

enitrat: tested with cairozero e2e

enitrat: fwiw I ran the tests on a local build of kakarot with this new class and it worked

